PR #1, opened 2020-07-09
Screen Reader

dwieeb:
TODO
- [x] memory leak stuff

see https://github.com/ionic-team/capacitor/pull/3364 and https://github.com/ionic-team/capacitor-testapp/pull/2


carlpoole:
Just wanted to make a note: this plugin is susceptible to memory leaks when used in embedded. I think we need to add a call to `removeTouchExplorationStateChangeListener` on the destroy/unload event.

dwieeb:
Something to be cognizant of when migrating these plugins. I added a Plugin Transfer Checklist to https://github.com/ionic-team/capacitor/issues/3227.

screen-reader/ios/Plugin/ScreenReaderPlugin.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+import Capacitor
+
+/**
+ * Please read the Capacitor iOS Plugin Development Guide
+ * here: https://capacitorjs.com/docs/plugins/ios
+ */
```
jcesarmobile, line 7:
```suggestion

```
This should be removed, it's there to help people get started when they develop plugins, but should not be part of the final plugin

screen-reader/ios/PluginTests/PluginTests.swift
```diff
@@ -0,0 +1,26 @@
...
+    func testEcho() {
+        // This is an example of a functional test case for a plugin.
+        // Use XCTAssert and related functions to verify your tests produce the correct results.
+
+        let implementation = ScreenReader()
+        let value = "Hello, World!"
+        let result = implementation.echo(value)
+
+        XCTAssertEqual(value, result)
+    }
```
jcesarmobile, line 25:
```suggestion

```

if we have no real test, better remove, it will fail as echo was removed

screen-reader/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@capacitor/screen-reader",
+  "version": "0.0.1",
+  "description": "Screen Reader APIs",
+  "main": "dist/plugin.js",
+  "module": "dist/esm/index.js",
+  "types": "dist/esm/index.d.ts",
+  "scripts": {
+    "lint": "npm run prettier -- --check && npm run swiftlint -- lint",
+    "fmt": "npm run prettier -- --write && npm run swiftlint -- autocorrect",
```
jcesarmobile, line 10:
```suggestion
    "fmt": "npm run prettier -- --write && npm run swiftlint -- autocorrect  --format",
```

Add the `--format` param

dwieeb:
I think we should start with 0.0.1 while Capacitor 3 is in alpha. We can switch them all to 1.0.0 when Capacitor 3 is released.

screen-reader/android/src/main/java/com/capacitorjs/plugins/screenreader/ScreenReader.java
```diff
@@ -0,0 +1,58 @@
...
+import android.content.Context;
+import android.speech.tts.TextToSpeech;
+import android.view.accessibility.AccessibilityManager;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
+public class ScreenReader {
+    private Context context;
+    private AccessibilityManager am;
```
carlpoole, line 12: Rename to accessibilityManager

screen-reader/android/src/main/java/com/capacitorjs/plugins/screenreader/ScreenReaderPlugin.java
```diff
@@ -0,0 +1,47 @@
...
+import com.getcapacitor.JSObject;
+import com.getcapacitor.NativePlugin;
+import com.getcapacitor.Plugin;
+import com.getcapacitor.PluginCall;
+import com.getcapacitor.PluginMethod;
+
+@NativePlugin(name = "ScreenReader")
+public class ScreenReaderPlugin extends Plugin {
+    public static final String EVENT_SCREEN_READER_STATE_CHANGE = "screenReaderStateChange";
+    private ScreenReader sr;
```
carlpoole, line 12: Rename to screenReader

PR #5, opened 2020-07-30
Haptics

jcesarmobile:
Android implementation has been reworked based on https://github.com/expo/expo/tree/master/packages/expo-haptics

see https://github.com/ionic-team/capacitor-testapp/pull/7 and https://github.com/ionic-team/capacitor/pull/3363


haptics/src/web.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  notification(options: HapticsNotificationOptions): void {
+    console.log('notification options', options);
+  }
+
+  vibrate(options?: VibrateOptions): void {
+    if (navigator.vibrate) {
+      const duration = options?.duration || 300;
+      navigator.vibrate(duration);
+    }
```
dwieeb, line 26: Is this usually how we handle feature detection? In the Accessibility plugin, we reject with an error: https://github.com/ionic-team/capacitor/blob/main/core/src/web/accessibility.ts#L21-L23

jcesarmobile: At the moment none of the Haptics methods return anything, so we can't reject it. But we can change it to return promises as other plugins do and then reject when appropriate.

dwieeb: We could also `throw` if the method is synchronous. 

We have two choices when features in a particular browser aren't supported:
1. Be silent about the missing feature and do nothing (perhaps log a warning). 
1. Throw an error (or reject) about the missing feature.

I think we should choose one of these and be consistent throughout all plugins.

haptics/src/web.ts
```diff
@@ -11,30 +11,38 @@ export class HapticsWeb extends WebPlugin implements HapticsPlugin {
     super({ name: 'Haptics' });
   }
 
-  impact(options: HapticsImpactOptions): void {
+  async impact(options: HapticsImpactOptions): Promise<void> {
     console.log('impact options', options);
+    return Promise.resolve();
```
dwieeb, line 16: Isn't necessary in an `async` function

```suggestion
```

haptics/src/web.ts
```diff
@@ -11,30 +11,38 @@ export class HapticsWeb extends WebPlugin implements HapticsPlugin {
...
   }
 
-  notification(options: HapticsNotificationOptions): void {
+  async notification(options: HapticsNotificationOptions): Promise<void> {
     console.log('notification options', options);
+    return Promise.resolve();
```
dwieeb, line 21:
```suggestion
```

haptics/src/web.ts
```diff
@@ -11,30 +11,38 @@ export class HapticsWeb extends WebPlugin implements HapticsPlugin {
...
   }
 
-  vibrate(options?: VibrateOptions): void {
+  async vibrate(options?: VibrateOptions): Promise<void> {
     if (navigator.vibrate) {
       const duration = options?.duration || 300;
       navigator.vibrate(duration);
+      return Promise.resolve();
```
dwieeb, line 28:
```suggestion
```

haptics/src/web.ts
```diff
@@ -11,30 +11,38 @@ export class HapticsWeb extends WebPlugin implements HapticsPlugin {
...
+    } else {
+      return Promise.reject('Vibrate API not available in this browser')
```
dwieeb: If you `throw` in an `async` function it rejects automatically

```suggestion
      throw new Error('Vibrate API not available in this browser');
```

haptics/src/web.ts
```diff
@@ -11,30 +11,38 @@ export class HapticsWeb extends WebPlugin implements HapticsPlugin {
...
     }
   }
 
-  selectionStart(): void {
+  async selectionStart(): Promise<void> {
     console.log('selectionStart');
+    return Promise.resolve();
```
dwieeb, line 36:
```suggestion
```

haptics/src/web.ts
```diff
@@ -11,30 +11,38 @@ export class HapticsWeb extends WebPlugin implements HapticsPlugin {
...
   }
 
-  selectionChanged(): void {
+  async selectionChanged(): Promise<void> {
     console.log('selectionChanged');
+    return Promise.resolve();
```
dwieeb, line 41:
```suggestion
```

haptics/src/web.ts
```diff
@@ -11,30 +11,38 @@ export class HapticsWeb extends WebPlugin implements HapticsPlugin {
...
   }
 
-  selectionEnd(): void {
+  async selectionEnd(): Promise<void> {
     console.log('selectionEnd');
+    return Promise.resolve();
```
dwieeb, line 46:
```suggestion
```

cameronrr:
Thanks for all the work on this. Two points for discussion below:

Message for'API Not Available'
In the Share plugin, a false feature detection returns 'Web Share API not available'. Is it worth having a consistent messaging from Capacitor plugins in this scenario i.e. if Haptics is not available, to output 'Web Haptics API not available' instead of 'Vibrate API not available in this browser'. Specificity could follow the standard message if needed i.e. 'Web Haptics API not available - not supported by this browser'. There might be value in a consistent format not only for clearer plugin development but for developers to quickly recognise the pattern and that it is from the Capacitor web plugin.

Vibrate Options
The Web API supports not only a number as the duration parameter, but also an array (Number[]) to define a vibration pattern. Is it possible to handle this in the web plugin to leverage the full set of Vibrate API features? I think the main consideration is the interaction with native plugins if the options/capabilities are different. I think the goal is a single interface for developers regardless of device so not sure how this would be approached.

haptics/src/web.ts
jcesarmobile: I know it's not needed, but I think it's better to explicitly return the promise to make it more clear and more consistent with iOS/Android code.

But also, I've just noticed that iOS and Android return an empty object (`{}`), instead of undefined, we should be consistent about that too. 
Since the types say `Promise<void>`, I think the native plugins should return undefined instead of the empty object, but it's not possible at the moment, we should make the resolve return that.
Or we should have an `EmptyObject` in core plugins could use to return `Promise<EmptyObject>` instead of `Promise<void>`

haptics/src/web.ts
jcesarmobile: I know, but as I said, I think it's better (more consistent) to return the promise explicitly for consistency with the native code

dwieeb:
@cameronrr Good input, thanks! I agree, we need to be consistent about error messages. I hope to have a code associated with each error so you can just check `if (e.code === 'ERR_WEB_API_UNAVAILABLE")` for example. As for the vibrate options, I'm not sure it's in the scope of this plugin. We want people to use the high-level haptics methods. We've debated removing the `vibrate()` method completely. I think there's room for an "advanced vibration" plugin in the [Capacitor community](https://github.com/capacitor-community/) (and keep in mind, iOS doesn't support patterns or even different length vibrations).

haptics/ios/Plugin/Haptics.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    @objc public func impact(_ style: String?) {
+        DispatchQueue.main.async {
+            if let style = style {
+                var impactStyle = UIImpactFeedbackGenerator.FeedbackStyle.heavy
+                if style == "MEDIUM" {
+                    impactStyle = UIImpactFeedbackGenerator.FeedbackStyle.medium
+                } else if style == "LIGHT" {
+                    impactStyle = UIImpactFeedbackGenerator.FeedbackStyle.light
+                }
```
ikeith, line 16: I think it would be better to move this string comparison to the Plugin class. Then this method could take a `FeedbackStyle` parameter. That would make this class far more usable by other code instead of just taking an undocumented `String` parameter where the expected values are unclear.

haptics/ios/Plugin/Haptics.swift
```diff
@@ -0,0 +1,71 @@
...
+    @objc public func notification(_ type: String?) {
+        DispatchQueue.main.async {
+            let generator = UINotificationFeedbackGenerator()
+            if let type = type {
+                var notificationType = UINotificationFeedbackGenerator.FeedbackType.success
+                if type == "WARNING" {
+                    notificationType = UINotificationFeedbackGenerator.FeedbackType.warning
+                } else if type == "ERROR" {
+                    notificationType = UINotificationFeedbackGenerator.FeedbackType.error
+                }
```
ikeith, line 36: Same note about the string comparison.

haptics/android/src/main/java/com/capacitorjs/plugins/haptics/Haptics.java
```diff
@@ -0,0 +1,69 @@
...
+        mVibrator.vibrate(duration);
+    }
+
+    @SuppressWarnings({ "deprecation" })
+    private void vibratePre26(long[] pattern, int repeat) {
+        mVibrator.vibrate(pattern, repeat);
+    }
+
+    public void impact(String style) {
+        this.performHaptics(HapticsImpactType.fromString(style));
```
ikeith, line 41: As in the iOS code, I think this would be better to translate in the Plugin class.

haptics/android/src/main/java/com/capacitorjs/plugins/haptics/arguments/HapticsImpactType.java
```diff
@@ -0,0 +1,43 @@
+package com.capacitorjs.plugins.haptics.arguments;
+
+public enum HapticsImpactType implements HapticsVibrationType {
+    LIGHT("LIGHT", new long[] { 0, 50 }, new int[] { 0, 110 }, new long[] { 0, 20 }),
+    MEDIUM("MEDIUM", new long[] { 0, 43 }, new int[] { 0, 180 }, new long[] { 0, 43 }),
+    HEAVY("HEAVY", new long[] { 0, 60 }, new int[] { 0, 255 }, new long[] { 0, 61 });
+
+    private final String mType;
```
carlpoole, line 8: I think we should abandon Hungarian notation in our plugins. Its kinda old and Google advises against it now in their Java style guide https://google.github.io/styleguide/javaguide.html#s5.1-identifier-names

haptics/ios/Plugin/Haptics.swift
jcesarmobile: Then I don't think we need the Haptics.swift class at all, users can just use the Haptics classes themselves

nicolocarpignoli:
Hi everybody. 
It seems that from iOS 13+ it is possible to use CoreHaptics API on iOS and handle more fine grained haptics feedbacks.
For example, we can manage the duration https://developer.apple.com/documentation/corehaptics/chhapticevent/3081790-duration?language=objc.

Does this PR introduced such behaviour, or it is not based on CoreHaptics?

Thank you

PR #6, opened 2020-07-31
Clipboard

carlpoole:
- [x] Android
- [x] iOS

closes https://github.com/ionic-team/capacitor/issues/376


clipboard/ios/Plugin/Clipboard.swift
```diff
@@ -0,0 +1,72 @@
...
+
+@objc public class Clipboard: NSObject {
+    
+    @objc enum ContentType: Int {
+        case string
+        case url
+        case image
+    }
+    
+    @objc func write(_ label: ContentType, _ content: String) {
```
ikeith, line 12: Omitting parameter names is idiomatic for JavaScript and makes sense for the methods exposed on the Plugin wrapper. However, internal implementations should better conform to platform conventions. So, for example, a more idiomatic way to declare this method would be:
```suggestion
    @objc func write(content: String, ofType type: ContentType) {
```

clipboard/ios/Plugin/Clipboard.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @objc func read() -> [String: Any] {
+        if UIPasteboard.general.hasStrings {
+            return [
+                "value": UIPasteboard.general.string!,
+                "type": "text/plain"
+            ]
+            
+        }
```
ikeith, line 40: Rather than checking `hasStrings`, `hasURLs`, etc. before accessing the data, it would be better to try to access the data with an optional binding. In addition to being simpler, it will also eliminate the force unwrap (which is legal since the value can be assumed to exist but causes linting problems).
```suggestion
        if if let stringValue = UIPastboard.general.string {
            return [
                "value": stringValue,
                "type": "text/plain"
            ]
        }
```
And so on for the other types.

clipboard/ios/Plugin/Clipboard.swift
```diff
@@ -0,0 +1,72 @@
...
+        switch label {
+        case ContentType.string:
+                UIPasteboard.general.string = content
+        case ContentType.url:
+            if let url = URL(string: content) {
+                UIPasteboard.general.url = url
+            }
+        case ContentType.image:
+            if let data = Data(base64Encoded: getCleanData(content)) {
+                let image = UIImage(data: data)
```
ikeith, line 22: These two steps should be combined into a single `if` statement since they are logically linked, could fail, and have a force unwrap below in the log statement (which will crash if the conversion did fail).
```suggestion
            if let data = Data(base64Encoded: getCleanData(content)), let image = UIImage(data: data) {
```

clipboard/ios/Plugin/ClipboardPlugin.swift
```diff
@@ -0,0 +1,27 @@
...
+@objc(ClipboardPlugin)
+public class ClipboardPlugin: CAPPlugin {
+    private let implementation = Clipboard()
+
+    @objc func read(_ call: CAPPluginCall) {
+        let result = implementation.read()
+        
+        if (!result.isEmpty) {
+            call.success(result)
+        }
```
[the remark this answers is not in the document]
dwieeb: Is there a swiftlint rule we can use to remove the parens automatically?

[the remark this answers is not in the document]
ikeith: Apparently! https://realm.github.io/SwiftLint/control_statement.html

carlpoole:
@ikeith I modified to return success or error. In conjunction with the test app PR I made, this is testable on iOS and Android. All seems to work okay. Currently does not contain any permission logic, which I think is an iOS 14 thing?

clipboard/ios/Plugin/ClipboardPlugin.swift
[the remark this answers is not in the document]
dwieeb: Added an issue: https://github.com/ionic-team/swiftlint-config/issues/3

ikeith:
> Currently does not contain any permission logic, which I think is an iOS 14 thing?

iOS 14 adds an alert when an application reads the clipboard but it doesn't actually require permission to access it so nothing is needed here. It's about making a previously invisible action visible to the user and shaming applications that abuse it.

clipboard/ios/Plugin/Clipboard.swift
```diff
@@ -0,0 +1,69 @@
...
+    @objc func read() -> [String: Any] {
+        if let stringValue = UIPasteboard.general.string {
+            return [
+                "value": stringValue,
+                "type": "text/plain"
+            ]
+        }
+        
+        if UIPasteboard.general.hasURLs {
+            let url = UIPasteboard.general.url!
```
ikeith, line 41: For consistency and to eliminate the force-unwrap, this can be simplified
```suggestion
        if let url = UIPasteboard.general.url {
```

clipboard/ios/Plugin/Clipboard.swift
```diff
@@ -0,0 +1,69 @@
...
+            return [
+                "value": url.absoluteString,
+                "type": "text/plain"
+            ]
+        }
+        
+        if UIPasteboard.general.hasImages {
+            let image = UIPasteboard.general.image!
```
ikeith, line 49: Same
```suggestion
        if let image = UIPasteboard.general.image {
```

clipboard/ios/Plugin/ClipboardPlugin.swift
```diff
@@ -0,0 +1,36 @@
...
+            success = true
+        } else if let imageBase64 = call.options["image"] as? String {
+            implementation.write(content: imageBase64, ofType: Clipboard.ContentType.image)
+            success = true
+        }
+        
+        if success {
+            call.resolve()
+        } else {
+            call.error("No data provided")
```
ikeith, line 33: Shouldn't this be `reject`? My understanding is that we were going to prefer `resolve` & `reject` over `success` & `error`.

carlpoole: Will change it now. Do we want to add the reject to the checklist also @dwieeb 
` I have switched all deprecated .success() calls to use .resolve().`

clipboard/android/src/main/java/com/capacitorjs/plugins/clipboard/Clipboard.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private Context context;
+    private String logTag = Logger.tags(this.getClass().getSimpleName());
+
+    public Clipboard(Context context) {
+        this.context = context;
+    }
+
+    public void write(String label, ContentType contentType, String content) {
+        ClipboardManager clipboard = (ClipboardManager) context.getSystemService(Context.CLIPBOARD_SERVICE);
```
dwieeb, line 26: Is it normal to use `getSystemService()` several times in methods instead of once in the constructor?

clipboard/android/src/main/java/com/capacitorjs/plugins/clipboard/Clipboard.java
```diff
@@ -0,0 +1,75 @@
...
+public class Clipboard {
+
+    public enum ContentType {
+        STRING,
+        IMAGE,
+        URL
+    }
+
+    private Context context;
+    private String logTag = Logger.tags(this.getClass().getSimpleName());
```
dwieeb, line 19: Aren't these usually private, static, final? We should be consistent

clipboard/android/src/main/java/com/capacitorjs/plugins/clipboard/Clipboard.java
```diff
@@ -0,0 +1,75 @@
...
+                data = ClipData.newPlainText(label, content);
+                break;
+        }
+
+        if (data != null && clipboard != null) {
+            clipboard.setPrimaryClip(data);
+        }
+    }
+
+    public JSObject read() {
```
dwieeb, line 42: If I were to use this class from Java, I would expect a more type-safe container for clipboard data. This may not be a use case for `JSObject`. Maybe a separate class instead?

clipboard/src/definitions.ts
```diff
@@ -0,0 +1,43 @@
...
+   */
+  read(): Promise<ClipboardReadResult>;
+}
+
+/**
+ * Represents the data to be written to the clipboard.
+ *
+ * @since 1.0.0
+ */
+export interface ClipboardWrite {
```
dwieeb, line 28: May I suggest `ClipboardWriteOptions` for a name instead?

clipboard/src/definitions.ts
```diff
@@ -0,0 +1,43 @@
...
+  string?: string;
+  image?: string;
+  url?: string;
+  label?: string; // Android only
```
dwieeb, line 32: These should each have JSDoc blocks, as well 😬 

JS documentation tools ignore `//` comments. If we want users to know something like that, they should be in JSDoc comment blocks. Then it will come through in VS Code and be generated on our website.

clipboard/src/definitions.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/**
+ * Represents the data read from the clipboard.
+ *
+ * @since 1.0.0
+ */
+export interface ClipboardReadResult {
+  value: string;
+  type: string;
```
dwieeb, line 42: Document these properties with JSDoc blocks.

clipboard/src/web.ts
```diff
@@ -0,0 +1,100 @@
+import { WebPlugin } from '@capacitor/core';
+import {
+  ClipboardPlugin,
+  ClipboardWrite,
+  ClipboardReadResult,
+} from './definitions';
+
+declare var navigator: any;
```
dwieeb, line 8: The TypeScript dom lib has this definition.

```suggestion
```

clipboard/src/web.ts
```diff
@@ -0,0 +1,100 @@
...
+      if (!navigator.clipboard.writeText) {
+        return Promise.reject(
+          'Writting to clipboard not supported in this browser',
+        );
+      }
+      await navigator.clipboard.writeText(
+        options.string !== undefined ? options.string : options.url,
+      );
+    } else if (options.image) {
+      if (!navigator.clipboard.write) {
```
dwieeb, line 34: I think we need to do more here. [`navigator.clipboard.write`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write) seems to have more browser adoption than [`ClipboardItem`](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem).

```suggestion
      if (!navigator.clipboard.write || !ClipboardItem) {
```

clipboard/src/web.ts
```diff
@@ -0,0 +1,100 @@
...
+  constructor() {
+    super({
+      name: 'Clipboard',
+      platforms: ['web'],
+    });
+  }
+
+  async write(options: ClipboardWrite): Promise<void> {
+    if (!navigator.clipboard) {
+      return Promise.reject('Clipboard API not available in this browser');
```
dwieeb, line 21: Within `async` functions, we prefer to use `throw new Error(msg)` syntax, which is equivalent to `return Promise.reject(new Error(msg))`. 

Additionally, when the current browser is unsupported, we want to throw the new `UnsupportedBrowserException` error. See https://github.com/ionic-team/capacitor-plugins/pull/1/files#diff-cfdb994cb3dbedf92d7eab756a4de2ccR17-R19.

clipboard/src/web.ts
```diff
@@ -0,0 +1,100 @@
...
+} from './definitions';
+
+declare var navigator: any;
+declare var ClipboardItem: any;
+
+export class ClipboardWeb extends WebPlugin implements ClipboardPlugin {
+  constructor() {
+    super({
+      name: 'Clipboard',
+      platforms: ['web'],
```
dwieeb, line 15: This is no longer needed.

```suggestion
```

clipboard/src/web.ts
```diff
@@ -0,0 +1,100 @@
...
+        const blob = await (await fetch(options.image)).blob();
+        const clipboardItemInput = new ClipboardItem({ [blob.type]: blob });
+        await navigator.clipboard.write([clipboardItemInput]);
+      } catch (err) {
+        return Promise.reject('Failed to write image');
+      }
+    } else {
+      return Promise.reject('Nothing to write');
+    }
+    return Promise.resolve();
```
dwieeb, line 47: This isn't necessary in `async` functions.

```suggestion
```

clipboard/src/web.ts
```diff
@@ -0,0 +1,100 @@
...
+        );
+      }
+      return this.readText();
+    } else {
+      try {
+        const clipboardItems = await navigator.clipboard.read();
+        const type = clipboardItems[0].types[0];
+        const clipboardBlob = await clipboardItems[0].getType(type);
+        const data = await this._getBlobData(clipboardBlob, type);
+        return Promise.resolve({ value: data, type });
```
dwieeb, line 67: Not necessary in `async` functions.

```suggestion
        return { value: data, type };
```

clipboard/src/web.ts
```diff
@@ -0,0 +1,100 @@
...
+      } catch (err) {
+        return this.readText();
+      }
+    }
+  }
+
+  private async readText() {
+    const text = await navigator.clipboard.readText();
+    return Promise.resolve({ value: text, type: 'text/plain' });
```
dwieeb, line 76:
```suggestion
    return { value: text, type: 'text/plain' };
```

clipboard/android/src/main/java/com/capacitorjs/plugins/clipboard/Clipboard.java
carlpoole: > I know you didn't write the web part, but we probably shouldn't just copy/paste these implementations. They should be evaluated ad hoc and modernized.

Yeah, thanks for your help with this. This isn't my area of expertise

clipboard/android/src/main/java/com/capacitorjs/plugins/clipboard/Clipboard.java
carlpoole: It's called twice in this class so its prob not a big deal but I did it and it looks cleaner so lets do it.

clipboard/android/src/main/java/com/capacitorjs/plugins/clipboard/Clipboard.java
carlpoole: Not exactly like this. I stole this code from another method. I am adjusting it to how its usually done and so its a static const.

clipboard/ios/Plugin/Clipboard.swift
```diff
@@ -0,0 +1,67 @@
...
+        return [:]
+    }
+
+    private func getCleanData(_ data: String) -> String {
+        let dataParts = data.split(separator: ",")
+        if let part = dataParts.last {
+            return String(part)
+        }
+        return data
+    }
```
jcesarmobile, line 66: this should be in a Capacitor's helper class, there is a similar method on Filesystem plugin

clipboard/ios/PluginTests/PluginTests.swift
```diff
@@ -0,0 +1,35 @@
...
+            "value": value
+        ], success: { (result, _) in
+            let resultValue = result!.data["value"] as? String
+            XCTAssertEqual(value, resultValue)
+        }, error: { (_) in
+            XCTFail("Error shouldn't have been called")
+        })
+
+        plugin.echo(call!)
+    }
```
jcesarmobile, line 34:
```suggestion
```

remove echo test

clipboard/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@capacitor/clipboard",
+  "version": "0.0.1",
+  "description": "The Clipboard API enables copy and pasting to/from the clipboard.",
+  "main": "dist/plugin.js",
+  "module": "dist/esm/index.js",
+  "types": "dist/esm/index.d.ts",
+  "scripts": {
+    "lint": "npm run prettier -- --check && npm run swiftlint -- lint",
+    "fmt": "npm run prettier -- --write && npm run swiftlint -- autocorrect",
```
jcesarmobile, line 10:
```suggestion
    "fmt": "npm run prettier -- --write && npm run swiftlint -- autocorrect --format",
```
missing --format param

clipboard/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "@rollup/plugin-node-resolve": "^8.1.0",
+    "prettier": "^2.0.5",
+    "prettier-plugin-java": "^0.8.0",
+    "rimraf": "^3.0.0",
+    "rollup": "^2.21.0",
+    "swiftlint": "^1.0.1",
+    "typescript": "~3.9.7"
+  },
+  "peerDependencies": {
+    "@capacitor/core": "^2.4.0"
```
jcesarmobile, line 35:
```suggestion
    "@capacitor/core": "^3.0.0-alpha.0"
```

we shouldn't use 2.x as peerDependency, should be the alpha or next (suggested alpha because the only merged plugin used that)

package.json
```diff
@@ -5,6 +5,7 @@
     "build": "lerna run build",
     "watch": "lerna run --parallel --stream watch",
     "lint": "lerna run lint",
+    "fmt": "lerna run fmt",
```
jcesarmobile, line 8:
```suggestion
```
I don't think this belongs to this PR

jcesarmobile (changes requested):
Added a few comments.

Also, I think you said the colors.xml, strings.xml and styles.xml were not needed and removed them from the template, but they are still here. Should be removed.

package.json
carlpoole: @dwieeb added this in the screen reader plugin and when I created this one I copied some of the same changes. Is this needed? https://github.com/ionic-team/capacitor-plugins/pull/1/files

clipboard/ios/PluginTests/PluginTests.swift
carlpoole: Do we also want to leave in the android tests? I noticed we were doing that. 

https://github.com/ionic-team/capacitor-plugins/blob/6356931b776b1e8cd0a509f79cdf37841d1037a0/clipboard/android/src/androidTest/java/com/getcapacitor/android/ExampleInstrumentedTest.java

https://github.com/ionic-team/capacitor-plugins/blob/6356931b776b1e8cd0a509f79cdf37841d1037a0/clipboard/android/src/test/java/com/getcapacitor/ExampleUnitTest.java

jcesarmobile: at the moment we are leaving them, but they are pointless, so we could remove them
but at least they pass, on iOS they fail because the echo method was deleted, and also the PluginTests.swift has changed a lot since you created this and it's out of date with the template

package.json
jcesarmobile: should probably be a separate PR with just that change, he should had probably done that and merged, not a big problem, but only needed once, when your PR or his PR get merged the other will conflict

clipboard/ios/Plugin/Clipboard.swift
carlpoole: Discussed with Ian and Dan. We're not sure its a great idea to move this into core. It's only used in 2 plugins and its a fairly simple string manipulation function. Moving into core would require creating a new helper class just for this in iOS and adds another dependency on the core we don't think is worth it.

clipboard/src/web.ts
```diff
@@ -0,0 +1,107 @@
...
+        return this.readText();
+      }
+    } else {
+      return this.readText();
+    }
+  }
+
+  private async readText() {
+    if (!('readText' in navigator.clipboard)) {
+      throw new Error('Reading from clipboard not supported in this browser');
```
dwieeb, line 71: This should use `UnsupportedBrowserException`

clipboard/src/web.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    const text = await navigator.clipboard.readText();
+    return { value: text, type: 'text/plain' };
+  }
+
+  private async writeText(text: string) {
+    if (!('writeText' in navigator.clipboard)) {
+      throw new Error('Writting to clipboard not supported in this browser');
```
dwieeb, line 80: This should use `UnsupportedBrowserException` (and fix the typo please)

clipboard/src/web.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    await navigator.clipboard.writeText(text);
+  }
+
+  private _getBlobData(clipboardBlob: Blob, type: string) {
```
dwieeb, line 86: Writing the return type explicitly is a good habit to get into, as changes in the body of the function may implicitly change the return type.

```suggestion
  private _getBlobData(clipboardBlob: Blob, type: string): Promise<string> {
```

clipboard/src/definitions.ts
```diff
@@ -0,0 +1,65 @@
...
+   */
+  read(): Promise<ClipboardReadResult>;
+}
+
+/**
+ * Represents the data to be written to the clipboard.
+ *
+ * @since 1.0.0
+ */
+export interface ClipboardWriteOptions {
```
dwieeb, line 28: Each property should also have `@since 1.0.0`

clipboard/src/definitions.ts
```diff
@@ -0,0 +1,65 @@
...
+   */
+  label?: string;
+}
+
+/**
+ * Represents the data read from the clipboard.
+ *
+ * @since 1.0.0
+ */
+export interface ClipboardReadResult {
```
dwieeb, line 55: Each property should also have `@since 1.0.0`

clipboard/ios/Plugin/Clipboard.swift
jcesarmobile: I mean putting it, in example in `CAPFile.swift` where we have file related helper methods.
Don't understand the "Add another dependency on the core" part, what dependency would it add? It would be a static method plugins can call.

carlpoole: Related: https://github.com/ionic-team/capacitor/pull/3457

carlpoole: Dependency probably wasn't the best choice of words. I meant that it adds another coupling with the core when the code could just exist in the plugin itself and not depend on the core being there necessarily.

jcesarmobile (changes requested):
At the moment if there is no clipboard information, on Android it returns an empty string with `"text/plain"` type.
On iOS, if empty, it doesn't return anything, nor rejects, so if users are using await the app will hang forever.
Both implementations should work the same way, either reject if there is no data to read, or both return an empty string with `"text/plain"` type.

Also, there is now an extension available to get the data from the data url instead of using `lastCommaSeparatedValue`
see https://github.com/ionic-team/capacitor/pull/3474 for more information

carlpoole:
> At the moment if there is no clipboard information, on Android it returns an empty string with `"text/plain"` type.
> On iOS, if empty, it doesn't return anything, nor rejects, so if users are using await the app will hang forever.
> Both implementations should work the same way, either reject if there is no data to read, or both return an empty string with `"text/plain"` type.
> 
> Also, there is now an extension available to get the data from the data url instead of using `lastCommaSeparatedValue`
> see [ionic-team/capacitor#3474](https://github.com/ionic-team/capacitor/pull/3474) for more information

I made both platforms reject if clipboard is empty with appropriate message. I updated to use the new Data extension and it worked in the context of the test app project (was pointing to main locally) but CI is failing here because its not yet released into Capacitor.

clipboard/android/src/main/java/com/capacitorjs/plugins/clipboard/Clipboard.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public void write(String label, ContentType contentType, String content) {
+        ClipData data = null;
+        switch (contentType) {
+            case STRING:
+            case IMAGE:
+            case URL:
+                data = ClipData.newPlainText(label, content);
+                break;
+        }
```
jcesarmobile, line 35:
```suggestion
    public void write(String label, String content) {
        ClipData data = ClipData.newPlainText(label, content);
```
We don't need the switch if all 3 do the same, and we don't need the contentType on the write function neither

clipboard/android/src/main/java/com/capacitorjs/plugins/clipboard/ClipboardPlugin.java
```diff
@@ -0,0 +1,52 @@
...
+        String urlVal = call.getString("url");
+        String label = call.getString("label");
+
+        if (strVal != null) {
+            implementation.write(label, Clipboard.ContentType.STRING, strVal);
+        } else if (imageVal != null) {
+            implementation.write(label, Clipboard.ContentType.IMAGE, imageVal);
+        } else if (urlVal != null) {
+            implementation.write(label, Clipboard.ContentType.URL, urlVal);
+        }
```
jcesarmobile, line 31:
```suggestion
        if (strVal != null) {
            implementation.write(label, strVal);
        } else if (imageVal != null) {
            implementation.write(label, imageVal);
        } else if (urlVal != null) {
            implementation.write(label, urlVal);
        }
```
The content type is not needed, we always write as text

jcesarmobile (changes requested):
At the moment the plugin write method can reject or resolve on iOS, but on Android always resolves.

Since the implementation can fail to write on both (if data is null, or clipboard is null, or failed to encode the image, or the type was not correct), I think the write method on the implementation class should return true/false or throw an error and the plugin should reject or resolve accordingly.

Also, can you remove the lock files from other plugins?

clipboard/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "compilerOptions": {
+    "allowUnreachableCode": false,
+    "declaration": true,
+    "esModuleInterop": true,
+    "lib": [
+      "dom",
+      "es2015"
```
dwieeb, line 8:
```suggestion
      "dom"
```

clipboard/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "module": "es2015",
+    "moduleResolution": "node",
+    "noFallthroughCasesInSwitch": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "outDir": "dist/esm",
+    "pretty": true,
+    "sourceMap": true,
+    "strict": true,
+    "target": "es2015"
```
dwieeb, line 19: New plugins now target ES2017

```suggestion
    "target": "es2017"
```

clipboard/src/web.ts
```diff
@@ -0,0 +1,126 @@
...
+const Clipboard = new ClipboardWeb();
+
+export { Clipboard };
+
+declare global {
+  interface Clipboard {
+    read(): Promise<any>;
+    write(data: any[]): Promise<any>;
+  }
+}
```
dwieeb, line 126: Usually stuff with `declare` is put at the top of the file (under imports is fine) because it's meta stuff

jcesarmobile (changes requested):
On iOS, write implementation still doesn't return if it failed or not, only rejects if there was no string, url or image, but should reject if the image fails to encode, or if the url fails to build, or if the content type is unknown

carlpoole:
I refined the iOS plugin as requested, please confirm @jcesarmobile @ikeith if any further changes should be made before merging

dwieeb (approved):

clipboard/ios/Plugin/Clipboard.swift
```diff
@@ -0,0 +1,78 @@
...
+        private let message: String
+
+        var localizedDescription: String {
+            return message
+        }
+
+        init(_ message: String) {
+            self.message = message
+        }
+    }
```
ikeith, line 22: This is actually a minor anti-pattern because you're performing a type erasure on the error. The only thing that distinguishes one error from another is the message, which is private and therefore invisible. Unless you need to store extra fields of information (in which case a struct is appropriate), an enum is more expressive. Additionally, you're overwriting the existing implementation of `localizedDescription`. The proper way to supply custom values to that method is to conform to `LocalizedError` and overwrite `errorDescription`.
```suggestion
  enum ClipboardError: LocalizedError {
    case invalidURL, invalidImage
    
    public var errorDescription: String? {
        switch self {
        case .invalidURL:
            return "Unable to form URL"
        case .invalidImage:
            return "Unable to encode image"
        }
    }
}
```

clipboard/ios/Plugin/Clipboard.swift
```diff
@@ -0,0 +1,78 @@
...
+        case ContentType.string:
+            UIPasteboard.general.string = content
+            return .success(())
+        case ContentType.url:
+            if let url = URL(string: content) {
+                UIPasteboard.general.url = url
+                return .success(())
+            } else {
+                let errMsg = "Unable to form URL"
+                CAPLog.print(errMsg)
```
ikeith, line 35: It would be simpler to adopt the suggested enum, then return `ClipboardError.invalidURL` here. And let the calling class log the message to CAPLog.

clipboard/ios/Plugin/Clipboard.swift
```diff
@@ -0,0 +1,78 @@
+import Foundation
+import Capacitor
+
+public class Clipboard: NSObject {
```
ikeith, line 4: If this is no longer exposed to Obj-C, it does not need to inherit from NSObject. Doing so means it can be instantiated from Obj-C but none of its methods can be called.
```suggestion
public class Clipboard {
```

ikeith (changes requested):

PR #8, opened 2020-08-04
Network

ikeith:
Work in progress

- [x] Remove plugin from core repo
- [x] Android
- [x] Web
- [x] iOS


network/ios/Plugin/Reachability.swift
```diff
@@ -0,0 +1,322 @@
...
+    let reachability = Unmanaged<Reachability>.fromOpaque(info).takeUnretainedValue()
+    reachability.reachabilityChanged()
+}
+
+public class Reachability {
+
+    public typealias NetworkReachable = (Reachability) -> Void
+    public typealias NetworkUnreachable = (Reachability) -> Void
+
+    @available(*, unavailable, renamed: "Conection")
```
dwieeb, line 55:
```suggestion
    @available(*, unavailable, renamed: "Connection")
```

network/src/definitions.ts
```diff
@@ -0,0 +1,28 @@
...
+   */
+  removeAllListeners(): void;
+}
+
+export interface NetworkStatus {
+  connected: boolean;
+  connectionType: 'wifi' | 'cellular' | 'none' | 'unknown';
+}
+
+export type NetworkStatusChangeCallback = (status: NetworkStatus) => void;
```
dwieeb, line 28: Please fully document this file (see [Storage plugin](https://github.com/ionic-team/capacitor-plugins/pull/15/files#diff-5f68b89cfcbd395c3f9ecef3bbdf9852) for an example). Feel free to rewrite the sentences, too. 

The JSDoc you write will be available for devs when they use these APIs in VS Code. Additionally, we can use it to autogenerate documentation for this plugin.

network/src/definitions.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/**
+ * Represents the state and type of the network connection.
+ *
+ * @since 1.0.0
+ */
+export interface NetworkStatus {
+  connected: boolean;
+  connectionType: 'wifi' | 'cellular' | 'none' | 'unknown';
```
dwieeb, line 36:
```suggestion
  /**
   * Whether there is an active connection or not.
   *
   * @since 1.0.0
   */
  connected: boolean;

  /**
   * The type of network connection currently in use.
   *
   * If there is no active network connection, `connectionType` will be `'none'`.
   *
   * @since 1.0.0
   */
  connectionType: 'wifi' | 'cellular' | 'none' | 'unknown';
```

network/rollup.config.js
```diff
@@ -0,0 +1,21 @@
+import nodeResolve from '@rollup/plugin-node-resolve';
+
+export default {
+  input: 'dist/esm/index.js',
+  output: {
+    file: 'dist/plugin.js',
+    format: 'iife',
+    name: 'capacitorPlugin', // TODO: change this
```
carlpoole, line 8: We need to name plugins here according to Dans checklist. Came here to see what you called yours so I could follow the same convention :)

network/android/src/main/java/com/capacitorjs/plugins/network/Network.java
```diff
@@ -0,0 +1,85 @@
...
+     * Interface for callbacks when network status changes.
+     */
+    interface NetworkStatusChangeListener {
+        void onNetworkStatusChanged();
+    }
+
+    @Nullable
+    private NetworkStatusChangeListener statusChangeListener;
+
+    private ConnectivityManager cm;
```
carlpoole, line 25: Let's rename this `connectivityManager` or something, for readability.

network/src/web.ts
```diff
@@ -0,0 +1,77 @@
...
+        window.navigator.mozConnection ||
+        window.navigator.webkitConnection;
+      let connectionType = connection
+        ? connection.type || connection.effectiveType
+        : 'wifi';
+
+      resolve({
+        connected: connected,
+        connectionType: connected ? connectionType : 'none',
+      } as NetworkStatus);
```
dwieeb, line 32: Using `x as X` is a forceful cast. To avoid it, assign object literals to a variable with an explicit type.

```suggestion
      const status: NetworkStatus = {
        connected: connected,
        connectionType: connected ? connectionType : 'none',
      }

      resolve(status);
```

network/src/web.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  getStatus(): Promise<NetworkStatus> {
+    return new Promise((resolve, reject) => {
+      if (!window.navigator) {
+        reject('Network info not available');
+        return;
+      }
+
+      let connected = window.navigator.onLine;
```
dwieeb, line 20: Prefer `const` vs `let` when the variable isn't reassigned.

```suggestion
      const connected = window.navigator.onLine;
```

network/src/web.ts
```diff
@@ -0,0 +1,77 @@
...
+      let connection =
```
dwieeb, line 21:
```suggestion
      const connection =
```

network/src/web.ts
```diff
@@ -0,0 +1,77 @@
...
+        window.navigator.connection ||
+        window.navigator.mozConnection ||
+        window.navigator.webkitConnection;
+      let connectionType = connection
```
dwieeb, line 25:
```suggestion
      const connectionType = connection
```

network/src/web.ts
```diff
@@ -0,0 +1,77 @@
...
+  listenerFunction: any = null;
+
+  constructor() {
+    super({ name: 'Network' });
+  }
+
+  getStatus(): Promise<NetworkStatus> {
+    return new Promise((resolve, reject) => {
+      if (!window.navigator) {
+        reject('Network info not available');
```
dwieeb, line 16: This should use `UnsupportedBrowserException` with a specific message about what browser API is missing, see https://github.com/ionic-team/capacitor-plugins/blob/main/screen-reader/src/web.ts#L17-L19 for an example.

jcesarmobile:
I think we should not be shipping the Reachability.swift file, it's a modified and out of date version of https://github.com/ashleymills/Reachability.swift, we should instead use https://github.com/ashleymills/Reachability.swift as a dependency of the plugin.

ikeith:
While we can update the copy of Reachability to the latest version, I think it would be better to keep it in this repo. If we added it as a dependency, it would be the only external one (Capacitor requires CapacitorCordova but CapacitorCordova is internal and it does not have any dependencies). That complicates getting the project to build directly (without cocoapods) as well as bringing all the usual downsides of a dependency.  And offers very little benefit - it is a small, stable project that has not had a code change in 10 months.

jcesarmobile:
I see a lot of benefits
First of all, the file is not an exact copy, so it's hard to maintain
Second, the file has the copyright notice from the other project
Third, would be good to have a plugin with an example on how to use an external dependency so plugin developers can see how to do it in their own plugin.
Fourth, we don't need to maintain the file anymore, nor add swiftlint rules on it as its code won't be seen be swiftlint.

> That complicates getting the project to build directly (without cocoapods) as well as bringing all the usual downsides of a dependency.
I don't see how it complicates things, Plugins projects are CocoaPods projects, and as I said, I see the dependency as a good thing instead of a downside. The verify command that Dan added builds the workspace, so that should be good for CocoaPods projects.

> And offers very little benefit - it is a small, stable project that has not had a code change in 10 months.
Last PR was merged 29 days ago, what you mean is probably last release, still, our version is even older.

ikeith:
> First of all, the file is not an exact copy, so it's hard to maintain
> Fourth, we don't need to maintain the file anymore, nor add swiftlint rules on it as its code won't be seen be swiftlint.
I think maintenance is a concern. I have imported the latest version of the file as of today and added it without modification. It was modified in the past for linting reasons and (I think) to modify the enum(s). But the way it is wrapped now makes that unnecessary. I have also added a nested SwiftLint config to ignore it so we it be updated without worrying about forking.

> Last PR was merged 29 days ago, what you mean is probably last release, still, our version is even older.
Its last code change was 10 months ago, any more recent changes are to the project/build settings and don't matter when it's directly included.

> Third, would be good to have a plugin with an example on how to use an external dependency so plugin developers can see how to do it in their own plugin.
That could definitely be useful but that should be cross-platform to provide the most benefit. This is iOS-specific. We might consider making a new demo plugin specifically for that purpose as documentation.

jcesarmobile (approved):
All android plugins already have external dependencies, so there is no need to do a plugin that has external dependencies for both iOS and Android as an example.

I think adding a file that is not ours, with a license that is not ours is a bad practice, and adding a swiftlint exception for not processing it when we have changed the whole code of capacitor to pass follow some rules is even worse.

But approving since I'm not going to change your mind and I don't want to block it.

ikeith:
> I think adding a file that is not ours, with a license that is not ours is a bad practice, and adding a swiftlint exception for not processing it when we have changed the whole code of capacitor to follow some rules is even worse.

After mulling it over, I've come to the conclusion that I agree. I have a lot of hesitation about adding an external dependency to Capacitor itself but that does not apply to the plugins. And since I updated the wrapping code to use an unmodified version of Reachability, it makes sense to reference it instead of copying it in.

jcesarmobile (changes requested):
The web implementation should be reworked, at the moment it can return `NetworkInformation.effectiveType`, which can have `slow-2g`, `2g`, `3g`, or `4g` values, or `NetworkInformation.type`, which can have `bluetooth`, `cellular`, `ethernet`, `none`, `wifi`, `wimax`, `other` or `unknown` values (according to https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation)
Or if we don't want to rework the implementation, then the types should be modified to allow all those values.

It was reported on https://github.com/ionic-team/capacitor/discussions/2444, but was wrongly converted to discussion and missed until now.

For iOS, I think the NetworkPlugin.swift class should not contain traces of Reachability, the point of having the implementation classes is to wrap any external API or SDK and the plugin class should know nothing about it.

For Android I made a comment related to the permission check.

And also made two suggested changes on the podspec and Podfile to use a version range.

network/android/src/main/java/com/capacitorjs/plugins/network/NetworkPlugin.java
```diff
@@ -0,0 +1,107 @@
...
+     * @param call
+     */
+    @SuppressWarnings("MissingPermission")
+    @PluginMethod
+    public void getStatus(PluginCall call) {
+        if (hasRequiredPermissions()) {
+            call.resolve(getStatusJSObject(implementation.getNetworkStatus()));
+        } else {
+            call.reject(PERMISSION_NOT_SET);
+        }
```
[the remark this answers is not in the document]
ikeith: I may not be needed under normal conditions but it shouldn't cause any harm to leave it either. And it could provide more useful errors in the unlikely case that the build configuration really becomes messed up.

network/src/web.ts
```diff
@@ -0,0 +1,118 @@
...
+      case '3g':
+        result = 'cellular';
+        break;
+      case '4g':
+        result = 'wifi';
+        break;
+      default:
+        break;
+    }
+  }
```
ikeith, line 48: [`type`](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/type) and [`effectiveType`](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/effectiveType) reference

This is a best effort to map the possible values onto the ones we are documented to returned. Since the native side doesn't have all of these possibilities, it didn't make sense to return them as-is.

jcesarmobile (approved):
Looks good.

My only concern are the iOS tests. It tests that the connection is wifi, so it fails if run locally without internet connection.
I guess it won't fail on the CI since it needs to have connection to work and on simulators the connection is always wifi. 
I guess we can delete or change them in the future if something changes, and we should have in mind that when testing locally we have to be connected or they will fail.

network/src/web.ts
```diff
@@ -0,0 +1,118 @@
...
+  constructor() {
+    super({ name: 'Network' });
+  }
+
+  async getStatus(): Promise<NetworkStatus> {
+    if (!window.navigator) {
+      throw new UnsupportedBrowserException(
+        'Browser does not support the Network Information API',
+      );
+    }
```
jcesarmobile, line 62: window.navigator is always defined, maybe it's better to just check if onLine is defined in navigator

```suggestion
    if ('onLine' in navigator) {
      throw new UnsupportedBrowserException(
        'Browser does not support the Network Information API',
      );
    }
```

But onLine has been available for a long long time, so maybe we can even remove the check
https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine

network/src/web.ts
```diff
@@ -0,0 +1,118 @@
...
+} from '@capacitor/core';
+import {
+  NetworkPlugin,
+  NetworkStatus,
+  NetworkStatusConnectionType,
+} from './definitions';
+
+declare var window: any;
+
+function translateConnection(connection: any): NetworkStatusConnectionType {
```
jcesarmobile, line 14:
```suggestion
function translateConnection(): NetworkStatusConnectionType {
  const connection =
    window.navigator.connection ||
    window.navigator.mozConnection ||
    window.navigator.webkitConnection;
```

network/src/web.ts
```diff
@@ -0,0 +1,118 @@
...
+        'Browser does not support the Network Information API',
+      );
+    }
+
+    const connected = window.navigator.onLine;
+    const connectionType = translateConnection(
+      window.navigator.connection ||
+        window.navigator.mozConnection ||
+        window.navigator.webkitConnection,
+    );
```
jcesarmobile, line 69:
```suggestion
    const connectionType = translateConnection();
```

network/src/web.ts
```diff
@@ -0,0 +1,118 @@
...
+  addListener(
+    eventName: 'networkStatusChange',
+    listenerFunc: (status: NetworkStatus) => void,
+  ): PluginListenerHandle {
+    const thisRef = this;
+    const connectionType = translateConnection(
+      window.navigator.connection ||
+        window.navigator.mozConnection ||
+        window.navigator.webkitConnection,
+    );
```
jcesarmobile, line 88:
```suggestion
    const connectionType = translateConnection();
```

jcesarmobile (changes requested):
made some minor code suggestions on web code

jcesarmobile (approved):
Looks good to me.

But @dwieeb, can you review the `web.ts` as it uses `declare var window: any;` and you told Carl in another PR to not use something similar?

network/src/web.ts
ikeith: After accepting this change, I have reverted it. `('onLine' in navigator)` would always fail and report that the API was not available.

network/src/web.ts
```diff
@@ -0,0 +1,114 @@
...
+  WebPlugin,
+  PluginListenerHandle,
+} from '@capacitor/core';
+import {
+  NetworkPlugin,
+  NetworkStatus,
+  NetworkStatusConnectionType,
+} from './definitions';
+
+declare var window: any;
```
dwieeb, line 12: We can be more specific about which types we're manipulating. We could go further and give `connection` a specific type as well, but there's only so many minutes in a day.

```suggestion
declare global {
  interface Navigator {
    connection: any;
    mozConnection: any;
    webkitConnection: any;
  }
}
```

PR #9, opened 2020-08-05
Text Zoom

dwieeb:
New plugin!

TODO:
- [x] Android
- [x] iOS

see https://github.com/ionic-team/capacitor-testapp/pull/3


text-zoom/ios/Plugin/TextZoom.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+@objc public class TextZoom: NSObject {
+    static var baseFontSize: Double = 17
+
+    @objc public func getPreferred() -> Double {
```
ikeith, line 6: `preferredFontSize` or something similar would be a more idiomatic name for internal (i.e. not exposed to JS) methods like this.

text-zoom/android/src/main/java/com/capacitorjs/plugins/textzoom/TextZoomPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+import android.webkit.WebView;
+import com.getcapacitor.JSObject;
+import com.getcapacitor.NativePlugin;
+import com.getcapacitor.Plugin;
+import com.getcapacitor.PluginCall;
+import com.getcapacitor.PluginMethod;
+
+@NativePlugin(name = "TextZoom")
+public class TextZoomPlugin extends Plugin {
+    private TextZoom tz;
```
carlpoole, line 15: I like these short names in the plugins but it doesn't seem consistent to have `tz` and `webView` and `activity` etc. I'm thinking it would be cleaner to use full words. I think in a couple we used "implementation" so far but `textZoom` is fine I think?

dwieeb: Yeah I mostly used implementation because it was easier and a good way to show the plugin author intent. `tz` was laziness, good call

studioromeo:
@imhoffd hey this is a bit of a long shot but were you able to use CSS to prevent some (but not all) text from resizing on android? I've been playing around with a few things like `text-size-adjust` but that didn't have any effect.

The docs here from android don't give much help on this either https://developer.android.com/reference/android/webkit/WebSettings#setTextZoom(int)

No worries if not but figured I'd ask incase you had.

PR #13, opened 2020-08-13
Geolocation

carlpoole:
- [x] Android
- [x] iOS


geolocation/ios/Plugin/GeolocationPlugin.swift
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    @objc func clearWatch(_ call: CAPPluginCall) {
+        guard let callbackId = call.getString("id") else {
+            CAPLog.print("Must supply id")
+            return
+        }
+        let savedCall = bridge?.getSavedCall(callbackId)
+        if savedCall != nil {
+            bridge?.releaseCall(savedCall!)
```
ikeith, line 70: The check can be combined into a single line to remove the force unwrap.
```suggestion
        if let savedCall = bridge?.getSavedCall(callbackId) {
            bridge?.releaseCall(savedCall)
```

geolocation/ios/Plugin/GeolocationPlugin.swift
```diff
@@ -0,0 +1,200 @@
...
+                self.locationManager.desiredAccuracy = kCLLocationAccuracyBestForNavigation
+            } else {
+                self.locationManager.desiredAccuracy = kCLLocationAccuracyThreeKilometers
+            }
+
+            self.locationManager.startUpdatingLocation()
+        }
+    }
+
+    @objc func clearWatch(_ call: CAPPluginCall) {
```
ikeith, line 63: Shouldn't this remove the relevant call from `callQueue`?

geolocation/ios/Plugin/GeolocationPlugin.swift
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    @objc func watchPosition(_ call: CAPPluginCall) {
+        call.save()
+        callQueue[call.callbackId] = .watch
+
+        DispatchQueue.main.async {
+            self.locationManager.delegate = self
+            self.locationManager.requestWhenInUseAuthorization()
```
ikeith, line 51: watchPosition should check the current permission status first. According [to the documentation for `requestWhenInUseAuthorization`](https://developer.apple.com/documentation/corelocation/cllocationmanager/1620562-requestwheninuseauthorization), the method does not trigger any callbacks if the status is not `notDetermined`. So if it has been denied, no callbacks will happen and it will just quietly do nothing.

geolocation/ios/Plugin/GeolocationPlugin.swift
```diff
@@ -0,0 +1,200 @@
...
+        var ret = JSObject()
+        var coords = JSObject()
+        coords["latitude"] = location.coordinate.latitude
+        coords["longitude"] = location.coordinate.longitude
+        coords["accuracy"] = location.horizontalAccuracy
+        coords["altitude"] = location.altitude
+        coords["altitudeAccuracy"] = location.verticalAccuracy
+        coords["speed"] = location.speed
+        coords["heading"] = location.course
+        ret["timestamp"] = NSNumber(value: Int((location.timestamp.timeIntervalSince1970 * 1000)))
```
ikeith, line 195: The `Int` doesn't need to be wrapped in an NSNumber in Swift. It will be automatically converted for the JSON serialization.

geolocation/ios/Plugin/GeolocationPlugin.swift
```diff
@@ -0,0 +1,200 @@
...
+        for (id, _) in callQueue {
+            if let call = bridge?.getSavedCall(id) {
+                call.reject(error.localizedDescription)
+            }
+        }
+    }
+
+    public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+        let removalQueue = callQueue.filter { $0.value == .singleUpdate }
+        callQueue = callQueue.filter { $0.value == .watch }
```
ikeith, line 97: It would be safer to test the inverse of the first predicate, otherwise calls could be lost if enum cases are ever added in the future. 
```suggestion
        callQueue = callQueue.filter { $0.value != .singleUpdate }
```

ikeith (changes requested):

geolocation/android/src/main/java/com/capacitorjs/plugins/geolocation/GeolocationPlugin.java
```diff
@@ -0,0 +1,201 @@
...
+import com.getcapacitor.annotation.CapacitorPlugin;
+import com.getcapacitor.annotation.Permission;
+import java.util.HashMap;
+import java.util.Map;
+
+@CapacitorPlugin(
+    name = "Geolocation",
+    permissions = {
+        @Permission(permission = Manifest.permission.ACCESS_COARSE_LOCATION, alias = "location"),
+        @Permission(permission = Manifest.permission.ACCESS_FINE_LOCATION, alias = "location")
```
dwieeb, line 18: Why do these share an alias? App devs may want to differentiate between the two, right?

geolocation/android/src/main/java/com/capacitorjs/plugins/geolocation/GeolocationPlugin.java
```diff
@@ -0,0 +1,201 @@
...
+    },
+    permissionRequestCode = GeolocationPlugin.GEOLOCATION_REQUEST_PERMISSIONS
+)
+public class GeolocationPlugin extends Plugin {
+
+    public static final int GEOLOCATION_REQUEST_PERMISSIONS = 7179;
```
dwieeb, line 24: Was `7179` copied from somewhere? I think we should try to keep the official request codes in the 90xx range.

geolocation/src/definitions.ts
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Accuracy level of the latitude and longitude coordinates in meters
+     *
+     * @since 1.0.0
+     */
+    accuracy: number;
+
+    /**
+     * Accuracy level of the altitude coordinate in meters (if available)
+     *
```
dwieeb, line 96: We can be more specific here

```suggestion
     * Accuracy level of the altitude coordinate in meters, if available.
     *
     * Available on all iOS versions and on Android 8.0+.
     *
```

geolocation/android/src/main/java/com/capacitorjs/plugins/geolocation/GeolocationPlugin.java
carlpoole: No I changed it from the original code so it didn't conflict with the codes in the core. I can change this to be back in the 90xx range if that's what we want

geolocation/android/src/main/java/com/capacitorjs/plugins/geolocation/GeolocationPlugin.java
carlpoole: Yes they might but as the plugin is currently written, these two permissions were both requested together. The plugins currently dont offer a way to differentiate between wither the developer wants to use coarse or fine location (in iOS or Android). If we want to look at allowing that we might want to plan it as future work. There are also changes to background location coming with Android 11 that need to be addressed in the future, potentially.

dwieeb (approved):

geolocation/ios/PluginTests/PluginTests.swift
```diff
@@ -0,0 +1,35 @@
...
+            "value": value
+        ], success: { (result, _) in
+            let resultValue = result!.data["value"] as? String
+            XCTAssertEqual(value, resultValue)
+        }, error: { (_) in
+            XCTFail("Error shouldn't have been called")
+        })
+
+        plugin.echo(call!)
+    }
```
jcesarmobile, line 34:
```suggestion
```
remove the test

ikeith (approved):

jcesarmobile (changes requested):
I've added some missing files and removed some unused files.
There are still a few missing files from the new iOS project structure that was changed after this PR was created, not sure if we should care about them, @ikeith?
Also iOS files are renamed in other plugins to not be just `Plugin.swift`

The implementation class should not return a JSObject since those are plugin related classes, should return a Location class, I've sent a PR to the branch so can be reviewed before merging into this branch https://github.com/ionic-team/capacitor-plugins/pull/89

geolocation/src/definitions.ts
```diff
@@ -0,0 +1,139 @@
...
+  checkPermissions(): Promise<GeolocationPermissionStatus>;
+
+  /**
+   * Request location permissions
+   *
+   * @since 1.0.0
+   */
+  requestPermissions(options: {
+    permissions: GeolocationPermissionType[];
+  }): Promise<GeolocationPermissionStatus>;
```
jcesarmobile, line 56: The only way of calling the function is like this `Geolocation.requestPermissions({permissions: ['location']})`, since permissions allows an array of `GeolocationPermissionType`, but `GeolocationPermissionType` only allows `'location'` value, so there is no point on using an array. And also, since only one alias can be checked, I don't think it's worth using options, just call it without parameters.
```suggestion
  requestPermissions(): Promise<GeolocationPermissionStatus>;
```

jcesarmobile:
I've added the rest of missing files and new xcode project structure, should be good now.
We will need a new alpha release to use the handlePermissions method

geolocation/android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,5 @@
+
+  <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+      package="com.capacitorjs.plugins.geolocation">
+  </manifest>
+
```
dwieeb, line 5:
```suggestion
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    package="com.capacitorjs.plugins.geolocation">
</manifest>
```

geolocation/src/web.ts
```diff
@@ -0,0 +1,82 @@
...
+          reject(err);
+        },
+        extend(
+          {
+            enableHighAccuracy: true,
+            timeout: 10000,
+            maximumAge: 0,
+          },
+          options,
+        ),
```
dwieeb, line 31: These should be equivalent, and then you can delete the `extend` utility.

```suggestion
        {
          enableHighAccuracy: true,
          timeout: 10000,
          maximumAge: 0,
          ...options,
        },
```

geolocation/src/definitions.ts
```diff
@@ -0,0 +1,152 @@
...
+    accuracy: number;
+
+    /**
+     * Accuracy level of the altitude coordinate in meters, if available.
+     *
+     * Available on all iOS versions and on Android 8.0+.
+     *
+     * @since 1.0.0
+     */
+    altitudeAccuracy: number | null;
```
dwieeb, line 94: I can't test because my iPhone hates the testapp right now, but are you sure these should be `number | null` and not `number | undefined`?

carlpoole: Added undefined to account for it being missing when Android < v8. In all other cases mobile returns some default value. Web can return null for the applicable values with null in the type.

dwieeb (approved):

geolocation/package.json
```diff
@@ -0,0 +1,81 @@
...
+  },
+  "devDependencies": {
+    "@capacitor/android": "^3.0.0-alpha.7",
+    "@capacitor/core": "^3.0.0-alpha.7",
+    "@capacitor/docgen": "^0.0.10",
+    "@capacitor/ios": "^3.0.0-alpha.7",
+    "@ionic/eslint-config": "^0.3.0",
+    "@ionic/prettier-config": "^1.0.1",
+    "@ionic/swiftlint-config": "^1.1.2",
+    "@rollup/plugin-node-resolve": "^9.0.0",
```
jcesarmobile, line 53:
```suggestion
```
not used anymore

geolocation/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "eslint": "^7.11.0",
+    "prettier": "^2.0.5",
```
jcesarmobile, line 55:
```suggestion
    "prettier": "~2.2.0",
```
we fixed to patch version for some reason

geolocation/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "prettier-plugin-java": "^0.8.3",
```
jcesarmobile, line 56:
```suggestion
    "prettier-plugin-java": "~1.0.0",
```
also fixed to patch version

geolocation/src/web.ts
```diff
@@ -0,0 +1,77 @@
...
+    return new Promise((resolve, reject) => {
+      navigator.geolocation.getCurrentPosition(
+        pos => {
+          resolve(pos);
+        },
+        err => {
+          reject(err);
+        },
+        {
+          enableHighAccuracy: true,
```
jcesarmobile, line 24:
```suggestion
          enableHighAccuracy: false,
```
the documented value for `enableHighAccuracy` is false

geolocation/src/web.ts
```diff
@@ -0,0 +1,77 @@
...
+  ): string {
+    const id = navigator.geolocation.watchPosition(
+      pos => {
+        callback(pos);
+      },
+      err => {
+        callback(null, err);
+      },
+      {
+        enableHighAccuracy: true,
```
jcesarmobile, line 45:
```suggestion
        enableHighAccuracy: false,
```
the documented value for `enableHighAccuracy` is `false`

jcesarmobile (changes requested):

geolocation/src/definitions.ts
```diff
@@ -0,0 +1,152 @@
+export type CallbackID = string;
+
+export type GeolocationState = PermissionState;
```
dwieeb, line 3: This should be imported from `@capacitor/core`, see https://github.com/ionic-team/capacitor-plugins/pull/33#discussion_r543776894

jcesarmobile:
There was [an issue](https://github.com/ionic-team/capacitor/issues/3789) reported on capacitor repo about getLocation method on iOS returning error while being prompted for the permission. It was fixed in [this PR](https://github.com/ionic-team/capacitor/pull/3802) and released on 2.4.3.
That or similar code should also be added here or the bug will be reintroduced.

jcesarmobile (approved):

ikeith (approved):

Issue #14, opened 2020-08-14
feat: allow to configure KEY_PREFIX of web implementation of 'Storage'


bogdbo:
Hello,

I'm talking about this file line 9 and 58
https://github.com/ionic-team/capacitor/blob/master/core/src/web/storage.ts

We already have an existing application and the current Web version of Storage will not read the existing values from local storage due to this KEY_PREFIX. 

I can see this is being used for `clear()` and `keys()` etc but right now it's breaking existing applications. I guess we can force KEY_PREFIX to an empty string for now but this should be improved in my opinion.

Thank you!

dwieeb:
@yacut For example on iOS, NativeStorage uses a custom UserDefaults group called `"NativeStorage"`, and we use the standard UserDefaults at the moment (with a key prefix). To me, it doesn't make sense to share their group, but I could be convinced otherwise. Any changes to this behavior would result in a breaking change for apps unless we add in migration logic. I would like to exclude the ability to extract values from NativeStorage from the scope of this issue and instead focus on the original request: the ability to change the key prefix. This is a completely reasonable request and will be addressed in Capacitor 3.

dwieeb:
And I just found out the Android implementation doesn't even use the prefix to begin with, so this will be very difficult to implement in a cross-platform way. We may need to provide a way to migrate data.

dwieeb:
I am going to rewrite the iOS implementation to use a non-shared [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults) and rewrite the web implementation to use IndexedDB. This will remove the need for a prefix, and instead promote the idea of a "storage group". There will be a way to migrate data from the Capacitor 2 implementation.

dwieeb:
Okay, the latest for this: If you use the storage group "NativeStorage", it will be backwards-compatible with `cordova-plugin-nativestorage`. For web this means you can access the root level of localStorage, fulfilling the original feature request in this issue. We'd like devs to migrate their data from NativeStorage, because the `clear()` of the NativeStorage plugin on web seems to be very destructive (but please correct me if I'm mistaken).

On iOS and web, the Capacitor Storage plugin uses the storage group (default of "CapacitorStorage") as a prefix to separate key/values from other data. On Android, it uses the group as an entirely separate SharedPreferences file.

[...]

PR #15, opened 2020-08-14
Storage

dwieeb:
- [ ] iOS
- [ ] Android
- [ ] Web
- [ ] https://github.com/ionic-team/capacitor-plugins/issues/14


storage/ios/Plugin/Storage.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+
+public struct StorageConfiguration {
+    var group = "CapacitorStorage"
+}
```
ikeith, line 5: If we're willing to drop Obj-C interoperability for this class, it would be cleaner and more idiomatic to change the group into an enum. That would shift the burden of checking for the magic "NativeStorage" string to the plugin class which is closer to where it belongs.

Also note the change to `let`, which should be preferred in this case because the group shouldn't be changed after initialization.
```suggestion
public struct StorageConfiguration {
    enum Group {
        case native, prefixed(String)
    }
    let group: Group
    
    init(for group: Group = .prefixed("CapacitorStorage")) {
        self.group = group
    }
}
```

storage/ios/Plugin/Storage.swift
```diff
@@ -0,0 +1,51 @@
...
+
+@objc public class Storage: NSObject {
+    private var configuration: StorageConfiguration
```
ikeith, line 8:
```suggestion
    private let configuration: StorageConfiguration
```

storage/ios/Plugin/Storage.swift
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    @objc func getKeys() -> [String] {
+        return keys.map { String($0.dropFirst(prefix.count)) }
+    }
+
+    private var prefix: String {
+        return configuration.group == "NativeStorage" ? "" : configuration.group + "."
+    }
```
ikeith, line 38: This should belong to the configuration. It can be added as a private extension to keep the implementation details internal. If adopting the enum pattern, it would look like this:
```suggestion
private extension StorageConfiguration {
    var prefixString: String {
        switch self.group {
        case .native:
            return ""
        case .prefixed(let prefix):
            return prefix + "."
        }
    }
}
```

storage/ios/Plugin/StoragePlugin.swift
```diff
@@ -0,0 +1,90 @@
...
+public class StoragePlugin: CAPPlugin {
+    private var storage = Storage(withConfiguration: StorageConfiguration())
+
+    @objc func configure(_ call: CAPPluginCall) {
+        let group = call.getString("group")
+        var configuration = StorageConfiguration()
+
+        if group != nil {
+            configuration.group = group!
+        }
```
ikeith, line 14: An `if let` would be cleaner and avoid the force unwrap
```suggestion
        if let group = group {
            configuration.group = group
        }
```

dwieeb: Thanks, I was wondering why it was upset at me.

storage/rollup.config.js
```diff
@@ -0,0 +1,21 @@
+import nodeResolve from '@rollup/plugin-node-resolve';
+
+export default {
+  input: 'dist/esm/index.js',
+  output: {
+    file: 'dist/plugin.js',
+    format: 'iife',
+    name: 'capacitorPlugin', // TODO: change this
```
jcesarmobile, line 8: you should name the plugin and remove the TODO

storage/ios/Plugin/Storage.swift
```diff
@@ -0,0 +1,64 @@
...
+
+    let group: Group
+
+    init(for group: Group = .named("CapacitorStorage")) {
+        self.group = group
+    }
+}
+
+public class Storage: NSObject {
```
ikeith, line 15: The `NSObject` should be removed. Swift structs (and enums with associated values) cannot be bridged to Obj-C so this class's designated initializer cannot be called from Obj-C code. As such, it's better to hide this class entirely and only make it available to Swift code.
```suggestion
public class Storage {
```

storage/rollup.config.js
ikeith:
```suggestion
    name: 'capacitorStorage',
```

storage/ios/Plugin/Storage.swift
dwieeb: Thanks, I think I saw that somewhere on the internet and wondered why I needed to do it

storage/ios/PluginTests/PluginTests.swift
```diff
@@ -0,0 +1,26 @@
...
+    func testEcho() {
+        // This is an example of a functional test case for a plugin.
+        // Use XCTAssert and related functions to verify your tests produce the correct results.
+
+        let implementation = Storage()
+        let value = "Hello, World!"
+        let result = implementation.echo(value)
+
+        XCTAssertEqual(value, result)
+    }
```
jcesarmobile, line 25:
```suggestion
```
Remove the echo test

PR #16, opened 2020-08-18
Browser

ikeith:
- [x] Remove plugin from capacitor https://github.com/ionic-team/capacitor/pull/3439
- [x] JS wrapper and implementation
- [x] iOS implementation
- [x] Android implementation
- [x] Test app integration https://github.com/ionic-team/capacitor-testapp/pull/9


browser/ios/PluginTests/PluginTests.swift
```diff
@@ -0,0 +1,26 @@
...
+    func testEcho() {
+        // This is an example of a functional test case for a plugin.
+        // Use XCTAssert and related functions to verify your tests produce the correct results.
+
+        let implementation = Browser()
+        let value = "Hello, World!"
+        let result = implementation.echo(value)
+
+        XCTAssertEqual(value, result)
+    }
```
jcesarmobile, line 25:
```suggestion
```

remove the echo test, we should probably add this to the todo list

lerna.json
```diff
@@ -4,7 +4,8 @@
     "motion",
     "screen-reader",
     "storage",
-    "text-zoom"
+    "text-zoom",
+    "browser"
```
jcesarmobile, line 8: I think we are putting them in alphabetical order, not by adding order, should be on top

browser/CapacitorBrowser.podspec
```diff
@@ -0,0 +1,18 @@
...
+  s.version = package['version']
+  s.summary = package['description']
+  s.license = package['license']
+  s.homepage = package['repository']['url']
+  s.author = package['author']
+  s.source = { :git => package['repository']['url'], :tag => s.version.to_s }
+  s.source_files = 'ios/Plugin/**/*.{swift,h,m,c,cc,mm,cpp}'
+  s.ios.deployment_target  = '11.0'
+  s.dependency 'Capacitor'
+  s.framework = 'SafariServices'
```
jcesarmobile, line 16: Is this needed? We don't have it in the Capacitor.podspec and the plugin works. 

In Objective-C, since iOS 7 you can use `@import SafariServices` instead of `#import <SafariServices/SafariServices.h>` and that autolinks the SafariServices framework. Looks like the Swift import works the same way.

browser/ios/PluginTests/PluginTests.swift
ikeith: Oops, good catch.

browser/CapacitorBrowser.podspec
ikeith: I don't know? I didn't notice that capacitor omitted it. AFAIK, any of the system frameworks are automatically linked by Xcode (unless you want them to be optionally linked); so it should build without it. But it seems like that dependency should still be documented?

jcesarmobile (changes requested):
I've not finish with the review yet, but I've added a few comments.

Also, I can't make the prefetch work, I've tested and works fine in the old code. Anyway, considering it's android only and I don't find it particularly helpful, maybe we can remove it?

Browser doesn't fire the events. We maybe can use `load` and `beforeunload` window events to fire the browser events, but only seems to work for same domain urls, so not sure if it's worth it.
But if not implemented, should be documented that they are not fired on web.

browser/src/web.ts
```diff
@@ -0,0 +1,34 @@
...
+    super({ name: 'Browser' });
+    this._lastWindow = null;
+  }
+
+  async open(options: BrowserOpenOptions): Promise<void> {
+    this._lastWindow = window.open(options.url, options.windowName || '_blank');
+  }
+
+  async prefetch(_options: BrowserPrefetchOptions): Promise<void> {
+    // Does nothing
```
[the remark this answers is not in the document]
dwieeb: I think so, or remove it since you don't think it's useful. We can add it back later

browser/src/definitions.ts
```diff
@@ -0,0 +1,124 @@
...
+   */
+  removeAllListeners(): void;
+}
+
+/**
+ * Represents the browser information passed to event listeners.
+ *
+ * @since 1.0.0
+ */
+export interface BrowserInfo {}
```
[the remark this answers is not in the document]
dwieeb: It looks like this is used for the `listenerFunc` parameter, but if we're not passing anything then we should just remove the parameter and this interface, yes. We can always add parameters to function types later without it being a breaking change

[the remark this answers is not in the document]
jcesarmobile: yeah, both listeners return an empty object at the moment, I'm saying that maybe we should return void instead as I've seen a few people confused about `BrowserInfo` not containing any info, so making it return void would be more clear that it doesn't return anything, but the native code should be updated too as it returns the empty object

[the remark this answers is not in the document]
dwieeb: I think we're saying the same thing. I'm saying we apply this change:
```diff
-listenerFunc: (info: BrowserInfo) => void
+listenerFunc: () => void
```

And then, yes. We should make sure the native code is using `resolve()` so it returns void, not an empty object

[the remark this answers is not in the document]
jcesarmobile: listeners are a bit different, they don't resolve or reject, just pass the result directly, at the moment an empty object, should return null instead

browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserPlugin.java
```diff
@@ -0,0 +1,116 @@
...
+        implementation.prefetch(urls);
+        call.resolve();
+    }
+
+    @Override
+    protected void handleOnResume() {
+        boolean ok = implementation.bindService();
+        if (!ok) {
+            Logger.error(getLogTag(), "Error binding to custom tabs service", null);
+        }
```
carlpoole, line 98:
```suggestion
        if (!implementation.bindService()) {
            Logger.error(getLogTag(), "Error binding to custom tabs service", null);
        }
```

browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserPlugin.java
```diff
@@ -0,0 +1,116 @@
...
+            call.reject("Must provide an array of URLs to prefetch");
+            return;
+        }
+
+        if (implementation.isPrepared()) {
+            call.reject("Browser session isn't ready yet");
+            return;
+        }
+
+        List<Uri> urls = new ArrayList();
```
carlpoole, line 79:
```suggestion
        List<Uri> urls = new ArrayList<>();
```

carlpoole: Small. Just ensures the ArrayList is created to contain type `Uri`.

carlpoole (approved):
Just minor cleanup

dwieeb (commented):
Can you copy in the following files from a plugin in `main`? There have been changes:
- `.gitignore`
- `.eslintignore`

ESLint should be set up in this plugin.

Please remove the Podfile.lock and package-lock.json from git for this plugin, we're ignoring them now

jcesarmobile:
Also missing the README.md and LICENSE files

jcesarmobile (changes requested):
Added a few comments

browser/ios/Plugin/BrowserPlugin.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
+import Capacitor
+
+@objc(CAPBrowserPlugin)
+public class CAPBrowserPlugin: CAPPlugin {
+    private let implementation = Browser()
+
+    @objc func open(_ call: CAPPluginCall) {
+        // validate the URL
```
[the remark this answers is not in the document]
ikeith: I would argue not; it's partly a separation of concerns and partly to avoid duplication. This checks if the user passed a valid URL, not what kind of URL it is. The specific protocols that are supported are an internal detail of the implementation. And it makes sense for the implementation to keep that check in case it ever gets used by some other code.

jcesarmobile (approved):

dwieeb (approved):

PR #19, opened 2020-08-21
Filesystem

jcesarmobile:
see https://github.com/ionic-team/capacitor/pull/3458 and https://github.com/ionic-team/capacitor-testapp/pull/10


filesystem/ios/PluginTests/PluginTests.swift
```diff
@@ -0,0 +1,26 @@
...
+        super.setUp()
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+    }
+
+    override func tearDown() {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+        super.tearDown()
+    }
+
+    func testEcho() {
```
carlpoole, line 16: I think the echo test needs to be removed

filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/Filesystem.java
```diff
@@ -0,0 +1,285 @@
...
+        return null;
+    }
+
+    /**
+     * Helper function to recursively delete a directory
+     *
+     * @param file The file or directory to recursively delete
+     * @throws IOException
+     */
+    public void deleteRecursively(File file) throws IOException {
```
carlpoole, line 242: I think with a sufficiently big enough file structure this could cause a stack overflow. One approach could be to convert this to use a Stack object so its on the heap instead and then iterate over the file addresses deleting them off that Stack?

filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/Filesystem.java
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    /**
+     * Helper function to recursively copy a directory structure (or just a file)
+     *
+     * @param src The source location
+     * @param dst The destination location
+     * @throws IOException
+     */
+    public void copyRecursively(File src, File dst) throws IOException {
```
carlpoole, line 262: I think same potential pitfall as deleteRecursively

filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/Filesystem.java
carlpoole: Could revisit it later as an optimization if nobody has encountered this yet

jcesarmobile: File objects are basically a path representation of the file, not the actual file, so they don't use a lot of memory. Considering this, I don't think a regular app usage could create a filesystem structure so big and complex that could lead to a stack overflow.

carlpoole: Yeah it's one of those things that can happen but will likely be pretty rare

filesystem/rollup.config.js
```diff
@@ -0,0 +1,21 @@
+import nodeResolve from '@rollup/plugin-node-resolve';
+
+export default {
+  input: 'dist/esm/index.js',
+  output: {
+    file: 'dist/plugin.js',
+    format: 'iife',
+    name: 'filesystemPlugin',
```
dwieeb, line 8:
```suggestion
    name: 'capacitorFilesystem',
```

filesystem/.gitignore
```diff
@@ -0,0 +1,64 @@
+# NPM renames .gitignore to .npmignore
+# In order to prevent that, we remove the initial "."
+# And the CLI then renames it
+
+# node files
```
dwieeb, line 5:
```suggestion
# node files
```

filesystem/ios/Plugin/Filesystem.swift
```diff
@@ -0,0 +1,156 @@
...
+
+@objc public class Filesystem: NSObject {
+
+    public enum FilesystemError: String, Error {
+        case noParentFolder = "Parent folder doesn't exist"
+        case noSave = "Unable to save file"
+        case failEncode = "Unable to encode data to utf-8"
+        case noAppend = "Unable to append file"
+        case notEmpty = "Folder is not empty"
+    }
```
ikeith, line 11: This will not work as expected. While it can conform to the `Error` protocol, calling `localizedDescription` on it will cause it to be bridged to an `NSError` and the `RawRepresentable` string will get lost. Instead of the defined string value it will print a generated, inscrutable description of the object. This should conform to LocalizedError to make sure the human-readable values are passed back.
```suggestion
    public enum FilesystemError: LocalizedError {
        case noParentFolder, noSave, failEncode, noAppend, notEmpty
        
        public var errorDescription: String? {
            switch self {
            case .noParentFolder:
                return "Parent folder doesn't exist"
            case .noSave:
                return "Unable to save file"
            case .failEncode:
                return "Unable to encode data to utf-8"
            case .noAppend:
                return "Unable to append file"
            case .notEmpty:
                return "Folder is not empty"
            }
        }
    }
```

filesystem/ios/Plugin/Filesystem.swift
```diff
@@ -0,0 +1,156 @@
...
+
+    @objc public func readFile(at fileUrl: URL, with encoding: String?) throws -> String {
```
ikeith, line 13: Are you sure this will work? It used to be that a Swift function cannot be exposed to Obj-C if it both throws and returns a value (since the Obj-C pattern is to return a `BOOL` for success and use a pointer for the result and a handle for the `NSError`). I don't know if that's changed recently, though. It might be better to not expose this class to Obj-C at all if the methods can't be cleanly mapped.

ikeith (changes requested):

filesystem/ios/Plugin/Filesystem.swift
jcesarmobile: I works in Swift, didn't try on Objective-C

filesystem/ios/Plugin/Filesystem.swift
jcesarmobile: yeah, on the Plugin class when it gets caught, I send the raw value to JS, not the localized description

ikeith: Ah, I missed the separate handling. But that could be simplified. Instead of multiple `catch` statements, it could use one for any type of error and call `localizedDescription` on it.

ikeith (approved):

filesystem/android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,5 @@
+
+  <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+      package="com.capacitorjs.plugins.filesystem">
+  </manifest>
+
```
dwieeb, line 5:
```suggestion
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    package="com.capacitorjs.plugins.filesystem">
</manifest>
```

filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/FilesystemPlugin.java
```diff
@@ -0,0 +1,494 @@
...
+        PluginRequestCodes.FILESYSTEM_REQUEST_WRITE_FILE_PERMISSIONS,
+        PluginRequestCodes.FILESYSTEM_REQUEST_WRITE_FOLDER_PERMISSIONS,
+        PluginRequestCodes.FILESYSTEM_REQUEST_READ_FILE_PERMISSIONS,
+        PluginRequestCodes.FILESYSTEM_REQUEST_READ_FOLDER_PERMISSIONS,
+        PluginRequestCodes.FILESYSTEM_REQUEST_DELETE_FILE_PERMISSIONS,
+        PluginRequestCodes.FILESYSTEM_REQUEST_DELETE_FOLDER_PERMISSIONS,
+        PluginRequestCodes.FILESYSTEM_REQUEST_URI_PERMISSIONS,
+        PluginRequestCodes.FILESYSTEM_REQUEST_STAT_PERMISSIONS,
+        PluginRequestCodes.FILESYSTEM_REQUEST_RENAME_PERMISSIONS,
+        PluginRequestCodes.FILESYSTEM_REQUEST_COPY_PERMISSIONS
```
dwieeb, line 35: Just a note that this is still using `PluginRequestCodes`

jcesarmobile: Yeah, @carlpoole said that we should change all the codes at once when all plugins are here so we don’t miss any

filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/FilesystemPlugin.java
```diff
@@ -0,0 +1,494 @@
...
+    },
+    permissions = {
+        @Permission(permission = Manifest.permission.READ_EXTERNAL_STORAGE, alias = "publicStorage"),
+        @Permission(permission = Manifest.permission.WRITE_EXTERNAL_STORAGE, alias = "publicStorage")
```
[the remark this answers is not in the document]
jcesarmobile: The permission is only needed if you try to access the public storage (shared), so I called it publicStorage instead of going for just "storage" as I think that might cause confusion and make users request it every time despite they don't need it for non public storage.
But if you have a better name, I'm open for suggestions

filesystem/src/web.ts
```diff
@@ -0,0 +1,585 @@
...
+    return this._copy(options, false);
+  }
+
+  async requestPermissions(): Promise<void> {
+    throw this.unimplemented('Not implemented on web.');
+  }
+
+  async checkPermissions(): Promise<FilesystemPermissionStatus> {
+    throw this.unimplemented('Not implemented on web.');
+  }
```
dwieeb, line 411: These also should return granted because we can always write to indexeddb

filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/FilesystemPlugin.java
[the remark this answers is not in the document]
dwieeb: My initial thought was `externalStorage`, but if `publicStorage` is more clear (and more cross-platform), then it's good 👍 I was mostly just curious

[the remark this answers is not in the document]
jcesarmobile: there are some external storage locations that don't need permission. The code had a `isPublicDirectory` function that checks if it's one of the locations that need the permission, that's why I also used public on the name.

filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/FilesystemPlugin.java
```diff
@@ -0,0 +1,494 @@
...
+    @Override
+    protected void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+        Logger.debug(getLogTag(), "handling request perms result");
+        this.handlePermissions(permissions, grantResults);
+        if (getSavedCall() == null) {
+            Logger.debug(getLogTag(), "No stored plugin call for permissions request result");
+            return;
+        }
+
+        PluginCall savedCall = getSavedCall();
```
carlpoole, line 456:
```suggestion
        Logger.debug(getLogTag(), "handling request perms result");
        PluginCall savedCall = getSavedCall();
         if (getSavedCall() == null) {
            Logger.debug(getLogTag(), "No stored plugin call for permissions request result");
            return;
        }
        
        if (!validatePermissions(permissions, grantResults)) {
            freeSavedCall();
            return;
        }
       
       if (savedCall.getMethodName().equals("requestPermissions")){
            savedCall.resolve(getPermissionStates());
            freeSavedCall();
            return;
        }
```

filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/FilesystemPlugin.java
carlpoole, line 39: This can be defined in a cleaner way now as of todays update: https://github.com/ionic-team/capacitor/pull/3822

```suggestion
        @Permission(permission = { Manifest.permission.READ_EXTERNAL_STORAGE, Manifest.permission.WRITE_EXTERNAL_STORAGE }, alias = "publicStorage")
```

jcesarmobile:
I’m aware of pending changes, you don’t need to remind them, but I’m not going to push them until we have a new alpha release. In fact I’ve made most of them on my local branch but didn’t push them because things keep changing again and again.
Also, please, don’t push to the branch because I’m wasting more time fixing conflicts and keeping things up to date than the time it took to initially create the PR.
It’s been 3 months already.

filesystem/android/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+    google()
+    jcenter()
+    mavenCentral()
+}
+
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation project(':capacitor-android')
+    testImplementation "junit:junit:$junitVersion"
```
dwieeb, line 53:
```suggestion
    implementation project(':capacitor-android')
    implementation "androidx.appcompat:appcompat:$androidxAppCompatVersion"
    testImplementation "junit:junit:$junitVersion"
```

filesystem/android/build.gradle
```diff
@@ -0,0 +1,56 @@
+ext {
+    junitVersion =  project.hasProperty('junitVersion') ? rootProject.ext.junitVersion : '4.12'
+    androidxJunitVersion =  project.hasProperty('androidxJunitVersion') ? rootProject.ext.androidxJunitVersion : '1.1.1'
```
dwieeb, line 3:
```suggestion
    junitVersion =  project.hasProperty('junitVersion') ? rootProject.ext.junitVersion : '4.12'
    androidxAppCompatVersion = project.hasProperty('androidxAppCompatVersion') ? rootProject.ext.androidxAppCompatVersion : '1.1.0'
    androidxJunitVersion =  project.hasProperty('androidxJunitVersion') ? rootProject.ext.androidxJunitVersion : '1.1.1'
```

ikeith (approved):

carlpoole (approved):

dwieeb (approved):

Issue #27, opened 2020-09-01
Ignore certain files when publishing


dwieeb:
This is a list of all published files from the TextZoom plugin, but we should ignore the same patterns in all plugins (`files`? `.npmignore`?):

```
CHANGELOG.md
CapacitorTextZoom.podspec
android/.gradle/6.1.1/executionHistory/executionHistory.bin
android/.gradle/6.1.1/executionHistory/executionHistory.lock
android/.gradle/6.1.1/fileChanges/last-build.bin
android/.gradle/6.1.1/fileContent/fileContent.lock
android/.gradle/6.1.1/fileHashes/fileHashes.bin
android/.gradle/6.1.1/fileHashes/fileHashes.lock
android/.gradle/6.1.1/fileHashes/resourceHashesCache.bin
android/.gradle/6.1.1/gc.properties
android/.gradle/6.1.1/javaCompile/classAnalysis.bin
android/.gradle/6.1.1/javaCompile/jarAnalysis.bin
android/.gradle/6.1.1/javaCompile/javaCompile.lock
android/.gradle/6.1.1/javaCompile/taskHistory.bin
android/.gradle/buildOutputCleanup/buildOutputCleanup.lock
android/.gradle/buildOutputCleanup/cache.properties
android/.gradle/buildOutputCleanup/outputFiles.bin
android/.gradle/checksums/checksums.lock
android/.gradle/vcs-1/gc.properties
android/build.gradle
android/gradle.properties
android/gradle/wrapper/gradle-wrapper.jar
android/gradle/wrapper/gradle-wrapper.properties
android/gradlew
android/gradlew.bat
android/proguard-rules.pro
android/settings.gradle
android/src/androidTest/java/com/getcapacitor/android/ExampleInstrumentedTest.java
android/src/main/AndroidManifest.xml
android/src/main/java/com/capacitorjs/plugins/textzoom/TextZoom.java
android/src/main/java/com/capacitorjs/plugins/textzoom/TextZoomPlugin.java
android/src/main/res/layout/bridge_layout_main.xml
android/src/test/java/com/getcapacitor/ExampleUnitTest.java
dist/esm/definitions.d.ts
dist/esm/definitions.js
dist/esm/definitions.js.map
dist/esm/index.d.ts
dist/esm/index.js
dist/esm/index.js.map
dist/esm/ios.d.ts
dist/esm/ios.js
dist/esm/ios.js.map
dist/plugin.js
dist/plugin.js.map
[...]

PR #32, opened 2020-09-01
fix: better ignore rules for npm distribution

dwieeb:
fixes https://github.com/ionic-team/capacitor-plugins/issues/27

see https://github.com/ionic-team/create-capacitor-plugin/pull/3


motion/.npmignore
```diff
@@ -0,0 +1,7 @@
+android/.gradle
+ios/Pods
+ios/Podfile.lock
```
jcesarmobile, line 3: motion is web only, so doesn't have any of this

jcesarmobile:
I think it should ignore everything that is in the gitignore too

motion/.npmignore
dwieeb: I considered this, but I'm fine specializing it

dwieeb:
> I think it should ignore everything that is in the gitignore too

Okay, I investigated this. git and npm treat gitignore slightly differently.

- We'll need to remove all trailing slashes in the gitignore rules for our template: https://github.com/ionic-team/create-capacitor-plugin/pull/4 
- Here's the diff after doing that:

```diff
diff --git a/0 b/1
index 70a8c7d..6756aa0 100644
--- a/0
+++ b/1
@@ -1,3 +1,4 @@
+.prettierignore
 CHANGELOG.md
 CapacitorTextZoom.podspec
 LICENSE
@@ -44,3 +45,8 @@ ios/PluginTests/Info.plist
 ios/PluginTests/PluginTests.swift
 ios/Podfile
 package.json
+rollup.config.js
+src/definitions.ts
+src/index.ts
+src/ios.ts
+tsconfig.json
```

So, there are still files we shouldn't distribute (`src/`, configuration files, etc)

PR #33, opened 2020-09-01
Camera

ikeith:
- [x]  Remove plugin from capacitor (https://github.com/ionic-team/capacitor/pull/3611)
 - [x] JS wrapper and implementation
 - [x]  iOS implementation
 - [x]  Android implementation
 - [x]  Test app integration (https://github.com/ionic-team/capacitor-testapp/pull/15)


dalepo:
Just a FYI there's an open issue regarding the camera plugin resetting the application (Android)
https://github.com/ionic-team/capacitor/issues/2931

asztal:
I've been working on a plugin of my own to implement video capture. It's at a very early stage at the moment but if I were to make a PR to the capacitor camera plugin, would it be considered, or would I have to make a community plugin?

I would just use the cordova plugin for capturing video but in my case I have received a lot of crash reports relating to the cordova camera plugin when used with Capacitor.

dwieeb:
@asztal Thanks for bringing it up! Video capture would belong in the Camera plugin for sure. I've added this issue to track it: https://github.com/ionic-team/capacitor-plugins/issues/78

We'd be happy to look at a PR that adds that functionality! We're hoping to merge this PR into `main` sometime this week.

barrcodes:
@ikeith @dwieeb I'm curious what the plan is here? Our enterprise application had a bug with Samsung devices which I was able to create a hotfix for by patching Camera.java. I was coming here to submit a PR to fix the issue, and noticed that the Camera plugin is gone.

Is a new plugin on the horizon?

ikeith:
@barrcodes As part of the work on capacitor 3, the core plugins are being split out into separate packages (this PR is for the migrated plugin into its own package). We are cleaning up a few things in the process but it is incomplete at the moment, as-is capacitor 3. If you have a critical fix, please open a PR for it against the camera plugin in [the 2.x branch](https://github.com/ionic-team/capacitor/tree/2.x) of capacitor.

camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
```diff
@@ -0,0 +1,602 @@
...
+ * On Android, this plugin sends an intent that opens the stock Camera app.
+ *
+ * Adapted from https://developer.android.com/training/camera/photobasics.html
+ */
+@CapacitorPlugin(
+    name = "Camera",
+    permissions = {
+        @Permission(permission = Manifest.permission.CAMERA, alias = "camera"),
+        @Permission(permission = Manifest.permission.WRITE_EXTERNAL_STORAGE, alias = "writePhotos"),
+        @Permission(permission = Manifest.permission.READ_EXTERNAL_STORAGE, alias = "readPhotos")
```
carlpoole, line 49: Internal "permission" element is renamed to "strings" now as of todays update: ionic-team/capacitor#3822

```suggestion
        @Permission(strings = { Manifest.permission.CAMERA }, alias = "camera"),
        @Permission(strings = { Manifest.permission.WRITE_EXTERNAL_STORAGE }, alias = "writePhotos"),
        @Permission(strings = { Manifest.permission.READ_EXTERNAL_STORAGE }, alias = "readPhotos")
```

camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
```diff
@@ -0,0 +1,602 @@
...
+                    editor.putString(permission, "prompt-with-rationale");
+                } else {
+                    // Permission denied permanently, store this state for future reference
+                    editor.putString(permission, "denied");
+                }
+
+                editor.apply();
+            }
+        }
+        // end block
```
carlpoole, line 496:
```suggestion
        if (!validatePermissions(permissions, grantResults)) {
            freeSavedCall();
            return;
        }
```

camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
```diff
@@ -0,0 +1,572 @@
...
+                    savedCall.release(bridge);
+                    return;
+                }
+            }
+            doShow(savedCall);
+        } else {
+            savedCall.resolve(getPermissionStates());
+            savedCall.release(bridge);
+        }
+    }
```
carlpoole, line 484:
```suggestion
    protected void onRequestPermissionsResult(PluginCall savedCall, int requestCode, String[] permissions, int[] grantResults) {
        Logger.debug(getLogTag(), "handling request perms result");

        if (savedCall.getMethodName().equals("getPhoto")) {
            for (int i = 0; i < grantResults.length; i++) {
                int result = grantResults[i];
                String perm = permissions[i];
                if (result == PackageManager.PERMISSION_DENIED) {
                    Logger.debug(getLogTag(), "User denied camera permission: " + perm);
                    savedCall.reject(PERMISSION_DENIED_ERROR);
                    return;
                }
            }
            doShow(savedCall);
        }
    }
```

camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
```diff
@@ -0,0 +1,572 @@
...
+
+    @Override
+    protected void handleOnActivityResult(int requestCode, int resultCode, Intent data) {
+        super.handleOnActivityResult(requestCode, resultCode, data);
+
+        PluginCall savedCall = getSavedCall();
+
+        if (savedCall == null) {
+            return;
+        }
```
carlpoole, line 494:
```suggestion
    protected void handleOnActivityResult(PluginCall savedCall, int requestCode, int resultCode, Intent data) {
        if (savedCall == null) {
            return;
        }
```

camera/ios/Plugin/CameraPlugin.swift
```diff
@@ -0,0 +1,439 @@
...
+            } else {
+                self?.call?.reject("User denied access to camera")
+            }
+        }
+    }
+
+    func showPhotos() {
+        // check for permission
+        let authStatus = PHPhotoLibrary.authorizationStatus()
+        if authStatus == .restricted || authStatus == .denied {
```
ikeith, line 298: Is this the behavior we want? Technically, the picker (either PHPickerViewController on iOS 14 or UIImagePickerController) does not require permission. The picker is presented out-of-process and requires user interaction so even if the user denies library access the application can still show the picker and get an image. The permission allows us to access the underlying asset and get the additional metadata about the image that the user selected. With denied permissions we can still get a picture but it won't have any metadata to return with it. That's potentially confusing and is inconsistent with other platforms and how the camera works, though.

camera/src/web.ts
```diff
@@ -0,0 +1,183 @@
...
+        if (customElements.get('pwa-camera-modal')) {
+          const cameraModal: any = document.createElement('pwa-camera-modal');
+          document.body.appendChild(cameraModal);
+          try {
+            await cameraModal.componentOnReady();
+            cameraModal.addEventListener('onPhoto', async (e: any) => {
+              const photo = e.detail;
+
+              if (photo === null) {
+                reject('User cancelled photos app');
```
dwieeb, line 26: We should use a `CapacitorException` for this, not just throw strings.

dwieeb (approved):

camera/src/definitions.ts
```diff
@@ -0,0 +1,197 @@
...
+   */
+  checkPermissions(): Promise<CameraPermissionStatus>;
+
+  /**
+   * Request camera and photo album permissions
+   *
+   * @since 1.0.0
+   */
+  requestPermissions(
+    permissions: CameraPluginPermissions | null,
```
jcesarmobile, line 36:
```suggestion
    permissions?: CameraPluginPermissions
```
I think it should be like this, so `Camera.requestPermissions();` can be called without params instead of passing an empty array

jcesarmobile (changes requested):
I've requested a few changes.

Also, I get random crashes on iOS requesting permissions.
I've added a button to request photos permission and another for camera permission and I first click the photos one and don't allow to access all photos, then by clicking the camera  or the all permissions button, it randomly crash with `EXC_BAD_ACCESS`.

Also, the Camera Intent on Android don't really need the Camera permission, so if the Camera permission is not present on the AndroidManifest.xml, the plugin should not require the permission and everything should still work. 
In v2 it requested the permission because the apps shipped with it, but on v3 won't be the case, so we should have that into account. Should first check if the permission is in the manifest, and only request it if present.

camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
```diff
@@ -0,0 +1,549 @@
...
+import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.net.Uri;
+import android.os.Bundle;
+import android.provider.MediaStore;
+import android.util.Base64;
+import androidx.core.content.FileProvider;
+import com.getcapacitor.FileUtils;
```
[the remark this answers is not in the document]
ikeith: This is used: https://github.com/ionic-team/capacitor-plugins/blob/e9752c999a4817a12d6c06c9e0abc207445b2ceb/camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java#L370

ikeith:
> Also, I get random crashes on iOS requesting permissions.
> I've added a button to request photos permission and another for camera permission and I first click the photos one and don't allow to access all photos, then by clicking the camera or the all permissions button, it randomly crash with `EXC_BAD_ACCESS`.

The only crash I can see is on the simulator when trying to open the camera. That's caused by both the plugin showing an alert that the camera doesn't exist in the simulator and the test app showing an alert with the error result. Trying to show an alert while one is on screen causes an exception to be thrown.

However, the plugin was not properly reading the parameters to request permissions for only a subset of permissions (so all calls to requestPermissions caused it to request everything). That has been fixed.

camera/ios/Plugin/CameraPlugin.swift
```diff
@@ -0,0 +1,420 @@
...
+
+/*
+ Runtime detection of iOS 14 with @available and #available is straightforward but that code will fail to compile
+ under Xcode 11. So we need to use statements in the form of
+
+ #if swift(>=5.3)
+ #endif
+
+ as a poor proxy for Xcode 12 detection. The conditionals should be removed once Xcode 12 is required.
+ */
```
jcesarmobile, line 14:
```suggestion
```

can be removed since the Xcode 11 support code has been removed

jcesarmobile:
I get the error on real devices and there is no alert present (because it would prevent me from clicking the request all permissions button).

Sometimes I get the `EXC_BAD_ACCESS `
other times I get this error

> -[__NSTaggedDate count]: unrecognized selector sent to instance 0x8000000000000000
> Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__NSTaggedDate count]: unrecognized selector sent to instance 0x8000000000000000'
> *** First throw call stack:
> (0x19122d654 0x190f4fbcc 0x191131dd8 0x1912317f8 0x19123371c 0x19e73227c 0x19e952aa4 0x19e731e50 0x1013011cc 0x101301230 0x10156e338 0x10156f730 0x101571e88 0x101580d74 0x101581698 0x190f44b38 0x190f47740)
> libc++abi.dylib: terminating with uncaught exception of type NSException
> *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__NSTaggedDate count]: unrecognized selector sent to instance 0x8000000000000000'
> terminating with uncaught exception of type NSException

it's not related to the limited permissions, I also get it if I accept both, just need to click the "Request All Permissions" button a few times, it usually crash on the 4-5 attempt, but sometimes it fails on the first one, other times takes more than 10 attempts.
Tested on iOS 13.6.1 and 14.3 beta devices, crashes on both, so not related to the beta.

ikeith:
Ohhh, I didn't understand that you were making _repeated_ calls to requestPermissions. That stirred up a race condition between the permission closures. Threading is hard. 🤕 That should be fixed along with the result not returning everything.

ikeith:
> Also, the Camera Intent on Android don't really need the Camera permission, so if the Camera permission is not present on the AndroidManifest.xml, the plugin should not require the permission and everything should still work.
> In v2 it requested the permission because the apps shipped with it, but on v3 won't be the case, so we should have that into account. Should first check if the permission is in the manifest, and only request it if present.

This behavior has been changed but it is dependent on https://github.com/ionic-team/capacitor/pull/3911

ikeith:
> > Also, the Camera Intent on Android don't really need the Camera permission, so if the Camera permission is not present on the AndroidManifest.xml, the plugin should not require the permission and everything should still work.
> > In v2 it requested the permission because the apps shipped with it, but on v3 won't be the case, so we should have that into account. Should first check if the permission is in the manifest, and only request it if present.
> 
> This behavior has been changed but it is dependent on [ionic-team/capacitor#3911](https://github.com/ionic-team/capacitor/pull/3911)

This has been updated again, now dependent on [ionic-team/capacitor#3939](https://github.com/ionic-team/capacitor/pull/3939)

carlpoole (approved):

camera/src/definitions.ts
```diff
@@ -0,0 +1,222 @@
+export type CameraPermissionState = PermissionState | 'limited';
```
dwieeb, line 1: Where is `PermissionState` imported?

ikeith: Does it need to be imported? It's defined in `lib.dom.d.ts`.

dwieeb: Yeah, it should use our definition from `@capacitor/core` which includes `prompt-with-rationale`. `lib.dom.d.ts` are global TypeScript types for the browser DOM, not our types.

dwieeb (approved):

dwieeb:
I wonder if `limited` should be `granted-limited` or something so it starts with the more general state, much like `prompt-with-rationale`.

Also, we may have talked about this before, but what are the actual implications of limited photo access? Why does the app developer need to know if a user has granted only a subset of photos?

ikeith:
`granted-limited` makes sense but, on the other hand, `limited` is what the platform API returns so it has the benefit of being easier to look up. But I don't have a strong opinion either way.

The practical result of `limited` is that we cannot get the underlying PHAsset for any image that the user has not chosen to include in the list of selected images. Which means we cannot get any metadata for it. The user can still select any image in the picker regardless of what permissions they grant, but we will only be able to read metadata on the images for which they have granted access. So the state is a signal to the developer that metadata is less likely to be returned but we can't know for sure until the user actually picks one.

dwieeb (approved):

PR #52, opened 2020-09-25
Toast

jcesarmobile:


dwieeb:
@adamdbradley Needs PWA element for web side

carlpoole (approved):

toast/ios/Plugin/ToastLabel.swift
```diff
@@ -0,0 +1,26 @@
+import UIKit
+
+@objc public class ToastLabel: UILabel {
```
ikeith, line 3: This should probably be internal, not public. There's no reason to expose it outside of the plugin.
```suggestion
@objc internal class ToastLabel: UILabel {
```

toast/ios/Plugin/ToastLabel.swift
```diff
@@ -0,0 +1,26 @@
...
+    public var padding: UIEdgeInsets? {
+        get {
+            return objc_getAssociatedObject(self, &AssociatedKeys.padding) as? UIEdgeInsets
+        }
+        set {
+            if let newValue = newValue {
+                objc_setAssociatedObject(self, &AssociatedKeys.padding, newValue as UIEdgeInsets, objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+            }
+        }
+    }
```
ikeith, line 17: This originally used an associated object because extensions can't add properties. But since this is now a subclass, it can be much simpler. 
```suggestion
    var padding: UIEdgeInsets?
```

ikeith (changes requested):

dwieeb (approved):

ikeith (approved):

PR #62, opened 2020-10-08
chore: Add docgen script and generate plugin docs

jcesarmobile:


dwieeb:
Were we also going to generate a JSON file of the docs @adamdbradley?

PR #64, opened 2020-10-13
chore(haptics): consistent enum naming style

dwieeb:
This would be a breaking change in Haptics if people were using it, so not merging right away (unlike https://github.com/ionic-team/capacitor/pull/3671)


jcesarmobile (approved):
If we are going to break it, better break it now

PR #71, opened 2020-10-16
feat: App plugin

jcesarmobile:
removed openUrl and canOpenUrl functions, will be a separate plugin
add getInfo function to get information about the app (was part of Device plugin but got removed because makes more sense here)

depends on https://github.com/ionic-team/capacitor/pull/3676

closes https://github.com/ionic-team/capacitor-plugins/issues/72


app/ios/Plugin/AppPlugin.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+import Capacitor
+
+@objc(AppPlugin)
+public class AppPlugin: CAPPlugin {
```
ikeith, line 5: This should use a deinit method to unregister from NotificationCenter so it won't crash if it gets released (likely under embedded use).

ikeith (changes requested):

app/src/web.ts
```diff
@@ -0,0 +1,40 @@
...
+      document.addEventListener(
+        'visibilitychange',
+        this.handleVisibilityChange.bind(this),
+        false,
+      );
+    }
+  }
+
+  exitApp(): never {
+    throw new Error('Method not implemented.');
```
dwieeb, line 18: Would it be a bad idea to call `window.close()` ?

app/src/web.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  async getInfo(): Promise<AppInfo> {
+    throw new Error('Method not implemented.');
```
dwieeb, line 22: Do we not know this information for Capacitor web? We could implement this in the future, maybe.

jcesarmobile: No, we don't.
There was a proposal of copying web files to a different folder, including the capacitor.config.json, if we do that we could read the file and get the app name and app id, but version and build won't be available, unless we also copy the package.json or make the cli put some value from it on the capacitor.config.json based on the package.json version

dwieeb: Yeah, that's what I was thinking. But we can do it later

app/src/web.ts
jcesarmobile: I've tried and it doesn't close the window, just shows `Scripts may close only the windows that were opened by them.`

dwieeb (approved):

ikeith (approved):

carlpoole (approved):

app/src/web.ts
dwieeb, line 18:
```suggestion
    throw this.unimplemented('Not implemented on web.');
```

app/src/web.ts
dwieeb, line 22:
```suggestion
    throw this.unimplemented('Not implemented on web.');
```

aparajita: Hmm... it seems to me it would be more useful (and developer-friendly) to at least attempt to return something useful on the web. I see no technical reason why it can't at least the name and version from package.json.

aparajita: @dwieeb Any thoughts?

dwieeb: We've talked about doing that in the future. We don't have access to `package.json` in the web runtime, but the Capacitor CLI could write that info somewhere where it's available.

app/src/web.ts
```diff
@@ -0,0 +1,40 @@
...
+      );
+    }
+  }
+
+  exitApp(): never {
+    throw this.unimplemented('Not implemented on web.');
+  }
+
+  async getInfo(): Promise<AppInfo> {
+    throw this.unimplemented('Not implemented on web.');
```
aparajita, line 22: I would suggest it will be a lot more useful (and developer-friendly) to return the information you can get from package.json, rather than throw an error.

jcesarmobile: dwieeb already answered you on the other comment, the package.json is not accesible from the app/plugin, would need some CLI work first.

aparajita: Github is refusing to show me the other comment and his response, that's why I had to post it again.

I forgot this is a separate plugin now. Maybe Capacitor could make the app's package.json (or a subset of it) available to plugins.

dwieeb: You have to click "Show resolved":
image

aparajita: Thanks, the github review UI is getting better, but is still somewhat confusing at times.

PR #74, opened 2020-10-19
feat: AppLauncher plugin

jcesarmobile:


dwieeb (approved):

app-opener/android/src/main/java/com/capacitorjs/plugins/appopener/AppOpenerPlugin.java
```diff
@@ -0,0 +1,69 @@
...
+import com.getcapacitor.NativePlugin;
+import com.getcapacitor.Plugin;
+import com.getcapacitor.PluginCall;
+import com.getcapacitor.PluginMethod;
+
+@NativePlugin(name = "AppOpener")
+public class AppOpenerPlugin extends Plugin {
+
+    @PluginMethod
+    public void canOpenUrl(PluginCall call) {
```
carlpoole, line 18: Just a note: this is code that Google did not like due to lack of prominent privacy disclosure. We probably want to wait before removing it though until we fully understand the situation and long-term plan.

carlpoole (approved):

app-opener/android/src/main/java/com/capacitorjs/plugins/appopener/AppOpenerPlugin.java
jcesarmobile: I think we just need to document that this plugin requires a privacy disclosure about this. The point of being an independent plugin is that only people interested in this feature will install it, so people not using it don't need the disclosure.

ikeith (approved):

carlpoole (approved):

dwieeb (approved):

PR #76, opened 2020-10-20
feat(browser): support other custom tabs providers

garaboncias:
Support other browsers, when chrome is not available (only Firefox or Edge installed)


jcesarmobile (changes requested):
Thanks for the PR!

For the code to work on Android 11+ you need to add this in the plugin's `AndroidManifest.xml`

```
<queries>
    <intent>
        <action android:name="android.support.customtabs.action.CustomTabsService" />
    </intent>
</queries>
```

Also, in some of my devices `getPackageName` is returning null and that crash the app, so in case it's null it should fall back to the hardcoded `CUSTOM_TAB_PACKAGE_NAME` value

cornflakes:
@jcesarmobile Added the requested changes.

PR #88, opened 2020-11-02
Add deviceInfo.name implementation for Android

### LINKED ISSUE #86: feat: Android, implement deviceInfo.Name property
### Description
For iOS the name property on device info is already implemented. See https://github.com/ionic-team/capacitor/issues/2130
But for Android the implementation is missing.

### Platform(s)
Android

### Preferred Solution
Added in api level 25 there is a way to easily get the device name.
https://developer.android.com/reference/android/provider/Settings.Global#DEVICE_NAME

If on api 25+ then return the device name. Otherwise return null.

If wanted, I could create a PR with the preferred solution in this repo for the next 2.x capacitor version and a separate PR for the 3.x capacitor version in the plugins repo.

DaanWasscher:
Fixes #86


device/README.md
```diff
@@ -75,7 +75,7 @@ Get the device's current language locale code.
 
 | Prop                  | Type                                                               | Description                                                                                                                                  | Since |
 | --------------------- | ------------------------------------------------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------- | ----- |
-| **`name`**            | <code>string</code>                                                | The name of the device. For example, "John's iPhone". This is only supported on iOS.                                                         | 1.0.0 |
+| **`name`**            | <code>string</code>                                                | The name of the device. For example, "John's iPhone". This is only supported on iOS and Android.                                             | 1.0.0 |
```
carlpoole, line 78:
```suggestion
| **`name`**            | <code>string</code>                                                | The name of the device. For example, "John's iPhone". This is only supported on iOS and Android 7.1 or above.                                             | 1.0.0 |
```

device/src/definitions.ts
```diff
@@ -10,7 +10,7 @@ export interface DeviceInfo {
   /**
    * The name of the device. For example, "John's iPhone".
    *
-   * This is only supported on iOS.
+   * This is only supported on iOS and Android.
```
carlpoole, line 13:
```suggestion
   * This is only supported on iOS and Android 7.1 or above.
```

carlpoole (commented):
Looks good, I would suggest adding the version info with Android since it won't work for all versions.

device/README.md
DaanWasscher: Yes, good idea!

dwieeb:
@DaanWasscher Can you run `npm run build` and commit the whitespace changes to the readme?

carlpoole (approved):

dwieeb:
Thanks @DaanWasscher! This will make it in to the Device plugin before the launch of Capacitor 3.0. 💙

PR #89, opened 2020-11-04
return Location object instead of JSObject

jcesarmobile:


carlpoole:
I do like this better. Feel free to merge it into my PR branch if you are done

PR #92, opened 2020-11-05
docs: add additional plugin documentation from v2 site

dwieeb:
We're missing the non-API documentation from our v3 plugin API docs:

https://capacitorjs.com/docs/v3/plugins
https://capacitorjs.com/docs/plugins

Still need docs for:
- [x] Camera
- [x] Filesystem
- [x] Geolocation
- [ ] Local Notifications
- [ ] Push Notifications
- [x] Splash Screen


aparajita:
I'll be making a PR for the new Splash Screen next week. Code is done, just need to document the API and write a demo app.

https://github.com/willsub/ws-capacitor-splashscreen

app-launcher/README.md
```diff
@@ -9,6 +9,22 @@ npm install @capacitor/app-launcher
 npx cap sync
 ```
 
+## Example
+
+```typescript
+import { App } from '@capacitor/app';
```
jcesarmobile, line 15:
```suggestion
import { AppLauncher } from '@capacitor/app-launcher';
```

app-launcher/README.md
```diff
@@ -9,6 +9,22 @@ npm install @capacitor/app-launcher
...
+
+const checkCanOpenUrl = async () => {
+  const { value } = await App.canOpenUrl({ url: 'com.getcapacitor.myapp' });
```
jcesarmobile, line 18:
```suggestion
  const { value } = await AppLauncher.canOpenUrl({ url: 'com.getcapacitor.myapp' });
```

app-launcher/README.md
```diff
@@ -9,6 +9,22 @@ npm install @capacitor/app-launcher
...
+
+  alert('Can open url: ', value);
+};
+
+const openPortfolioPage = async () => {
+  await App.openUrl({ url: 'com.getcapacitor.myapp://page?id=portfolio' });
```
jcesarmobile, line 24:
```suggestion
  await AppLauncher.openUrl({ url: 'com.getcapacitor.myapp://page?id=portfolio' });
```

haptics/README.md
```diff
@@ -9,6 +9,45 @@ npm install @capacitor/haptics
...
 
+## Permissions
+
+On Android, you must add the `VIBRATE` permission to your app's
+`AndroidManifest.xml` file.
+
+```xml
+<uses-permission android:name="android.permission.VIBRATE" />
+```
+
```
jcesarmobile, line 20:
```suggestion
```

So, the plugin is already adding the permissions, so users don't have to add it
https://github.com/ionic-team/capacitor-plugins/blob/main/haptics/android/src/main/AndroidManifest.xml#L3

I think we said that plugins should not add permissions, but I think that should only be the case for "hard" permissions that requires the user to prompt and accept them, but for "soft" permissions such as this one, I think it's better to add them in the manifest instead of making users manually add them.

If you don't agree, then the permission should be removed from the `AndroidManifest.xml`

network/README.md
```diff
@@ -9,6 +9,34 @@ npm install @capacitor/network
...
+On Android, you must add the `ACCESS_NETWORK_STATE` permission to your app's
+`AndroidManifest.xml` file.
+
+```xml
+<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+```
+
+This permission allows the app to access information about the current network,
+such as whether it is connected to Wi-Fi or cellular.
+
```
jcesarmobile, line 23: Same as for haptics, the permission is already added by the plugin https://github.com/ionic-team/capacitor-plugins/blob/main/network/android/src/main/AndroidManifest.xml#L3

so I think we should remove this section, or remove the permission from the manifest

status-bar/README.md
```diff
@@ -9,6 +9,50 @@ npm install @capacitor/status-bar
...
 
+## iOS Note
+
+This plugin requires "View controller-based status bar appearance"
+(`UIViewControllerBasedStatusBarAppearance`) set to `YES` in `Info.plist`. Read
+about [Configuring iOS](https://capacitorjs.com/docs/ios/configuration) for
+help.
+
+The status bar visibility defaults to visible and the style defaults to
+`StatusBarStyle.Light`. You can change these defaults by adding
```
jcesarmobile, line 20:
```suggestion
`StatusBarStyle.Default`. You can change these defaults by adding
```

I think the default is Default, not Light

storage/README.md
```diff
@@ -2,13 +2,58 @@
...
+API should be used instead. This API will fall back to using `localStorage`
+when running as a Progressive Web App.
+
+This plugin will use
+[`UserDefaults`](https://developer.apple.com/documentation/foundation/userdefaults)
+on iOS and
+[`SharedPreferences`](https://developer.android.com/reference/android/content/SharedPreferences)
+on Android. Stored data is cleared if the app is uninstalled.
+
+**Note**: this API is _not_ meant for high-performance data storage
```
jcesarmobile, line 15: I think another word should be used instead of high-performance, because I've seen people thinking this means slow, which is not true.

jcesarmobile (changes requested):

haptics/README.md
dwieeb: I agree, I just forgot we're able to do this because each plugin has its own manifest

PR #94, opened 2020-11-10
Local Notifications

dwieeb:
TODO:
- [x] Add shared [`UNUserNotificationCenterDelegate`](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate) to Capacitor core
- [x] Fix `requestPermissions` on Android
- [x] Documentation
- [x] rename events
- [x] definitions.ts channel method updates (SDK 26+) 

testapp PR: https://github.com/ionic-team/capacitor-testapp/pull/53


local-notifications/ios/Plugin/LocalNotificationsDelegate.swift
```diff
@@ -0,0 +1,87 @@
...
+    public func requestPermissions(with completion: ((Bool, Error?) -> Void)? = nil) {
+        let center = UNUserNotificationCenter.current()
+        center.requestAuthorization(options: [.badge, .alert, .sound]) { (granted, error) in
+            completion?(granted, error)
+        }
+    }
+
+    public func checkPermissions(with completion: ((Bool, Error?) -> Void)? = nil) {
+        let center = UNUserNotificationCenter.current()
+        center.requestAuthorization(options: [.badge, .alert, .sound]) { (granted, error) in
```
jcesarmobile, line 21: this will request the permissions, I don't think that's expected on checkPermissions

jcesarmobile (changes requested):
Added a few comments.

Also, the iOS plugin class still uses error/success instead of reject/resolve.

And a lot of the notification classes use PluginCall and JSObject classes, since we wanted users to be able to use the plugins independently, I think that should be reworked. Notifications was one of the plugins where users were more interested in reusing.

I've noticed that a lot of type names have been renamed, don't forget to document those changes. But also, I'm concerned that as PushNotifications plugin also shares a few of the methods, the auto import of editors such us VS Code could lead to problems if we use the same names.

local-notifications/ios/Plugin/LocalNotificationsDelegate.swift
dwieeb: I missed my own TODO 😅

dwieeb:
> Also, the iOS plugin class still uses error/success instead of reject/resolve.

Fixed!

> since we wanted users to be able to use the plugins independently, I think that should be reworked.

I agree, but I think we should do it later as a feature. It's just a nice-to-have and we wanted to move a bit quicker with these plugins. I will create an issue to remind us if you agree.

> I've noticed that a lot of type names have been renamed, don't forget to document those changes

Instead, I added them all back with `@deprecated` to make it easier to migrate.

> the auto import of editors such us VS Code could lead to problems if we use the same names

Possibly, but these are self-contained packages and I don't think we should prefix every export with the name of the package--it makes it more difficult to read in my opinion. It's also difficult to import the plugin class correctly (`LocalNotifications`, not `LocalNotification` or any `LocalNotification...` imports in the list as you try to autocomplete it). I suppose it's going to remain difficult until this plugin reaches 2.0 because of all the deprecated types

jcesarmobile:
> > I've noticed that a lot of type names have been renamed, don't forget to document those changes
> 
> Instead, I added them all back with `@deprecated` to make it easier to migrate.
> 
> > the auto import of editors such us VS Code could lead to problems if we use the same names
> 
> Possibly, but these are self-contained packages and I don't think we should prefix every export with the name of the package--it makes it more difficult to read in my opinion. It's also difficult to import the plugin class correctly (`LocalNotifications`, not `LocalNotification` or any `LocalNotification...` imports in the list as you try to autocomplete it). I suppose it's going to remain difficult until this plugin reaches 2.0 because of all the deprecated types

Then we should change the types of the other plugins, all of them are prefixed at the moment

dwieeb:
@carlpoole I need your suggestion for the error in CI:

```
LocalNotificationRestoreReceiver.java:53: error: no suitable constructor found for CapConfig(AssetManager,<null>)
        CapConfig config = new CapConfig(context.getAssets(), null);
```

I can't use `CapConfig.fromFile` because it's package-private and we don't want plugins creating new instances, but what do I do in this case? I am using a `BroadcastReceiver` which doesn't have the same context as regular plugin APIs: https://github.com/ionic-team/capacitor-plugins/pull/94/files#diff-f2f4575f3cb40956c43792baa18bce30af06c925ac87b25c2a92ec7acafc4599R17-R57

carlpoole (approved):

local-notifications/ios/Plugin/LocalNotificationsDelegate.swift
```diff
@@ -0,0 +1,87 @@
+import Capacitor
+import UserNotifications
+
+public class LocalNotificationsDelegate: NSObject, NotificationHandlerProtocol {
+
+    public var plugin: CAPPlugin?
```
ikeith, line 6: Retain cycle issue.
```suggestion
    public weak var plugin: CAPPlugin?
```

local-notifications/ios/Plugin/LocalNotificationsDelegate.swift
```diff
@@ -0,0 +1,87 @@
+import Capacitor
+import UserNotifications
+
+public class LocalNotificationsDelegate: NSObject, NotificationHandlerProtocol {
```
ikeith, line 4: Similarly to push notifications, I think this class should be renamed since the term `delegate` can cause confusion.

dwieeb:
LGTM

thomasvidas (approved):

ikeith (approved):

PR #97, opened 2020-11-11
feat(share): support sharing file with text

yarinvak:
add support for sharing file combined with a text


jcesarmobile:
Thanks for the PR, but I've tried and I can share files with text without problems.
Please, when creating PRs that fix an issue, if the issue has not been reported, provide a sample app where it can be reproduced.

PR #99, opened 2020-11-12
chore: remove unused layout files

dwieeb:
see https://github.com/ionic-team/create-capacitor-plugin/pull/29


jcesarmobile (changes requested):
I think the idea of having the `res/.gitkeep` in the template is ok so users can see the folder if they need to add some resource, but we don't need the res folder for any of the plugins, so we can delete it

dwieeb:
Agreed!

jcesarmobile (approved):

PR #100, opened 2020-11-12
chore: update gradle to 6.7

dwieeb:
Gradle can be updated in the future by running this:

```
brew update
brew upgrade
npx lerna exec --ignore @capacitor/motion 'cd android && gradle wrapper --distribution-type all'
```

Note: this still is doing this in CI:
```
Downloading https://services.gradle.org/distributions/gradle-6.7-all.zip
```


jcesarmobile:
We usually update to the latest that Android Studio recommends, not to the latest latest.

At this moment, Android Studio is recommending updating Android Gradle plugin to version 4.1.1 and Gradle to version 6.5

dwieeb:
I mostly made this PR to test if CI would still download 6.7. I can make a PR for the 6.5 upgrade if you'd like.

PR #106, opened 2020-11-22
chore: update gradle to 6.5

dwieeb:
Update gradle in the plugins to the version that Android Studio recommends.

Gradle can be updated in the future by running this:

```
npx lerna exec --concurrency 1 --ignore @capacitor/motion 'cd android && ./gradlew wrapper --gradle-version 6.5 --distribution-type all'
```

Then, update all the gradle plugins in `build.gradle` files.

ref: https://github.com/ionic-team/capacitor-plugins/pull/100#issuecomment-726806441


jcesarmobile (changes requested):
You also have to update the gradle plugin to latest (I think it’s 4.1.1, but verify)
https://github.com/ionic-team/capacitor-plugins/blob/main/action-sheet/android/build.gradle#L15

jcesarmobile (approved):

PR #109, opened 2020-11-23
feat(device): add browser/webview version to getInfo()

### LINKED ISSUE #91: feat: output WebView version on DeviceInfo
### Description

It would be nice to know the current WebView version on a device.
The device plugin already returns several information about the device - like OS version etc.

This feature request is about extending the device information with the current WebView version.

Android pseudo code
```
PackageInfo pInfo = WebView.getCurrentWebViewPackage();
r.put("webViewVersion", pInfo.versionName);  // e.g. 69.0.3497.100
```

### Platform(s)

Android
iOS

### Preferred Solution

- extend given device plugin with property "webViewVersion" of type string

### Alternatives

- alternatively a own plugin, but imho this would be overhead

thomasvidas:
Added `browserEngine` and `webViewVersion` to the device plugin. On Android it uses `WebView.getCurrentWebViewPackage();` and on iOS we just use the OS version. For web, there is some user agent parsing to determine the browser and version.

Browser engine is the browser name: Chrome, Firefox, Edge, etc.
WebView version is the browser major version. Not the underlying WebKit or Blink engine.

Closes #91


device/android/src/main/java/com/capacitorjs/plugins/device/Device.java
```diff
@@ -3,11 +3,14 @@
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.content.pm.PackageInfo;
 import android.os.BatteryManager;
 import android.os.Build;
 import android.os.Environment;
 import android.os.StatFs;
 import android.provider.Settings;
+import android.webkit.WebSettings;
```
jcesarmobile, line 12:
```suggestion
```
unused

jcesarmobile (changes requested):
Requested a minor change

I've been testing and `webViewVersion` is very inconsistent on web, I get the major version only, while on native it includes the whole version (i.e. on iOS 13.6.1 I get 13.6.1 on native, but 13 on web).
On iOS chrome it returns 604
On Desktop Chrome both `webViewVersion` and `browserEngine` are empty.

Also, why is `browserEngine` optional? I think it should always return something.
And I find the name very confusing, when I think of a browser engine I think in the engine it internally uses (i.e. Gecko on Firefox)

I personally don't find `browserEngine` useful because it's a string and can return anything, so it's extremely complex to use it to do checks as you don't know what will be returned.

thoughts @dwieeb?

device/android/src/main/java/com/capacitorjs/plugins/device/Device.java
```diff
@@ -78,4 +81,22 @@ public String getName() {
 
         return null;
     }
+
+    public String getWebViewVersion() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            PackageInfo info = WebView.getCurrentWebViewPackage();
+            return info.versionName;
+        }
```
jcesarmobile, line 89:
```suggestion
        PackageInfo info = null;
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
            info = WebView.getCurrentWebViewPackage();
        } else {
            PackageManager pm = this.context.getPackageManager();
            try {
                info = pm.getPackageInfo("com.google.android.webview", 0);
            } catch (PackageManager.NameNotFoundException e) {}
        }
        if (info != null) {
            return info.versionName;
        }
```

With this code you can also get the version on Android 5-6 devices, so only returns `Build.VERSION.RELEASE`, but ideally we should find a way of getting it on Android 7 too, because `Build.VERSION.RELEASE` is not helpful

dwieeb:
I think `browserEngine` is _somewhat_ useful on Android, not useful on iOS, and will be difficult to maintain on web. There are alternatives (such as [Ionic Framework's platform detection](https://ionicframework.com/docs/core-concepts/cross-platform#platform-detection)) that people are already using. I vote for removing `browserEngine` from this PR at this time.

I can't get `webViewVersion` to work on desktop Chrome or Firefox.

dwieeb (commented):
- `browserEngine` is still sprinkled in with the web code
- you may need to `npm install` and try formatting again. we shouldn't be seeing the `uuid4` diff

thomasvidas:
Thought it might be useful in case we wanted to use it in the future, but you're right, no reason to keep it!

device/src/web.ts
```diff
@@ -110,6 +111,30 @@ export class DeviceWeb extends WebPlugin implements DevicePlugin {
...
+        /(opera|chrome|safari|firefox|msie|trident(?=\/))\/?\s*(\d+)/i,
+      ) || [];
+    let versions: RegExpMatchArray | RegExpExecArray | null;
+    if (/trident/i.test(match[1])) {
+      versions = /\brv[ :]+(\d+)/g.exec(_ua) || [];
+      uaFields.browserVersion = versions[1] || '';
+    } else if (match[1] === 'Chrome') {
+      versions = _ua.match(/\bOPR|Edge\/(\d+)/);
+      if (versions !== null) {
+        uaFields.browserVersion = versions[1] || '';
```
jcesarmobile, line 125: should probably return `uaFields` here, otherwise the value gets overwritten on line 136
same for line 121, but I don't thing we support internet explorer, that's what trident stands for, right? that can probably be removed, thoughts @dwieeb?

should it try to get the full version instead of the major version as it does now? on iOS/Android it returns the full version

thomasvidas: Trident is IE yes, it might be some older versions of Edge too but I'd have to double check. If we don't support IE11 then we can remove that .

All of the major browsers are versioned using only the major version. Firefox, Chrome, and Edge don't use minor versions anymore when releasing new versions, it is just related to the build number. Really only the major version is somewhat useful. I am not sure about Safari, that might still have the minor version.

jcesarmobile: It's not about usefulness but about consistency, since android and iOS return them.
If not useful, maybe android and iOS should return just the major?

thomasvidas: True, that makes sense. I will change web to include the minor version because it seems desktop and mobile Safari both still use the minor version pretty actively compared to other browsers

thomasvidas:
This is changed on web now, was waiting for `alpha.7` release due to build issues I was getting.

Rather than using messy regex. I changed the code to use feature detection to detect browsers. 
```typescript
const isFirefox = !!(window as any).InstallTrigger;
const isSafari = !!(window as any).ApplePaySession;
const isChrome = !!(window as any).chrome;
const isEdge = /Edge/.test(navigator.userAgent);
```

From there, I just use string manipulation. 
```typescript
    // FF and Edge User Agents both end with "/MAJOR.MINOR"
    if (isFirefox || isEdge) {
      const reverseUA = ua.split('').reverse().join('');
      const reverseVersion = reverseUA.split('/')[0];
      const version = reverseVersion.split('').reverse().join('');
      uaFields.browserVersion = version;
    } else if (isSafari || isChrome) {
      // Safari version comes after "... Version/MAJOR.MINOR ...""
      // Chrome version comes after "... Chrome/MAJOR.MINOR ...""
      const searchWord = isChrome ? 'Chrome' : 'Version';
      const words = ua.split(' ');
      words.forEach(word => {
        if (word.includes(searchWord)) {
          const version = word.split('/')[1];
          uaFields.browserVersion = version;
        }
      });
    }
```

Below is a screenshot. Left is Firefox, Middle is Chrome, Right is Safari. Verified it is working on Android Chrome and iOS Safari on web as well.

Screen Shot 2020-12-03 at 10 59 35 AM

device/src/web.ts
```diff
@@ -106,6 +107,31 @@ export class DeviceWeb extends WebPlugin implements DevicePlugin {
       uaFields.operatingSystem = 'unknown';
     }
 
+    // Check for browsers based on non-standard javascript apis, not user agent
+    const isFirefox = !!(window as any).InstallTrigger;
+    const isSafari = !!(window as any).ApplePaySession;
+    const isChrome = !!(window as any).chrome;
```
dwieeb, line 113: It is preferred to use module augmentation on global types instead of casting them to `any`. You can do this by adding to the `declare global` at the top. Something like this:

```ts
declare global {
  interface Window {
    InstallTrigger?: any;
    ApplePaySession?: any;
    chrome?: any;
  }

  ...
}
```

dwieeb (commented):
I think the change to feature detection is fine--but we really need some unit tests in here to test the user agent parsing.

device/src/web.ts
```diff
@@ -106,6 +113,54 @@ export class DeviceWeb extends WebPlugin implements DevicePlugin {
...
+      } else if (isEdgeIOS) {
+        searchWord = 'EdgiOS';
+      } else if (isSafari) {
+        searchWord = 'Version';
+      } else if (isChrome) {
+        searchWord = 'Chrome';
+      }
+
+      const words = ua.split(' ');
+      words.forEach(word => {
```
dwieeb, line 151: `.forEach` is discouraged. (It's a remnant of pre-ES6 days when `for..of` wasn't a thing).

`for (const word of words) {` is preferred 😊

device/package.json
```diff
@@ -27,6 +27,7 @@
     "native"
   ],
   "scripts": {
+    "test": "uvu -r esm -r ts-node/register src/__tests__",
```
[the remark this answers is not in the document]
thomasvidas: Yeah, I added `esm` since ts-node doesn't have esm support: https://github.com/TypeStrong/ts-node/issues/1007 It should be fine for our use case.

Alternatively, we could
- run `build` as a pretest script
- write the tests in javascript
- use node's native module support instead of esm

jcesarmobile (approved):
looks good

PR #110, opened 2020-11-24
chore: Add example unit tests to every plugin

thomasvidas:
- Added jest dependency to each plugin
- Add example.ts test to expand upon
- Updated plugin dependencies to match each other


dwieeb (approved):

carlpoole (approved):

dwieeb:
I'm not sure we should merge this yet. We may want to wait until more plugins are merged into `main`.

thomasvidas:
Based on the discussion we had yesterday, I'll be migrating this from jest to [uvu](https://github.com/lukeed/uvu) and [sinon](https://github.com/sinonjs/sinon)

jcesarmobile:
I'm going to close. We should have tests where it makes sense, but adding a 2+2 = 4 tests on every plugin will only delay the CI execution, which is pretty slow already

PR #115, opened 2020-11-29
feat(keyboard): add types for config files

dwieeb:
Add the keyboard config types for `capacitor.config.ts` files.

ref: https://github.com/ionic-team/capacitor/pull/3858/commits/79334e4275aa2ce1c07f3337ed7c7f1be7e7f018


jcesarmobile (changes requested):
Comments should mention the configurations are iOS only
https://capacitorjs.com/docs/apis/keyboard#keyboard-configuration-ios-only

jcesarmobile (approved):

PR #138, opened 2020-12-14
chore(screen-reader): rename screenReaderStateChange to stateChange

imhoffd:
- [x] docgen needs to support `@deprecated`
- [ ] other events?


jcesarmobile:
I think this can be closed. It's a new plugin, so I don't think we need to have backward compatibility with the old plugin event name, just document the change in the upgrade guide.

jcesarmobile (approved):

PR #140, opened 2020-12-14
fix(android): Migrate plugins from Color.parseColor() to WebColor.parseColor()

thomasvidas:
Depends on https://github.com/ionic-team/capacitor/pull/3947

Change all Color.parseColor() to WebColor.parseColor() to make sure that Colors from web to Android are correctly parsed. Android format is #AARRGGBB and Web Format is #RRGGBBAA


jcesarmobile:
I think only the colors that come from the web part should use the new WebColor (meaning, `ActionSheet.java` doesn't need those changes)

Or in case we want all colors to be `WebColors`, https://github.com/ionic-team/capacitor-plugins/pull/33 should be updated as it has a clone class of `ActionSheet.java`

jcesarmobile (approved):
looks good

you should probably add a note about it on the upgrade guide
https://github.com/ionic-team/capacitor/issues/3320

thomasvidas:
Yeah I'm planning on it 👍

PR #143, opened 2020-12-15
chore: update to capacitor alpha 10

jcesarmobile:
update all dependencies, not just android


jcesarmobile:
looks like only android was published

Issue #133, opened 2020-10-19
Memory leak?


aparajita:
https://github.com/ionic-team/capacitor/blob/c3e821ef2a4cf20c31b3b0bd6b9ca95c717b0219/android/capacitor/src/main/java/com/getcapacitor/Splash.java#L41-L42

Android Studio is giving me the following warning about these lines:

> Do not place Android context classes in static fields; this is a memory leak

PR #149, opened 2020-12-18
feat: Add SplashScreen plugin

jcesarmobile:
closes https://github.com/ionic-team/capacitor-plugins/issues/133
closes https://github.com/ionic-team/capacitor-plugins/issues/150


splash-screen/src/web.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import type {
+  SplashScreenPlugin,
+  SplashScreenShowOptions,
+  SplashScreenHideOptions,
+} from './definitions';
+
+export class SplashScreenWeb extends WebPlugin implements SplashScreenPlugin {
+  async show(_options?: SplashScreenShowOptions): Promise<void> {
+    return undefined;
```
dwieeb, line 11:
```suggestion
    throw this.unimplemented('Not implemented for web.');
```

splash-screen/src/web.ts
```diff
@@ -0,0 +1,17 @@
...
+  }
+
+  async hide(_options?: SplashScreenHideOptions): Promise<void> {
+    return undefined;
```
dwieeb, line 15:
```suggestion
    throw this.unimplemented('Not implemented for web.');
```

splash-screen/CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
+# Contributing
```
dwieeb, line 1: This file can be removed

splash-screen/src/web.ts
jcesarmobile: it was like that, so I didn't want to change the behavior

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
```diff
@@ -0,0 +1,354 @@
...
+import com.getcapacitor.Logger;
+
+/**
+ * A Splash Screen service for showing and hiding a splash screen in the app.
+ */
+public class SplashScreen {
+
+    private ImageView splashImage;
+    private ProgressBar spinnerBar;
+    private WindowManager wm;
```
carlpoole, line 28: rename to windowManager

carlpoole:
Couple of ideas

`SplashScreenConfig` and `SplashScreenSettings` could use Builders so the resulting object is immutable - it looks like those are only accessed inside SplashScreen and not set there.

the `show` method in SplashScreen might also benefit from a method signature without a Settings parameter so it just uses a default SplashScreenSettings. (similar to the old one that just took an activity like this:
 `public static void show(final Activity a) {		
         show(a, DEFAULT_LAUNCH_SHOW_DURATION, DEFAULT_FADE_IN_DURATION, DEFAULT_FADE_OUT_DURATION, DEFAULT_AUTO_HIDE, null, null);		
     }`)

jcesarmobile:
As the idea of the implementation class is to make it easier to reuse the plugin by native code, it's easier to have a single object rather than a bunch of params.
Also, iOS had a TODO about the show having too many parameters, which wasn't good for the linter and the TODO comment suggested to use a struct, so I did. So on Android I used classes since there are no structs.

My code is based on Ian's approach on camera PR, which was approved by you and Dan, so if we want to use builders his PR should also be changed.

I removed the show with no params because it was not being used and iOS didn't have an equivalent, but I can put it back if you want. But then iOS should also have something like that for consistency.

carlpoole:
> I removed the show with no params because it was not being used and iOS didn't have an equivalent, but I can put it back if you want. But then iOS should also have something like that for consistency.

I'm thinking from a reusability standpoint for this method:

```show(final AppCompatActivity activity, final SplashScreenSettings settings, final SplashListener splashListener)```

You would call it:

`show(activity, new SplashScreenSettings(), listener)`

or you could also do

`show(activity, listener)` 

if you just want defaults. If the first one is acceptable then that's fine.

dwieeb (approved):

PR #151, opened 2020-12-21
feat(browser): Add close() method for android

galaharon11:
I'm new to ionic and capacitor and I had a problem I couldn't solve: Browser cannot be closed on android. I found out that you can't close CustomTabs on android. After searching for a while, I found out an optional workaround for that problem (based on an this answer on StackOverflow: https://stackoverflow.com/a/41596629/5976838).
The workaround works perfectly for me so I uploaded it here.


dwieeb:
Hi @galaharon11, would you be able to run `npm run fmt` to reformat the Java so CI passes?

sathishsundar05:
@galaharon11  your solution is working thanks but,

Missed to add the below imports,

import android.content.Intent;
import android.content.Context;

PR #157, opened 2020-12-22
fix(share): share iOS images as UIImages

### LINKED ISSUE #156: bug: Share plugin seems to ignore instagram
### Capacitor Version
Latest Dependencies:

  @capacitor/cli: 2.4.5
  @capacitor/core: 2.4.5
  @capacitor/android: 2.4.5
  @capacitor/electron: 2.4.5
  @capacitor/ios: 2.4.5

Installed Dependencies:

  @capacitor/cli 2.3.0
  @capacitor/core 2.4.3
  @capacitor/android 2.4.5
  @capacitor/ios 2.4.5
  @capacitor/electron not installed

  Found 21 Capacitor plugins for ios:
    @capacitor-community/firebase-analytics (0.2.0)
    @ionic-enterprise/auth (3.3.1)
    @ionic-enterprise/identity-vault (4.2.7)
    @ionic-enterprise/offline-storage (2.0.2)
    cordova-plugin-advanced-http (3.1.0)
    cordova-plugin-app-version (0.1.9)
    cordova-plugin-badge (0.8.8)
    cordova-plugin-camera (4.1.0)
    cordova-plugin-device (2.0.3)
    cordova-plugin-file (6.0.2)
    cordova-plugin-file-opener2 (3.0.5)
    cordova-plugin-file-transfer (1.7.1)
    cordova-plugin-inappbrowser (4.1.0)
    cordova-plugin-iroot (2.1.0)
    cordova-plugin-network-information (2.0.2)
    cordova-plugin-pincheck (0.0.6)
    cordova-plugin-streaming-media (2.3.0)
    cordova-plugin-whitelist (1.3.4)
    cordova-plugin-x-socialsharing (6.0.2)
    es6-promise-plugin (4.2.2)
    cordova-plugin-calendar (5.1.5)
[success] iOS looking great! 👌

### Platform(s)
- IOS (Version 14)
- Android (Version 8)

### Current Behavior
I need to use the Share plugin to share some stuff through social media.
I tried with the method from the docs:
`
import { Plugins } from '@capacitor/core';
const { Share } = Plugins;
`
...
`async share() {
        const shareRet = await Share.share({
            title: 'See cool stuff',
            text: 'Really awesome thing you need to see right meow',
            url: 'http://ionicframework.com/',
            dialogTitle: 'Share with buddies'
        });
    }`

When I call share() from a template, the native modal is correctly opened and it contains most social media apps on the device, _**except instagram**_, which is the issue because I need it.

### Expected Behavior
Instagram should be present in the options

### Code Reproduction
`
import { Plugins } from '@capacitor/core';
const { Share } = Plugins;
`
...
`async share() {
        const shareRet = await Share.share({
            title: 'See cool stuff',
            text: 'Really awesome thing you need to see right meow',
            url: 'http://ionicframework.com/',
            dialogTitle: 'Share with buddies'
        });
    }`

jcesarmobile:
At the moment share plugin shares images as urls and some apps are not able to share them properly or don't appear at all like instagram.

This PR try to convert the url to UIImage and if possible share it, otherwise it shares the url as before

closes #156

### DIFF
--- a/share/ios/Plugin/SharePlugin.swift
+++ b/share/ios/Plugin/SharePlugin.swift
@@ -12,7 +12,11 @@ public class SharePlugin: CAPPlugin {
         }
 
         if let url = call.getString("url"), let urlObj = URL(string: url) {
-            items.append(urlObj)
+            if let imageData = try? Data(contentsOf: urlObj), let image = UIImage(data: imageData) {
+                items.append(image)
+            } else {
+                items.append(urlObj)
+            }
         }
 
         let title = call.getString("title")


share/ios/Plugin/SharePlugin.swift
```diff
@@ -12,7 +12,11 @@ public class SharePlugin: CAPPlugin {
         }
 
         if let url = call.getString("url"), let urlObj = URL(string: url) {
-            items.append(urlObj)
+            if let imageData = try? Data(contentsOf: urlObj), let image = UIImage(data: imageData) {
```
ikeith, line 15: I'm concerned that this approach is too naive. We should not blindly try to load any URL without knowing what it is pointing to. If it's a giant video file, large pdf or simply on the other side of a slow network connection, it will block the thread while loading before it can even figure out if it's actually an image. And if the data is too large, it could run out of RAM and cause memory pressure problems for the whole app.

It either needs to be much smarter about checking the origin of the URL, guessing content type, and switching to a background thread when necessary, or it should put that burden on the app developer by requiring an additional parameter to know when to attempt this.

jcesarmobile: what about doing it for local files only?

ikeith: I think only trying with file urls would be a lot safer and less likely to cause side effects. I'd also suggest checking the extension for image types.

That does mean that developers would need to download and save an image first if they wanted this behavior but as long as we document the rules I think that's a reasonable compromise.

jcesarmobile: There are a lot of cases where the image doesn't have a file extension, so checking for it can be useless, but most times you can get the mime type from the url or from the data, so probably better to get it from there.

PR #165, opened 2020-12-28
fix(camera): add storage permissions to plugin's AndroidManifest.xml

dwieeb:
I think we missed these permissions for the Camera plugin. This will allow app developers to remove these permissions from their app's `AndroidManifest.xml` files.


dwieeb:
Oops, these are "dangerous" permissions, so the app developer should choose to include them in their manifests.

PR #175, opened 2021-01-07
feat(android): permission requesting using AndroidX Activity 1.2 flow updates 

carlpoole:
See: https://github.com/ionic-team/capacitor/pull/4033

These updates are compatible with the AndroidX Activity 1.2 permission request flow updates when merged with Capacitor

**Plugins Updated**:
- Filesystem
- Camera
- Geolocation


geolocation/android/src/main/java/com/capacitorjs/plugins/geolocation/GeolocationPlugin.java
```diff
@@ -47,6 +44,68 @@ public void getCurrentPosition(final PluginCall call) {
...
+                    public void success(Location location) {
+                        call.resolve(getJSObjectForLocation(location));
+                    }
+
+                    @Override
+                    public void error(String message) {
+                        call.reject(message);
+                    }
+                }
+            );
```
dwieeb, line 73: Can you use `getPosition()` here?

carlpoole: Potentially. I noticed it, and the boolean in the `sendLocation` call made here differs from the one in `getPosition()`, and I didn't want to start digging into that improvement here in the scope of permissions refactoring.

PR #204, opened 2021-01-20
docs(android): document plugin variables

dwieeb:
I'm not sure if we should document ALL the plugin variables in EVERY plugin, or just the unique ones.

### DIFF
--- a/action-sheet/README.md
+++ b/action-sheet/README.md
@@ -9,6 +9,12 @@ npm install @capacitor/action-sheet
 npx cap sync
 ```
 
+### Variables
+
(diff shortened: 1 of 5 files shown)


jcesarmobile:
I think it's ok to only put the ones specific for the plugin as the others are in capacitor/android and are included by default in `variables.gradle`

dwieeb:
Okay 👍 

I think we should document the capacitor/android variables somewhere, too 👉 https://github.com/ionic-team/capacitor-site/issues/208

jcesarmobile (changes requested):
I think it should say the default version they use

jcesarmobile (approved):

Issue #208, opened 2021-01-22
Push notification is shown when the app is in foreground


jcesarmobile:
### Plugin(s)
push-notifications

### Capacitor Version
3

### Platform(s)
ios

### Current Behavior
The push notification is shown when the app is in foreground

### Expected Behavior
In Capacitor 2, the push notification was not show when the app is in foreground

### Code Reproduction
https://github.com/ionic-team/capacitor-site/pull/210

kamekaioken:
On iOS The Push plugin version 0.3.10 is always showing (again?) the notification alert. Sound is mute, but the alert is shown ignoring the [configuration](https://capacitorjs.com/docs/v3/apis/push-notifications#push-notifications-appearance-in-foreground).
I am using the Plugin on iOS with Firebase as described [here](https://capacitorjs.com/docs/v3/guides/push-notifications-firebase).

PR #209, opened 2021-01-22
fix(ios): do not show notifications when in foreground

jcesarmobile:
In Capacitor 2, push notifications where not shown when the app was in foreground (by default, but can be configured to show them)
This PR changes the behavior to do the same

closes https://github.com/ionic-team/capacitor-plugins/issues/208

### DIFF
--- a/push-notifications/ios/Plugin/PushNotificationsHandler.swift
+++ b/push-notifications/ios/Plugin/PushNotificationsHandler.swift
@@ -49,7 +49,7 @@ public class PushNotificationsHandler: NSObject, NotificationHandlerProtocol {
             return presentationOptions
         }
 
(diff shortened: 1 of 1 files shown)


dwieeb:
I think this was intentional. @theproducer was this to fix a reported issue?

theproducer:
No, it wasn't based on any reported issue.

>push notifications where not shown when the app was in foreground **(by default, but can be configured to show them)**

This I did not realize, sorry!

theproducer (approved):
I see the mistake I made: returning `presentationOptions` returns the config options for how to display notifications - outside of that I should simply returned an empty options set.

dwieeb (approved):

PR #211, opened 2021-01-22
feat(local-notifications): add more info to pending notifications

theproducer:
This PR adds more useful information to the array of returned pending notification.

closes https://github.com/ionic-team/capacitor-plugins/issues/193


local-notifications/src/definitions.ts
```diff
@@ -163,6 +163,42 @@ export interface LocalNotificationDescriptor {
...
+   */
+  extra?: any;
+
+  /**
+   * Repeat delivery of this notification at the date and time specified by
+   * `at`.
+   *
+   * @since 1.0.0
+   */
+  repeats?: boolean;
```
dwieeb, line 201: I thought this property was a part of `Schedule`, why the duplicate here?

local-notifications/src/definitions.ts
```diff
@@ -163,6 +163,42 @@ export interface LocalNotificationDescriptor {
    * @since 1.0.0
    */
   id: string;
+
+  /**
+   * The title of the notification.
+   *
+   * @since 1.0.0
+   */
+  title?: string;
```
dwieeb, line 172: `title` isn't optional when creating notifications, I don't think it can be undefined

local-notifications/src/definitions.ts
```diff
@@ -163,6 +163,42 @@ export interface LocalNotificationDescriptor {
...
+
+  /**
+   * The body of the notification, shown below the title.
+   *
+   * @since 1.0.0
+   */
+  body?: string;
```
dwieeb, line 179: `body` isn't optional when creating notifications, I don't think it can be undefined

local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationSchedule.java
```diff
@@ -37,16 +37,19 @@ public LocalNotificationSchedule() {}
     private void buildEveryElement(JSObject schedule) {
         // 'year'|'month'|'two-weeks'|'week'|'day'|'hour'|'minute'|'second';
         this.every = schedule.getString("every");
+        if (this.every != null) {
+            this.repeats = true;
+        }
```
dwieeb, line 42: What is this change for?

local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationSchedule.java
```diff
@@ -37,16 +37,19 @@ public LocalNotificationSchedule() {}
...
 
     private void buildCountElement(JSObject schedule) {
         this.count = schedule.getInteger("count", 1);
     }
 
     private void buildAtElement(JSObject schedule) throws ParseException {
-        this.repeats = schedule.getBool("repeats");
         String dateString = schedule.getString("at");
         if (dateString != null) {
+            this.repeats = schedule.getBool("repeats");
```
carlpoole, line 52: Does moving this inside the if check on `at` prevent the `repeats` flag being set for notifications that are built with the `on` details in `buildOnElement`? I noticed these two methods are called sequentially but for some reason the repeats flag was in here.

local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationSchedule.java
theproducer: In my mind, having a schedule that states something will happen every X implies that this is a repeating notification.

jcesarmobile:
I think that instead of editing the LocalNotificationDescriptor type, they should return LocalNotificationSchema type. Not really sure if it’s worth it for schedule since it will just return what the user sent.
cancel should keep the LocalNotificationDescriptor with just the id, we can’t ask users to provide a full notification when we just need the id for canceling.

local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationSchedule.java
jcesarmobile: According to the docs, repeats is only used for at, so I don’t think we should change its value to something different than the user set. And also the change is unrelated to the original intent of the PR, I would try to avoid unrelated changes. If not setting repeats to true here can cause issues with every, it should be reported and evaluated in a separate issue.

local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotification.java
```diff
@@ -247,18 +247,28 @@ public static LocalNotification buildNotificationFromJSObject(JSObject jsonObjec
...
+                jsSchedule.put("count", schedule.getCount());
+                jsSchedule.put("on", schedule.getOn());
+                jsSchedule.put("repeats", schedule.isRepeating());
+                jsNotification.put("schedule", jsSchedule);
             }
-            jsArray.put(notification);
+
+            jsNotification.put("extra", notification.getExtra());
+
+            jsArray.put(jsNotification);
```
jcesarmobile, line 271: This is missing some notification information.
There is an easier way of getting the original notification as JSObject, I can't put it as suggested change but should be something like
```
try {
    JSObject jsNotification = new JSObject(notification.getSource());
    jsArray.put(jsNotification);
} catch (JSONException e) {}
```

theproducer: On iOS we do not store the original notification object used to create the actual system notification (maybe its worth doing in a separate PR), so I limited what gets returned in the request to the fields that are most likely to be relevant and common across both iOS and Android:

From https://github.com/ionic-team/capacitor-plugins/issues/193
> Would it be possible to create a new feature that takes an ID and return an object containing details such as **title, body, and schedule** for a given local notif?

local-notifications/ios/Plugin/LocalNotificationsPlugin.swift
```diff
@@ -540,9 +545,36 @@ public class LocalNotificationsPlugin: CAPPlugin {
...
+            notification["extra"] = extra
+        }
+
+        if let notificationSchedule = request.content.userInfo["cap_schedule"] as? [String: Any?] {
+            var schedule: JSObject = [:]
+            for(key, value) in notificationSchedule {
+                schedule[key] = value as? JSValue
+            }
+            notification["schedule"] = schedule
+        }
```
ikeith, line 574: Suggested changes as referenced in my comment on https://github.com/ionic-team/capacitor/pull/4133


```suggestion
        let userInfo = JSTypes.coerceDictionaryToJSObject(request.content.userInfo)
        let notificationExtras = userInfo["cap_extra"] as? JSObject ?? userInfo

        if let extra = notificationExtras {
            notification["extra"] = extra
        }

        if let schedule = userInfo["cap_schedule"] as? JSObject {
            notification["schedule"] = schedule
        }
```

local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotification.java
jcesarmobile: On Android we can return everything, and it's easier to return everything, but if we don't want to return everything because it's not possible on iOS, the it's probably better to not return LocalNotificationSchema as if we return that, users would expect to get the whole LocalNotificationSchema that was originally sent to schedule method.

theproducer: Good point

carlpoole (approved):

ikeith (approved):

PR #222, opened 2021-01-28
feat(android): implements Activity Result API changes for permissions and activity results

carlpoole:
Updates to implement changes to the Activity Results API and permission requests, see: https://github.com/ionic-team/capacitor/pull/4127

- Removes the `permissionCallback` element from the `@PluginMethod` annotation. The permission result callback methods are now annotated directly with `@PermissionCallback` and the method name is passed in at request time.
- Removed request codes and added `@ActivityCallback` methods.

### DIFF
--- a/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationsPlugin.java
+++ b/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationsPlugin.java
@@ -1,10 +1,12 @@
 package com.capacitorjs.plugins.localnotifications;
 
 import android.content.Intent;
+import com.getcapacitor.Bridge;
 import com.getcapacitor.JSArray;
 import com.getcapacitor.JSObject;
 import com.getcapacitor.Plugin;
 import com.getcapacitor.PluginCall;
+import com.getcapacitor.PluginHandle;
 import com.getcapacitor.PluginMethod;
 import com.getcapacitor.annotation.CapacitorPlugin;
 import com.getcapacitor.annotation.Permission;
@@ -15,6 +17,7 @@ import org.json.JSONArray;
 @CapacitorPlugin(name = "LocalNotifications", permissions = @Permission(strings = {}, alias = "display"))
 public class LocalNotificationsPlugin extends Plugin {
 
+    private static Bridge staticBridge = null;
     private LocalNotificationManager manager;
     private NotificationStorage notificationStorage;
     private NotificationChannelManager notificationChannelManager;
@@ -26,6 +29,7 @@ public class LocalNotificationsPlugin extends Plugin {
         manager = new LocalNotificationManager(notificationStorage, getActivity(), getContext(), this.bridge.getConfig());
         manager.createNotificationChannel();
         notificationChannelManager = new NotificationChannelManager(getActivity());
+        staticBridge = this.bridge;
     }
 
     @Override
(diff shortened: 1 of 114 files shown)


local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationsPlugin.java
```diff
@@ -44,12 +44,6 @@ protected void handleOnNewIntent(Intent data) {
         }
     }
 
-    @Override
-    protected void handleOnActivityResult(PluginCall savedCall, int requestCode, int resultCode, Intent data) {
```
carlpoole, line 48: @dwieeb when I was updating this plugin it looks like this result callback wasn't used. I looked into it but couldn't see where this might be reached. Since you looked over this code recently, do you know if that's true? I think this is safe to delete.

dwieeb: Yeah, it should be.

share/android/src/main/java/com/capacitorjs/plugins/share/SharePlugin.java
```diff
@@ -89,7 +100,7 @@ public void share(PluginCall call) {
         if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP_MR1) {
             PendingIntent pi = PendingIntent.getBroadcast(
                 getContext(),
-                REQUEST_SHARE,
+                123,
```
dwieeb, line 103: Could you explain a little more about what's going on here?

dwieeb (commented):
Exciting stuff! 😊

share/android/src/main/java/com/capacitorjs/plugins/share/SharePlugin.java
carlpoole: The request code chosen here isn't important for the result behavior. It's helpful if you need to differentiate pending intents, but the result behavior we perform in `activityResult` doesn't check request codes for its purpose. I replaced it with a magic number here which probably isn't ideal as just `123` so I could change it to `0` and maybe put a comment here so it isn't confusing?

dwieeb: Thanks!

> so I could change it to `0` and maybe put a comment here so it isn't confusing?

Agreed! Would be helpful for the next person who reads the code.

share/android/src/main/java/com/capacitorjs/plugins/share/SharePlugin.java
```diff
@@ -37,6 +37,17 @@ public void onReceive(Context context, Intent intent) {
         }
     }
 
+    @ActivityCallback
+    private void activityResult(PluginCall call, ActivityResult result) {
+        if (result.getResultCode() == Activity.RESULT_CANCELED) {
```
jcesarmobile, line 42:
```suggestion
        if (result.getResultCode() == Activity.RESULT_CANCELED && !stopped) {
```
It's missing the ` && !stopped` check. Without it, when you open an app to complete the share and get back to the app the promise is rejected instead of resolving with the app package name.

jcesarmobile (changes requested):
just a minor thing on the share plugin changes

jcesarmobile (approved):

PR #224, opened 2021-01-29
simplify code

jcesarmobile:
I noticed you were getting the notifications as strings, converting to LocalNotification and then the LocalNotification to JSObject, but you can get the notifications as JSObjects directly from NotificationStorage, avoiding the double conversion, that simplifies the code, does less work and avoids the data loss I commented on the original PR


theproducer:
These changes were made manually to the PR branch.

PR #233, opened 2021-02-02
fix(filesystem): copy doesn't resolve on Android

### LINKED ISSUE #232: Filesystem.copy doesn't return Promise.resolve when it success.
### Plugin(s)

`@capacitor/filesystem` 

### Capacitor Version

```
Installed Dependencies:

  @capacitor/cli: 3.0.0-beta.1
  @capacitor/core: 3.0.0-beta.1
  @capacitor/android: 3.0.0-beta.1
  @capacitor/ios: 3.0.0-beta.1

```

### Platform(s)
Android

### Current Behavior

https://github.com/ionic-team/capacitor-plugins/blob/main/filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/FilesystemPlugin.java#L325
According to this `_copy` method implementation, it doesn't call `call.resolve()`. Now it only call `call.reject` when it failed. 

So, in TS code, when i use 
```
 await Filesystem.copy(...); 
```
It doesn't pass to next statement. 
image

### Expected Behavior

```
 await Filesystem.copy(...); 
```

It should be pass when copy is finished. 

### Code Reproduction
```
 await Filesystem.copy(...); 
```
with success arguments.

leo6104:
closes #232

### DIFF
--- a/filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/FilesystemPlugin.java
+++ b/filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/FilesystemPlugin.java
@@ -340,6 +340,7 @@ public class FilesystemPlugin extends Plugin {
         }
         try {
             implementation.copy(from, directory, to, toDirectory, doRename);
+            call.resolve();
         } catch (CopyFailedException ex) {
             call.reject(ex.getMessage());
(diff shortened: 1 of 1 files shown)


jcesarmobile (approved):
good catch, thanks!

PR #239, opened 2021-02-04
fix(filesystem): rmdir doesn't resolve on iOS

### LINKED ISSUE #238: Filesystem.rmdir doesn't call `resolve` function when it success.
### Plugin(s)
FileSystem

### Capacitor Version

```
  @capacitor/cli: 3.0.0-beta.1
  @capacitor/core: 3.0.0-beta.1
  @capacitor/android: 3.0.0-beta.1
  @capacitor/ios: 3.0.0-beta.1
```

### Platform(s)
iOS

### Current Behavior
https://github.com/ionic-team/capacitor-plugins/blob/main/filesystem/ios/Plugin/FilesystemPlugin.swift#L167

It doesn't call `resolve()` when it success. 

image

Never end native call. 

### Expected Behavior
Should be called.

leo6104:
closes #238

### DIFF
--- a/filesystem/ios/Plugin/FilesystemPlugin.swift
+++ b/filesystem/ios/Plugin/FilesystemPlugin.swift
@@ -165,6 +165,7 @@ public class FilesystemPlugin: CAPPlugin {
 
         do {
             try implementation.rmdir(at: fileUrl, recursive: recursive)
+            call.resolve()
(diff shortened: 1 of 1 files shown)


jcesarmobile (approved):
thanks, I was looking into another issue and noticed this, was going to fix it but saw your PR, good catch!

PR #243, opened 2021-02-08
refactor(android): permission API cleanup changes

carlpoole:
Changes related to permission API cleanup https://github.com/ionic-team/capacitor/pull/4171

### DIFF
--- a/network/android/src/main/java/com/capacitorjs/plugins/network/NetworkPlugin.java
+++ b/network/android/src/main/java/com/capacitorjs/plugins/network/NetworkPlugin.java
@@ -42,7 +42,7 @@ public class NetworkPlugin extends Plugin {
     @SuppressWarnings("MissingPermission")
     @PluginMethod
     public void getStatus(PluginCall call) {
-        if (hasRequiredPermissions()) {
+        if (isPermissionDeclared("Network")) {
             call.resolve(getStatusJSObject(implementation.getNetworkStatus()));
         } else {
             call.reject(PERMISSION_NOT_SET);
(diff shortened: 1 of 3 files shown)


network/android/src/main/java/com/capacitorjs/plugins/network/NetworkPlugin.java
```diff
@@ -42,7 +42,7 @@ protected void handleOnDestroy() {
     @SuppressWarnings("MissingPermission")
     @PluginMethod
     public void getStatus(PluginCall call) {
-        if (hasRequiredPermissions()) {
+        if (isPermissionDeclared("Network")) {
```
jcesarmobile, line 45: there is no `Network` alias, that's the plugin name, I get a 
> isPermissionDeclared: No alias defined for Network or missing @CapacitorPlugin annotation.

jcesarmobile (changes requested):
Fixed the conflicts so I could properly test.
Added a comment on Network plugin changes, the other two looks good.

network/android/src/main/java/com/capacitorjs/plugins/network/NetworkPlugin.java
carlpoole: Thanks. I dont think this plugin needs a permission annotation - the permission isn't dangerous and doesn't require permission checks from the user. We put this permissions in the manifest of the plugin now so it should always be there.

jcesarmobile: yeah, that was what I commented on the first implementation PR, but @ikeith said it did no harm to keep the check

https://github.com/ionic-team/capacitor-plugins/pull/8/files#r477256795

dwieeb (approved):

jcesarmobile (approved):

Issue #240, opened 2021-02-05
filesystem: default Directory is different on each platform


jcesarmobile:
### Plugin(s)
filesystem

### Capacitor Version

```
PASTE OUTPUT HERE
```

### Platform(s)
Android
iOS
web

### Current Behavior
on iOS if no Directory is provided, Documents is used as default
on web if no Directory is provided, Data is used as default
on Android if no Directory is provided, no directory is used as default, instead it assumes it's an absolute url

### Expected Behavior
Al platforms should behave the same

### Other Technical Details
Android allows absolute urls in the filesystem, while iOS and web will always use a default root if not specified, so I don't think we can just put a default value for Android's Directory as that would mean the access through absolute urls will break.
So maybe we should allow using absolute urls on iOS and web?

PR #250, opened 2021-02-12
feat(filesystem): Allow the use of absolute urls on iOS and web

jcesarmobile:
Android allowed using file urls when no directory was passed.
This also allows absolute file urls on iOS and web platforms when no directory is passed instead of using a default directory

closes https://github.com/ionic-team/capacitor-plugins/issues/240

### DIFF
--- a/filesystem/ios/Plugin/Filesystem.swift
+++ b/filesystem/ios/Plugin/Filesystem.swift
@@ -136,14 +136,14 @@ import Foundation
     /**
      * Get the SearchPathDirectory corresponding to the JS string
      */
-    @objc public func getDirectory(directory: String) -> FileManager.SearchPathDirectory {
+    public func getDirectory(directory: String?) -> FileManager.SearchPathDirectory? {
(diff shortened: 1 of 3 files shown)


filesystem/ios/Plugin/Filesystem.swift
```diff
@@ -136,32 +136,30 @@ import Foundation
...
      * Get the SearchPathDirectory corresponding to the JS string
      */
-    @objc public func getDirectory(directory: String) -> FileManager.SearchPathDirectory {
+    public func getDirectory(directory: String?) -> FileManager.SearchPathDirectory? {
         switch directory {
         case "DOCUMENTS":
             return .documentDirectory
         case "CACHE":
             return .cachesDirectory
         default:
```
jcesarmobile, line 145: Need to also return documents directory for all other options, should only return nil when no directory is passed

ikeith (approved):

dwieeb (approved):

PR #253, opened 2021-02-15
feat(local-notifications): Support setting seconds in Schedule "on"

theproducer:
This PR adds the ability to define seconds on the `on` property of `LocalNotificationSchedule`.

fixes: https://github.com/ionic-team/capacitor/issues/1679


carlpoole (approved):

jcesarmobile (approved):
Looks good, but I think this should be a feature, not a fix, despite it fixes an issue too, the main change is to allow passing seconds (and involves changing the types)

Issue #221, opened 2021-01-28
local-notifications getPending shouldn't return notifications that already fired


jcesarmobile:
### Plugin(s)
local-notifications

### Capacitor Version
3

### Platform(s)
android

### Current Behavior
getPending returns notifications that were already fired and are still on the notification drawer

### Expected Behavior
getPending shouldn't return notifications that were already fired

PR #256, opened 2021-02-16
fix(local-notifications): Make getPending not return already fired notifications

theproducer:
This PR adds changes to the local-notification plugin to filter out already fired notifications from the getPending notifications call.

Since Capacitor supports Android API 21+, we can't rely on `NotificationManager.getActiveNotifications()` to easily get a list of notifications in the status bar.  Instead, this PR sets a `visible` flag on each LocalNotification that gets set when a notification is fired, then cleared when a notification is interacted with.

fixes: https://github.com/ionic-team/capacitor-plugins/issues/221

### DIFF
--- a/action-sheet/package.json
+++ b/action-sheet/package.json
@@ -44,10 +44,10 @@
     "prepublishOnly": "npm run build"
   },
   "devDependencies": {
-    "@capacitor/android": "^3.0.0-beta.2",
-    "@capacitor/core": "^3.0.0-beta.2",
+    "@capacitor/android": "^3.0.0-beta.3",
+    "@capacitor/core": "^3.0.0-beta.3",
     "@capacitor/docgen": "0.0.15",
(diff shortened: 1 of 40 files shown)


jcesarmobile (commented):
I've tested and it works, but it adds a ton of complexity and I think using `storage.deleteNotification` when the notification fires is enough to not show it on pending and I don't see any side effects on doing that.
The only thing is have in mind that should only be deleted if not repeatable, so probably should be done inside of the `rescheduleNotificationIfNeeded` method or make it return a boolean to know if it was rescheduled or not and delete it only if not rescheduled

theproducer:
Thanks - just deleting the notification worked just as well without any issues!

jcesarmobile (approved):
looks good

let's wait for next beta before merging so tests pass again (hopefully)

Issue #257, opened 2021-02-17
Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference


jcesarmobile:
### Plugin(s)
local-notifications

### Capacitor Version

```
PASTE OUTPUT HERE
```

### Platform(s)
android

### Current Behavior
app crash with NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference

### Expected Behavior
to not crash

### Code Reproduction
in the sample app tap the "SCHEDULE NOW" and app will crash with the mentioned message

PR #258, opened 2021-02-17
fix(local-notifications): Checking for null schedule in notification JSObject

theproducer:
This PR fixes a crash caused by attempting create a LocalNotificationSchedule object when its source `schedule` JSObject was null.

fixes: https://github.com/ionic-team/capacitor-plugins/issues/257

### DIFF
--- a/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationSchedule.java
+++ b/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationSchedule.java
@@ -20,8 +20,7 @@ public class LocalNotificationSchedule {
 
     private Boolean whileIdle;
 
-    public LocalNotificationSchedule(JSObject jsonNotification) throws ParseException {
-        JSObject schedule = jsonNotification.getJSObject("schedule");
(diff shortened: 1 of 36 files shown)


local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationSchedule.java
```diff
@@ -20,8 +20,7 @@
 
     private Boolean whileIdle;
 
-    public LocalNotificationSchedule(JSObject jsonNotification) throws ParseException {
-        JSObject schedule = jsonNotification.getJSObject("schedule");
+    public LocalNotificationSchedule(JSObject schedule) throws ParseException {
         if (schedule != null) {
```
jcesarmobile, line 24: Since you are checking if `schedule` is `null` before creating the `LocalNotificationSchedule` in `LocalNotification.java`, you can remove the `null` check from here. Other than that, changes look good.

theproducer: Yeah - I thought I took care of that, thanks for pointing it out!

jcesarmobile (approved):

Issue #130, opened 2019-09-07
Feature Request: Please add support for App.minimizeApp()


naveedahmed1:
Description of the problem:

Currently we are able to use hardware back button to exit app by listening to `backButton `event and then calling `App.exitApp();`:

```
App.addListener('backButton', (data: AppUrlOpen) => {
            if (!(this.routerOutlet && this.routerOutlet.canGoBack())) {
                App.exitApp();
            }
        });
```

It would be great, that if an option to Minimize App could also be added to the capacitor and we could use something like `App.minimizeApp(); `

There's a cordova plugin https://github.com/tomloprod/cordova-plugin-appminimize for this purpose, but I believe since its a common use case, it should be added to the Capacitor.

Affected platform
- [x ] Android
- [ x] iOS

OS of the development machine
- [x ] Windows

phal0r:
I wanted to solve this problem for ionic react, but unfortunately there is no `canGoBack()` function, which could indicate, that we are on the first entry of the history stack. So even if capacitor would have a minimize function, it would not work well for ionic react.

At the same time I had a look at the cordova minimize plugin and the capacitor app plugin and I want to suggest the following:

Capacitor app plugin has built-in support for navigating back with the hardware back button on android. See: https://github.com/ionic-team/capacitor-plugins/blob/main/app/android/src/main/java/com/capacitorjs/plugins/app/AppPlugin.java#L45

The app plugin could be enhanced with setters to enable minimizing or closing the app, if `bridge.getWebView().canGoBack()` returns `false`. This would be quite easy to achieve and would be transparent for any js framework and flavor, that is being used for development. Straightforward documentation and less traffic on the bridge as it is handled native completely. Thoughts on that?

kawazoe:
@phal0r The behavior of the back button does not always depend on the webview's ability to go back, but might depend on the ability of a specific router-outlet to go back. For instance, with multiple tabs, while the web view might be able to go back "across tabs", this is not necessarily desired. This is how I usually implement it in Angular. I am not familiar with routing in ionic react, but you probably have an equivalent feature, or the ability to save this information in your store.

nothingkid:
Added "minimizeApp" method to "app" plugin
https://github.com/ionic-team/capacitor-plugins/pull/743

PR #265, opened 2021-02-22
feat(app): Add canGoBack to backButton event.

lishid:
This PR adds the `canGoBack` (boolean) status to the event object of the `backButton` event on Android.

This status isn't normally accessible from the JS side, but would be very useful if the app uses the browser history stack for handling navigation.


lishid:
Related: https://github.com/ionic-team/capacitor-plugins/issues/130#issuecomment-778804244

lishid:
Weird, lint seems to have passed on my previous commit just 10 minutes before, and my new commit only adds the auto-generated docs. It also seems to say '@capacitor/text-zoom' so I'm guessing it's just flaky? Is there any way to re-run the lint?

lishid:
Got lint to re-run and clear the PR!

lishid:
@imhoffd Any feedback? I know you guys are probably busy. Hopefully this small patch won't take long.

lishid:
@jcesarmobile @theproducer

jcesarmobile (approved):
thanks for the PR
I changed the version to 1.0.0 as it refers to the plugin version, not the capacitor version
it's still not 1.0.0, but will be once we release the final plugins

PR #277, opened 2021-03-01
fix(camera): decode content uri when retrieving image from gallery

eidng8:
The problem seems to be the encoded URI. The aforementioned line `String filename = contentUri.getLastPathSegment()` is really the real point of failure.

On a redmi phone (maybe some others too?), the `contentUri` points to something like `content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FDCIM%2FCamera%2FIMG_20210301_082750.jpg`. So you see the encoded part. `getLastPathSegment()` extract the whole `%2Fstorage%2Femulated%2F0%2FDCIM%2FCamera%2FIMG_20210301_082750.jpg` part, then decodes and returns it. Eventually, the following `outFile` got a path like `...package..name.../cache/storage/emulated/0/DCIM/Camera/IMG_20210301_082750.jpg`, and you see where it crashes.

I'm not sure if I'm doing it the tidy way, it's the most straightforward way I can think of.


jcesarmobile (approved):
looks good, thanks!

PR #283, opened 2021-03-02
docs(geolocation): use proper function syntax on example

jeremytenjo:
added missing => to printCurrentPosition  function


jcesarmobile (approved):
good catch, thanks!

Issue #254, opened 2019-12-10
bug: Can't find variable: Notification on iOS 


jcesarmobile:
When using LocalNotifications plugin on iOS (web), you'll get that error when trying to request permissions. Should be handled to return a proper error.

PR #285, opened 2021-03-02
fix(local-notification): Throw unavailable if Notification API not supported

theproducer:
During `checkPermissions`, check to see if the Notifications API is supported by the browser and throws `unavailable` if not.

Closes: https://github.com/ionic-team/capacitor-plugins/issues/254


jcesarmobile (changes requested):
Looks good, but should be done in `requestPermissions` and on `schedule` too (maybe others too, but those two for sure)

theproducer:
Yeah, I think those two methods are it - most use of LocalNotification has to go through those three methods to do anything.

jcesarmobile (approved):

Issue #284, opened 2021-03-02
bug: Failed to construct 'Notification': Illegal constructor


theproducer:
Using LocalNotification on Android Chrome fails when attempting to build a notification:

```
Unhandled Rejection (TypeError): Failed to construct 'Notification': Illegal constructor.  Use ServiceWorkerRegistration.showNotification() instead.
```

jcesarmobile:
related
https://bugs.chromium.org/p/chromium/issues/detail?id=481856

theproducer:
Yes - I don't think we are going support using Service Workers in right now (I could be wrong?) so the plan is just to check for `new Notification()` support and return unavailable if not.

PR #295, opened 2021-03-04
fix(local-notifications): Adding check for `new Notification` support

theproducer:
This PR adds a check for `new Notification` support in browsers.

https://developers.google.com/web/updates/2015/05/notifying-you-of-changes-to-notifications?hl=en#android_notifications

closes: https://github.com/ionic-team/capacitor-plugins/issues/284


local-notifications/src/web.ts
```diff
@@ -94,6 +94,26 @@ export class LocalNotificationsWeb
     return { display };
   }
 
+  protected checkNotificationSupport = (): boolean => {
+    if (!window.Notification || !Notification.requestPermission) {
```
jcesarmobile, line 98:
```suggestion
    if (!('Notification' in window) || !Notification.requestPermission) {
```
Why change this? I don't prefer one over the other, we should be consistent with this window checks, most plugins use it like this

theproducer: Just for consistency with the requestPermission check.  It doesn't really make a difference - I can change it back.

local-notifications/src/web.ts
```diff
@@ -94,6 +94,26 @@ export class LocalNotificationsWeb
     return { display };
   }
 
+  protected checkNotificationSupport = (): boolean => {
```
jcesarmobile, line 97: I think it would be more meaningful if renamed to `hasNotificationSupport`

jcesarmobile (approved):

PR #311, opened 2021-03-12
docs(haptis): Update README example to use ImpactStyle

skotturi:
Capacitor v3 has deprecated `HapticsImpactStyle` in favor of `ImpactStyle` but the docs were still using the old syntax. VSCode suggested the change but it would be nice of the docs mentioned this change especially as developers migrate from v2 to v3.


jcesarmobile (approved):
Good catch, thanks!

Issue #313, opened 2021-03-15
capacitor-filesystem: stat return inconsistent timestamps between platforms


lishid:
## Bug Report
`Filesystem.stat` returns milliseconds for Android and seconds for iOS.

### Plugin(s)
capacitor-filesystem

### Capacitor Version
3.0.0-rc.0

### Platform(s)
iOS

### Current Behavior
ctime and mtime are turned in seconds.

### Expected Behavior
ctime and mtime are returned in milliseconds.

### Code Reproduction
Execute `Filesystem.stat({...})` for any file.

### Additional Context
Probably just need a `* 1000` for these two:
https://github.com/ionic-team/capacitor-plugins/blob/98e91cd745fb12bf46f99233bb527f147dbba58b/filesystem/ios/Plugin/FilesystemPlugin.swift#L219-L220

Might be worth adding this to the docs too:
https://github.com/ionic-team/capacitor-plugins/blob/98e91cd745fb12bf46f99233bb527f147dbba58b/filesystem/src/definitions.ts#L390

PR #321, opened 2021-03-18
fix(filesystem): Convert stat ctime/mtime timestamp to milliseconds

theproducer:
This PR makes `Filesystem.stat` return `ctime` and `mtime` in milliseconds iOS, bringing it in line with Android.

closes: https://github.com/ionic-team/capacitor-plugins/issues/313

### DIFF
--- a/filesystem/README.md
+++ b/filesystem/README.md
@@ -420,13 +420,13 @@ Required on Android, only when using <a href="#directory">`Directory.Documents`<
 
 #### StatResult
 
-| Prop        | Type                | Description               | Since |
-| ----------- | ------------------- | ------------------------- | ----- |
-| **`type`**  | <code>string</code> | Type of the file          | 1.0.0 |
-| **`size`**  | <code>number</code> | Size of the file          | 1.0.0 |
-| **`ctime`** | <code>number</code> | Time of creation          | 1.0.0 |
-| **`mtime`** | <code>number</code> | Time of last modification | 1.0.0 |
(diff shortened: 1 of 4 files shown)


jcesarmobile (changes requested):
iOS is returning decimals, while Android and web aren't, so I think iOS shouldn't return the decimals neither.

The types should be updated to specify that it returns the value in milliseconds.

Android returns null for `ctime`, seems like it's possible to get the value on Android 26+, so we should probably return the value when available.

Should be documented on the upgrade guide since it can be considered a breaking change for iOS.

jcesarmobile (changes requested):
Looks like the creation time code is not working, it returns the last modified value too.
So not sure if it's best to remove it again.

I've noticed that `lastAccessTime` seems to be the creation date, so maybe we can return the min between that and `creationTime` (just in case it's an android bug and they fix it in the future). That or remove the whole ctime return.
If we keep the ctime return, instead of 26 it's better to use `Build.VERSION_CODES.O`, and as there is already a `data.put("ctime", null);`, put that in the else of the android version check.

Also the README need to be re generated since the types changed.

theproducer:
Thanks - I'll open a PR in the docs for updates to the upgrade guide.

jcesarmobile:
Not sure if you saw my last comment since I requested changes and you requested review again without making any changes

theproducer:
Sorry, just saw them!

jcesarmobile (approved):
looks good

PR #324, opened 2021-03-23
fix(share): set mime type to */* if no mine type can be found

pablo-mendoza:
Allow arbitrary files to be shared in Android by setting the mime type to &ast;/&ast; if no mine type can be found.


jcesarmobile (approved):
looks good, thanks!

PR #335, opened 2021-04-02
fix(keyboard): Use new Config name for scroll

jcesarmobile:


ikeith:
Dependent on https://github.com/ionic-team/capacitor/pull/4441

ikeith (approved):

Issue #322, opened 2021-03-19
GetPreferred hangs in TextZoom plugin on IOS


jezrsmith:
### Plugin(s)
@capacitor/TextZoom

### Capacitor Version

```
Latest Dependencies:

  @capacitor/cli: 2.4.7
  @capacitor/core: 2.4.7
  @capacitor/android: 2.4.7
  @capacitor/ios: 2.4.7

Installed Dependencies:

  @capacitor/cli: 3.0.0-rc.0
  @capacitor/core: 3.0.0-rc.0
  @capacitor/android: 3.0.0-rc.0
  @capacitor/ios: 3.0.0-rc.0
```

### Platform(s)
IOS

### Current Behavior
calling TextZoom.getPreferred() hangs on IOS. Works OK on Android (simulator)

### Expected Behavior
calling TextZoom.getPreferred() should return preferred zoom level.

### Code Reproduction
Can be reproduced using the capacitor-testapp
Just run the app, go to the TextZoom screen and click Get Preferred Zoom. Hangs on both IOS simulator and real device (IOS 14.4.1

### Other Technical Details
IOS 14.4.1

alexsasharegan:
I've been attempting to debug this for a couple hours today. Something looks amiss with how there is a javascript implementation of the ios plugin alongside the ios sub package and swift implementation. I'm referring to this: https://github.com/ionic-team/capacitor-plugins/blob/main/text-zoom/src/index.ts#L6

alexsasharegan:
See my PR (#338) for progress. I think I've found most of the problems that would take this plugin to a functional state on iOS, but I've taken it about as far as I know how.

PR #338, opened 2021-04-06
Text-Zoom base fixes

alexsasharegan:
This appears to correct the major problems encountered in #322 , but someone iOS savvy will have to complete the implementations of `get` and `set`. This at least stubs the methods to complete the interface of the plugin.

**Fixes**

1. Renames ios typescript impl to `web.ts`
2. Updates web plugin registration to function loader
3. Updates web plugin to extend base class
4. Updates web plugin to avoid self-reference causing infinite loop
5. Stubs missing iOS plugin methods

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ See [`CONTRIBUTING.md`](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRI
     brew install swiftlint
     ```
 
-Sometimes, it may be necessary to work on Capacitor in parellel with the plugin(s). In this case, a few extra steps are necessary:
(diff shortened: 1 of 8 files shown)


alexsasharegan:
I'm not sure how to generate the objective C header files (or whatever the `.h` and `.m` files are). But those appear to need updating for the `get` and `set` methods.

jcesarmobile:
Thanks for the PR, but the idea of this plugin was to use web code for get/set and this PR breaks that.

At the moment there is a bug in plugins that mix native code with web code and we are working on a fix in capacitor/core, that would fix the issue you linked and a few more

alexsasharegan:
I can change the PR to meet your design goals too. But frankly, I just want the plugin to work. So be it. At least I know I can implement this for an unstable patch to get a production feature out.

Issue #337, opened 2021-04-06
bug: extra data passed to a scheduled LocalNotification missing when received via localNotificationActionPerformed (ios)


moxaj:
### Capacitor Version

Latest Dependencies:

  @capacitor/cli: 2.4.7
  @capacitor/core: 2.4.7
  @capacitor/android: 2.4.7
  @capacitor/ios: 2.4.7

Installed Dependencies:

  @capacitor/cli: 3.0.0-beta.6
  @capacitor/android: 3.0.0-beta.6
  @capacitor/ios: 3.0.0-rc.0
  @capacitor/core: 3.0.0-beta.6

### Platform(s)

ios

### Current Behavior

Here's how I schedule a local notification:

```
await LocalNotifications.schedule({
  notifications: [
    {
      id: 0,
      title: 'Title',
      body: 'Body',
      extra: {a: 'a', b: 'b'}
    }
  ]
});
```

Here's how I handle it once tapped:

```
LocalNotifications.addListener('localNotificationActionPerformed', async event => {
  const extra = event.notification.extra;
  // ...
});
```

Currently, `extra` is an empty object. The title and body are displayed as expected.  Works on android, bugged on ios.

### Expected Behavior

I'd expect `extra` to be `{a: 'a', b: 'b'}`.

### Code Reproduction

I did some (very shallow) debugging, and I believe this line to be the issue:

https://github.com/ionic-team/capacitor-plugins/blob/2a83fcb536cdfc5b601f363212353201de40ca5b/local-notifications/ios/Plugin/LocalNotificationsHandler.swift#L82

If I replace it with `"extra": notificationRequest["extra"] as? JSObject ?? [:],`, everything works as expected.

### Other Technical Details

`npm --version` output: 6.14.6

`node --version` output: v12.18.3

`pod --version` output (iOS issues only): 1.10.1

vitorlopez:
Is there an idea about when this fix is going to be pushed to NPM?

PR #340, opened 2021-04-07
fix(local-notifications): extra not being returned on notification events

jcesarmobile:
extra was not being returned because it needed to be coerced, but also because userInfo was split into `cap_schedule` and `cap_extra`.

So reused `makePendingNotificationRequestJSObject` as it was already handling `cap_schedule` and `cap_extra` to generate the notification of `makeNotificationRequestJSObject`

closes https://github.com/ionic-team/capacitor-plugins/issues/337


vitorlopez:
Thanks for that PR!

theproducer (approved):
Looks good and works.

PR #347, opened 2021-04-15
doc(storage): add async to setName example function

RageBrahimi:
Add `async` to `setName` example.


jcesarmobile (approved):
good catch, thanks!

PR #349, opened 2021-04-16
fix(push-notifications): proper return of push notification object properties

theproducer:
- Renames `extra` back to `data`
- Remove `actionTypeId` and `attachments`
- Adds back `subtitle` and `badge`

closes: https://github.com/ionic-team/capacitor-plugins/issues/332


jcesarmobile (commented):
sound wasn't on Capacitor 2.x neither, any reason to leave it here?

theproducer:
> sound wasn't on Capacitor 2.x neither, any reason to leave it here?

Nope - at least as far as I am aware.

It looks like right at the start of the port to Cap 3, properties from local and push notifications were melded together mistakenly or something.

jcesarmobile (changes requested):
Before the return there is a
`let notificationRequest = notificationRequestLookup[request.identifier] ?? [:]` line that is no longer used, can you delete it?

jcesarmobile (approved):

Issue #367, opened 2021-04-29
Device.getInfo pollutes localStorage without warning/consent


boennemann:
### Plugin(s)
device

### Platform(s)
web

### Current Behavior
When calling Device.getInfo() on the web this plugins writes a unique id to the localStorage.

### Expected Behavior
Storing information on the users device requires consent under the [EPrivacy Regulation](https://en.wikipedia.org/wiki/EPrivacy_Regulation)

It is totally unexpected that an innocent sounding API like **get**Info is actually creating the information in question, writing to the disk and thus very likely messing with the compliance of the consuming application.

This functionality should either be warned about, or even better not be the default.

### Code Reproduction

```js
await Device.getInfo()
localStorage.getItem('_capuid')
```

PR #370, opened 2021-05-03
feat(device): Add getId function

jcesarmobile:
create a new getId function that returns the uuid and remove it from getInfo
document that the uuid gets stored on localStorage on web

closes https://github.com/ionic-team/capacitor-plugins/issues/367

### DIFF
--- a/device/ios/Plugin/DevicePlugin.swift
+++ b/device/ios/Plugin/DevicePlugin.swift
@@ -6,9 +6,13 @@ public class DevicePlugin: CAPPlugin {
     private let implementation = Device()
 
     @objc func getId(_ call: CAPPluginCall) {
-        call.resolve([
-            "uuid": UIDevice.current.identifierForVendor!.uuidString
-        ])
+        if let uuid = UIDevice.current.identifierForVendor {
+            call.resolve([
+                "uuid": uuid.uuidString
+            ])
+        } else {
+            call.reject("Id not available")
+        }
     }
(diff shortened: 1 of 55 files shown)


thomasvidas (approved):

device/ios/Plugin/DevicePlugin.swift
```diff
@@ -5,6 +5,11 @@ import Capacitor
 public class DevicePlugin: CAPPlugin {
     private let implementation = Device()
 
+    @objc func getId(_ call: CAPPluginCall) {
+        call.resolve([
+            "uuid": UIDevice.current.identifierForVendor!.uuidString
```
ikeith, line 10: Although the previous implementation forced unwrapped the optional, we should avoid doing so where possible. [The documentation](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor) is unclear on when the method might return nil; it only provides one example scenario where it can happen. But adding a fallback is easy insurance against a crash.

```suggestion
            "uuid": UIDevice.current.identifierForVendor?.uuidString ?? ""
```

carlpoole (approved):

device/ios/Plugin/DevicePlugin.swift
jcesarmobile: I try to not change code when moving it from one place to another.
Anyway, I've made it to reject if the identifier is not available, I think that's better than returning an empty string.

ikeith (approved):

PR #379, opened 2021-05-07
bugfix(camera): use a distinguishable permission denied string for camera and photos

crixx:
### Problem
When calling `Camera.getPhoto` the user is presented with an actionSheet allowing to either choose the camera or use an existing file from photos. Currently, when photos is choosen, and the user does not grant permissions, the returned error says "Unable to access camera, user denied permission request". 

### Solution
In order to be able to distinguish, which option (camera/photos) was chosen, this PR adds an explicit error string for the case where photos permissions are not granted.


jcesarmobile (approved):
thanks for the PR, I'll sync the ios/android wording in a separate PR

PR #400, opened 2021-05-18
fix(camera): Remove deprecated method saveCall(call)

ghonche-yqr:
Replace deprecated method `saveCall(call)` with suggested method `bridge.saveCall(call)`


jcesarmobile:
Thanks for the PR.
The call is actually not retrieved but passed to the functions that need it, so it should not be replaced with the non deprecated save, but be removed, I sent a PR removing it
https://github.com/ionic-team/capacitor-plugins/pull/401

PR #429, opened 2021-05-25
Error. Forgot the denial `not`.

ivansglazunov:


jcesarmobile (changes requested):
The readme is generated from the types, can you change ti there too?

https://github.com/ionic-team/capacitor-plugins/blob/main/push-notifications/src/definitions.ts#L36

Issue #24, opened 2020-08-28
remove `preserveAspectRatio` option


imhoffd:
`preserveAspectRatio` is deprecated and not really used, remove before @capacitor/camera 2.0.0 is released

sandstrom:
Thanks for creating this issue @dwieeb 

I'd even consider removing it, because I'd say there is ~0% use for resize without preserved aspect ratio. Less code to maintain.

dwieeb:
I think you're right. @jcesarmobile What do you think?

jcesarmobile:
It has been removed on 4.0.0 beta 0

PR #448, opened 2021-05-30
fix(camera): fix `preserveAspectRatio` option on iOS

imhoffd:
Hi. 🤓

It looks like the resize functionality from Capacitor 2 never made it into the rewrite for the iOS Camera plugin. The [`width` and `height` options](https://capacitorjs.com/docs/apis/camera#imageoptions) are always used literally. But the `preserveAspectRatio` option is supposed to make `width` and `height` act as boundaries and maintain the aspect ratio when resizing.

For functionality reference, look at the behavior on Android.
See the original PR: https://github.com/ionic-team/capacitor/pull/3309

### DIFF
--- a/camera/ios/Plugin/CameraPlugin.swift
+++ b/camera/ios/Plugin/CameraPlugin.swift
@@ -117,6 +117,7 @@ public class CameraPlugin: CAPPlugin {
             settings.shouldResize = true
         }
         settings.shouldCorrectOrientation = call.getBool("correctOrientation") ?? true
+        settings.preserveAspectRatio = call.getBool("preserveAspectRatio") ?? false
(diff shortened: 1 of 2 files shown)


jcesarmobile:
looks like the idea was to remove preserveAspectRatio from the plugin https://github.com/ionic-team/capacitor-plugins/issues/24

we will have to wait for the 2.0.0 release

jcesarmobile:
I've just tested on both iOS and Android and both of them are preserving aspect ratio, no matter what the option is, there is no code on any of them to not preserve the aspect ration

jcesarmobile:
yeah, I confirm the code is already there, there is an UIImage extension that handles the resizing

https://github.com/ionic-team/capacitor-plugins/blob/main/camera/ios/Plugin/CameraExtensions.swift#L68

PR #460, opened 2021-06-04
fix(camera): fix photo resizing on iOS

imhoffd:
Hi again 🤓 

My [earlier PR](https://github.com/ionic-team/capacitor-plugins/pull/448) was made because resizing images on iOS by specifying `width` and `height` doesn't work, so I thought the logic was missing, but it turns out it just doesn't work.

To test the bug, just add `width` and `height` options to the TestApp Camera page. Notice it does not preserve the aspect ratio. This PR fixes that.


jcesarmobile (approved):
looks good, I couldn't reproduce at first because it's only a problem when width is bigger than height

PR #467, opened 2021-06-10
docs(local-notifications): export in module style

reslear:


jcesarmobile (approved):
good catch, thanks!

jcesarmobile:
wops, that's actually autogenerated, so I have to change the generator to use the other style and do a new build of the docs, sorry, won't be able to merge your PR, but will get it fixed

PR #468, opened 2021-06-10
fix(camera): Hide the temporary file input for web

boboldehampsink:
Otherwise it shows up while the file prompt is open


jcesarmobile:
thanks, there used to be problems in the past when trying to forward the click to the input file if it was hidden with display none or visibility hidden, but looks like it works fine now on all major browsers.

I've gone with a similar alternative that uses the hidden attribute instead of the hidden style.

https://github.com/ionic-team/capacitor-plugins/pull/484

PR #483, opened 2021-06-17
fix(browser): browserFinished does not fire when closing popovers

inbreaks:


jcesarmobile (approved):
thanks for the PR, I made a few small changes and one addition to also fire the event when closing the popover on iPads on iOS 12

PR #489, opened 2021-06-18
fix(share): Prevent share if sharing in progress

jcesarmobile:
If you call share twice in a row it causes issues, on android two sheets appear but both return the result of the first one, on iOS the second one is not presented and the promise never resolves nor rejects.

This PR detects if a share is in progress and prevent it rejecting with an error-

closes https://github.com/ionic-team/capacitor-plugins/issues/323


carlpoole (approved):

elylucas (approved):

qliqdev:
While inapp browser is opened share is not working 

throwing

`Can't share while sharing is in progress`

https://github.com/ionic-team/capacitor-plugins/blob/2fe419b4af67f5ee0350880c62ebeda93badc3d8/share/ios/Sources/SharePlugin/SharePlugin.swift#L67-L70
is this expression right?

PR #505, opened 2021-06-28
feat: Added Device Machine Model name for iOS devices

oneclickwebstudio:
### iOS
This feature needs many developers on Capacitor. As i tested, all devices on ios prints device machine name as well

For example:
```
{
  ...
    model: "iPhone",
    modelName: "iPhone13,1"
  ...
}
```

### Android
In android devices `android.os.Build.MODEL` prints as well. So, no need to edit.


jcesarmobile:
version 4.0.0 beta 0 of the plugin will return the proper model on the model field instead of adding an additional modelName.

PR #507, opened 2021-06-28
bugfix(keyboard) Add no-op Keyboard web plugin

MortenGregersen:
When using the Keyboard plugin, we get the following log in the console, for an app using Capacitor in a desktop web browser:

Screenshot 2021-06-29 at 14 19 09

`Error: Uncaught (in promise): Error: "Keyboard" plugin is not implemented on web`

We want to silence this with a no-op keyboard web plugin.


MortenGregersen:
@elylucas can we get this merged?

MortenGregersen:
@thomasvidas can we get this merged?

PR #513, opened 2021-07-10
feat(push-notifications): Send an Android broadcast with the remoteMessage inside the firebase onMessageReceived event

NePheus:
**Content of this PR**
This pull request is intended to send a broadcast message in android, after an incoming firebase push notification.

**Why**
Unfortunately it is not possible to register more than one FirebaseMessagingService in the same app (see [here](https://stackoverflow.com/questions/47659458/android-can-an-android-app-have-multiple-firebasemessagingservices)). With help of the BroadcastReceiver, another plugin can hook in here and also listen to the remote messages with a broadcast listener:
`registerReceiver(MyBroadcastReceiver, new IntentFilter("android.intent.action.capacitor.pushnotification"));`

**Example use case**
I want to send a (silent) data notification to my app to show a local notification with a fullscreen intent (i.e. call action). My new capacitor plugin is handling the implementation of that and must be notified by the push-notifications plugin about the data notification.

### DIFF
--- a/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/MessagingService.java
+++ b/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/MessagingService.java
@@ -1,5 +1,6 @@
 package com.capacitorjs.plugins.pushnotifications;
 
+import android.content.Intent;
 import androidx.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


NePheus:
This is my new plugin, that will only work with this adjustment: https://github.com/NePheus/capacitor-fullscreen-notification

PR #522, opened 2021-07-16
fix(camera): Don't try to correct exif on content urls

jcesarmobile:
We can't really write to content urls, so we need to figure out a way of correcting the exif orientation on content urls, for now just avoid the crash by not trying to reset it if it's a content url.

closes https://github.com/ionic-team/capacitor-plugins/issues/524


jcesarmobile:
closing in favor of https://github.com/ionic-team/capacitor-plugins/issues/524

PR #523, opened 2021-07-20
feat(keyboard): Add keyboardAnimationDuration to KeyboardInfo

viablecell:
1. Add keyboardAnimationDuration to KeyboardInfo
2. iOS, Android, notifyListeners  keyboardWillShow and keyboardDidShow with keyboardAnimationDuration

### DIFF
--- a/keyboard/android/src/main/java/com/capacitorjs/plugins/keyboard/KeyboardPlugin.java
+++ b/keyboard/android/src/main/java/com/capacitorjs/plugins/keyboard/KeyboardPlugin.java
@@ -75,9 +75,11 @@ public class KeyboardPlugin extends Plugin {
         switch (event) {
             case Keyboard.EVENT_KB_WILL_SHOW:
             case Keyboard.EVENT_KB_DID_SHOW:
-                String data = "{ 'keyboardHeight': " + size + " }";
+                // Convert Handler().postDelayed() 350ms to 0.35s
+                String data = "{ 'keyboardHeight': " + size + ", 'keyboardAnimationDuration': " + 0.35 + " }";
                 bridge.triggerWindowJSEvent(event, data);
                 kbData.put("keyboardHeight", size);
+                kbData.put("keyboardAnimationDuration", 0.35);
                 notifyListeners(event, kbData);
                 break;
             case Keyboard.EVENT_KB_WILL_HIDE:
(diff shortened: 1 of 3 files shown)


viablecell (commented):
Reviewed

keyboard/android/src/main/java/com/capacitorjs/plugins/keyboard/KeyboardPlugin.java
```diff
@@ -75,9 +75,11 @@ void onKeyboardEvent(String event, int size) {
         switch (event) {
             case Keyboard.EVENT_KB_WILL_SHOW:
             case Keyboard.EVENT_KB_DID_SHOW:
-                String data = "{ 'keyboardHeight': " + size + " }";
+                // Convert Handler().postDelayed() 350ms to 0.35s
+                String data = "{ 'keyboardHeight': " + size + ", 'keyboardAnimationDuration': " + 0.35 + " }";
```
dminkovsky, line 79: Thanks so much for contributing this. Is 350 a constant on Android?

viablecell:
```
@PluginMethod
    public void show(final PluginCall call) {
        execute(
            () ->
                new Handler()
                    .postDelayed(
                        () -> {
                            implementation.show();
                            call.resolve();
                        },
                        350
                    )
        );
    }
```
350 is method param `long delayMillis`

dminkovsky: I'm sorry, I meant: on Android, is 350 milliseconds always the time it takes for the keyboard animation to complete. I'm wondering why that number is hard-coded.

Thank you.

austin43: I may be speaking out of ignorance here, but is this value only useful on iOS?  Perhaps the android part of this should be removed, and the docs should specify this as an iOS only value for the plugin?

dtzxporter:
Is there any chance we can merge this soon? This is extremely useful for adapting the keyboard on iOS.

We should probably match react native and do 0 for android. (It's been implemented in react for years).

https://github.com/facebook/react-native/blob/8cdc9e7f04e2dd3026d15dcd5765952bfb2c5c08/Libraries/Components/Keyboard/Keyboard.d.ts#L50

eric-horodyski:

More context:

PR #528, opened 2021-07-25
fix(local-notifications): Fixed wrong rollup config

trancee:
I tried to use the latest version of the Local Notifications plugin by linking the `dist/plugin.js` and it turns out the configuration of this plugin is wrong. Instead of using `capacitorLocalNotifications` it was set to `capacitorHaptics` and instead of using the globals `capacitorExports` it was using `capacitorLocalNotifications`.

Console output:
```
capacitor-local-notifications.js:150 Uncaught ReferenceError: capacitorLocalNotifications is not defined
    at capacitor-local-notifications.js:150
(anonymous) @ capacitor-local-notifications.js:150
```

### DIFF
--- a/local-notifications/rollup.config.js
+++ b/local-notifications/rollup.config.js
@@ -4,9 +4,9 @@ export default {
     {
       file: 'dist/plugin.js',
       format: 'iife',
-      name: 'capacitorHaptics',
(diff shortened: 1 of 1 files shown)


jcesarmobile (approved):
good catch, thanks!

PR #532, opened 2021-07-27
feat(storage): make StorageConfiguration init public

fadoscha:
In order to use StoragePlugin in the native context the StorageConfiguration inits needs to be public


jcesarmobile (approved):
good catch, thanks!

PR #546, opened 2021-08-15
fix(filesystem): allow copy if from is not parent of to

tachibana-shin:
Preview error:

image


Shin-Tachibana:
Preview error:

image

jcesarmobile (changes requested):
can you do it in a way that don't require adding external dependencies?

Shin-Tachibana:
I removed the dependencies. Please reconsider this request

jcesarmobile (approved):
looks good, thanks!

PR #594, opened 2021-09-02
bugfix(CameraPlugin): add missing call rejection in cameraPermissions…

### LINKED ISSUE #593: bug(CameraPlugin, Android) - endless loop when photo permission is denied with saveToGallery option
### Plugin(s)
  @capacitor/camera: 1.0.5

### Capacitor Version
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 3.2.2
  @capacitor/core: 3.2.2
  @capacitor/android: 3.2.2
  @capacitor/ios: 3.2.2

Installed Dependencies:

  @capacitor/cli: 3.0.0
  @capacitor/core: 3.0.0
  @capacitor/android: 3.0.0
  @capacitor/ios: 3.0.0

[success] iOS looking great! 👌

### Platform(s)
Android

### Current Behavior
saveToGallery is set to true

1. User grants access to Camera
2. User denies access to Photos
3. cameraPermissionsCallback does not reject the call and repeats with do doShow(call)
4. doShow(call) -> User is asked again  for Photo Permissions -> 2.

Loop (especially if User says do not ask again when denying access)

### Expected Behavior
1. User grants access to Camera
2. User denies access to Photos
3. cameraPermissionsCallback rejects the call

llendi:
…Callback when isSaveToGallery is true

Fixes #593

### DIFF
--- a/camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
+++ b/camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
@@ -65,6 +65,7 @@ public class CameraPlugin extends Plugin {
     private static final String INVALID_RESULT_TYPE_ERROR = "Invalid resultType option";
     private static final String PERMISSION_DENIED_ERROR_CAMERA = "User denied access to camera";
     private static final String PERMISSION_DENIED_ERROR_PHOTOS = "User denied access to photos";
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Thanks for the PR.
We have checked iOS behaviour and there, if the user doesn't grant the permission, the image is still returned instead of an error. 
So I've sent a PR that makes Android to behave the same way and removes the infinite loop, so I'm closing yours.
https://github.com/ionic-team/capacitor-plugins/pull/599

PR #601, opened 2021-09-09
Added weekday functionality for local-notifications

### LINKED ISSUE #129: feat: `weekday` support in LocalNotifications
# Feature Request

## Describe the Feature Request
I'd like to schedule recurring local notifications on Tuesdays and Thursdays at a given hour.

## Platform Support Requested

- Android
- iOS

## Describe Preferred Solution
I'd like to schedule it using the `on` keyword together with a (new) `weekday` (0-6) keyword. 

## Describe Alternatives
I don't see a way around this, as `day` is only 0-30. I could use `every` — but that doesn't let me set the hour of delivery. https://github.com/ionic-team/capacitor/issues/2752 goes into further detail.

## Related Code
```...
schedule: {
  on: {
    hour: 20,
    weekday: 2
  }
},
...
schedule: {
  on: {
    hour: 20,
    weekday: 4
  }
},
```

## Additional Context
- iOS: [DateComponent](https://developer.apple.com/documentation/foundation/datecomponents) has [`weekday`](https://developer.apple.com/documentation/foundation/datecomponents/1780094-weekday), support so it could easily be added in [`LocalNotifications.swift`](https://github.com/ionic-team/capacitor/blob/master/ios/Capacitor/Capacitor/Plugins/LocalNotifications.swift#L276) from what I gather. 
- Android: I don't know Java but it seems in [`LocalNotificationsSchedule.java`](https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationSchedule.java) we are using [`SimpleDateFormat`](https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationSchedule.java) which has a [`u` — day number of week](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html) (1-7) support.

kim-jos:
Fixes #129 
How I fixed the problem problem:
1. No work was done for IOS because IOS already has a built-in weekday feature (0 - 6) (Sun - Sat)
2. As can be seen in the LocalNotificationSchedule.java file, I added 1 to the weekday integer because Java implements weekdays (1 - 7) (Sun - Sat) which is different from IOS.

The following code would now work.
```
schedule: {
  on: {
    hour: 20,
    weekday: 0  //Sunday
  }
},
```

### DIFF
--- a/.github/workflows/publish-android.yml
+++ b/.github/workflows/publish-android.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
-        ref: publish-android
+        ref: native-publish
     - name: set up JDK 11
       uses: actions/setup-java@v2
       with:
--- a/.github/workflows/publish-ios.yml
+++ b/.github/workflows/publish-ios.yml
@@ -1,0 +1,38 @@
+name: Publish Native iOS Libraries
+
+on:
+  workflow_dispatch:
+    inputs:
+      plugins:
+        description: 'Specify an optional subset of plugins to publish (space delimited)'
+        required: false
+        default: ''
+jobs:
+  publish-ios:
+    runs-on: macos-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
+      - uses: actions/checkout@v2
+        with:
+          ref: native-publish
+      - name: Install Cocoapods
+        run: | 
+          gem install cocoapods
+      - name: Restore Dependency Cache
+        id: cache-modules
+        uses: actions/cache@v2
+        with:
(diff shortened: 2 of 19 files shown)


jcesarmobile (changes requested):
Thanks for the PR!

I've been testing and it doesn't work on iOS, looks like it needs code changes in the `getDateComponents` function for getting the weekday param.
I've put some code there but then it seems that iOS starts on 1 instead of 0 as you said.

jcesarmobile:
closing in favor of https://github.com/ionic-team/capacitor-plugins/pull/756 as it addressed the changed I requested here

PR #612, opened 2021-09-16
Add missing `await` statement

peschee:
The JS code examples are missing `await` statements for functions that return promises.


jcesarmobile (approved):
Thanks for the PR!

Since the plugin doesn't return anything it's not really needed to await for the promise unless you want your code to wait, but since other plugins that return nothing have await in the examples I'll merge for consistency.

Issue #621, opened 2021-09-18
Keyboard opening causes webview to resize to 0x0 in a "scenes" based iOS app


joeflateau:
### Plugin(s)

@capacitor/keyboard 1.0.1

### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 3.2.3
  @capacitor/core: 3.2.3
  @capacitor/android: 3.2.3
  @capacitor/ios: 3.2.3

Installed Dependencies:

  @capacitor/cli: 3.0.1
  @capacitor/core: 3.0.1
  @capacitor/android: 3.0.1
  @capacitor/ios: 3.0.1
```

### Expected Behavior

When triggering the keyboard to open in a scenes-based iOS app, the webview is resized to 0x0. This is because the keyboard plugin reads window size from `UIApplication.shared.window.bounds ` but `UIApplication.shared.window` is not available in a Scenes based app.

### Code Reproduction

- Create an app using scenes: https://developer.apple.com/documentation/uikit/app_and_environment/scenes/specifying_the_scenes_your_app_supports
- Put a text input on the page
- Focus the text input

joeflateau:
SplashScreen is also affected (it is not visible)

joeflateau:
Remove @capacitor/keyboard and install https://www.npmjs.com/package/@justicointeractive/capacitor-keyboard to see the bug fixed

PR #622, opened 2021-09-18
bugfix(keyboard) keyboard opening causes app to disappear in scenes app

joeflateau:
Fixes #621

PR #623, opened 2021-09-19
bugfix(splashscreen) splashscreen not visible in scenes app

joeflateau:
Referenced by #621

### DIFF
--- a/splash-screen/ios/Plugin/SplashScreen.swift
+++ b/splash-screen/ios/Plugin/SplashScreen.swift
@@ -9,6 +9,7 @@ import Capacitor
     var config: SplashScreenConfig = SplashScreenConfig()
     var hideTask: Any?
     var isVisible: Bool = false
+    var wasAddedToViewHeirarchy: Bool = false
 
     init(parentView: UIView, config: SplashScreenConfig) {
         self.parentView = parentView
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Thanks for the PR!

I get this error on line 131
> Variable used within its own initial value 

Anyway, there is a similar pull request and I'm merging that one since I like it more 
https://github.com/ionic-team/capacitor-plugins/pull/631

joeflateau:
@jcesarmobile thanks! want me to redo my other PR (#622) to find the window the way #631 does?

jcesarmobile:
Thanks for the offering, but I was already working on it.

If you can verify if my pull request fix your issue it would be helpful
https://github.com/ionic-team/capacitor-plugins/pull/729

joeflateau:
@jcesarmobile I ran the app with #631 and #729 fixes installed and the app worked as expected. I had concerns about finding the window by looking at the first scene rather than finding the owner window from the cap web view, so I tried:

- in info.plist, phone scene listed first before carplay, launch the phone scene first
- in info.plist, carplay scene listed first, launch phone scene first
- in info.plist, carplay scene listed first, launch carplay scene first
- in info.plist, phone scene listed first, launch carplay first

All worked as expected, so does seem like there's some internal consistency that the phone scene will be the `first` scene and CarPlay second, but on iPads where you can have your own app side-by-side with itself I think you'd end up with the wrong dimensions if the cap view is not in the first scene. But for my own purposes, these resolve my issues.

jcesarmobile:
If it works for you then I think it's ok.
Once we drop iOS 12 support we might make the default template a scenes app as Xcode does and then more people could face issues, if that happens we will need to revisit.

PR #624, opened 2021-09-20
docs(haptics): iOS devices older than iPhone 7 only support vibrate(...)

ilindaniel:
See comment by C6Silver below this answer: https://stackoverflow.com/a/39957091
I tested this on an iPhone 6s running iOS 14.8 and I can confirm that `impact(...)` and `notification(...)` are not working.

### DIFF
--- a/haptics/README.md
+++ b/haptics/README.md
@@ -2,6 +2,8 @@
 
 The Haptics API provides physical feedback to the user through touch or vibration.
 
+**Note:** On iOS, devices older than iPhone 7 only support `vibrate(…)` as haptic feedback.
+
 ## Install
 
 ```bash


jcesarmobile:
In any case it should mention than the Taptic Engine hardware is required, not iPhone models as iOS platform also supports iPods and iPads and none of those have Taptic Engine despite being newer than iPhone 7

jcesarmobile (changes requested):
In any case it should mention than the Taptic Engine hardware is required, not iPhone models as iOS platform also supports iPods and iPads and none of those have Taptic Engine despite being newer than iPhone 7

ilindaniel:
Made some changes. Unfortunately I couldn't find an official list of devices with the 2nd Gen Taptic Engine

ilindaniel:
I did some research and figured out that iPads and iPods do not have the Apple Taptic Engine built in.
I think we can put the "iPhone 7 and above" notice back.
What do you think?

ilindaniel:
Similar to this Cordova plugin: https://github.com/EddyVerbruggen/cordova-plugin-taptic-engine#supported-platforms

jcesarmobile (changes requested):
I've added duration to vibration on iOS, which also requires taptic engine, so to describe everything, I think it would be best to document this per function (in the src/definitions.ts file and then run npm run build so the changes are also added to the README.md) instead of a general note.
At the moment all methods require taptic engine on iOS (also the selection one that you missed) and vibration to use the duration (but this one will fall back to regular vibration where there isn't taptic enginer but there is a vibrator). No iPod touch nor iPad device have taptic engine nor vibrator.

Also, on Android there are devices without vibrator, so not sure if it should be noted too.

jcesarmobile:
the readme now mentions that if there is no Taptic engine or Vibrator the plugin does nothing. It doesn't enter into this level of detail, but should be enough.

PR #631, opened 2021-09-24
feat(splash-screen): Make splash work in apps that use scenes

RishatShamsutdinov:
On iOS 13 there is scenes (`UIScene`). When you're using scenes, the `window` property of the `UIApplicationDelegate` is `nil`.


jcesarmobile (approved):
Thanks for the PR!

I've made some minor changes, one for making sure the window we get is the key window, and another to avoid the force unwrap of the window.

PR #638, opened 2021-09-27
[ios] Fix a bug with geolocation/az when skipping the camera/camera roll prompt.

gabeschine:
* Replace null-initialized instances of CLLocation and CLHeading with optionals
      since using the default constructors for those is not supported behavior.
    * Short-circuit code to add EXIF data if a location has not been found for some reason.
    * Start the location/az acquisition when the camera is shown: this codepath is guaranteed
      to be called even if the call instructs the plugin to skip the prompt.
    
    Tested: manually on iPhone 12 Pro


gabeschine:
Sorry wrong target repository!

PR #665, opened 2021-10-22
fix(camera): set direction on web

Leo1305Pineda:
Is require PR to this from [PR #93](https://github.com/ionic-team/pwa-elements/pull/93)
Resolved open rear camera from initial opening.


camera/src/web.ts
```diff
@@ -47,6 +47,7 @@ export class CameraWeb extends WebPlugin implements CameraPlugin {
   ) {
     if (customElements.get('pwa-camera-modal')) {
       const cameraModal: any = document.createElement('pwa-camera-modal');
+      cameraModal.facingMode = options.direction === CameraDirection.Rear ? 'environment' : 'user';
```
jcesarmobile, line 50: Since `CameraDirection.Rear` is the default, the conditional should check for `CameraDirection.Front`, otherwise if `options.direction` is undefined it will use front camera as undefined is different from `CameraDirection.Rear`

```suggestion
      cameraModal.facingMode = options.direction === CameraDirection.Front ? 'user' : 'environment';
```

Aarbel:
@Leo1305Pineda do you plan to update your PR ?

Leo1305Pineda:
> 

I have a build of this and it works fine for web, ios, and android.

Aarbel:
@Leo1305Pineda sure, we all have our custom fixes of web rtc camera, but would be really great to write a few tests to make it open-source friendly ;)

camera/src/web.ts
Leo1305Pineda: image
For possible cases it delivers the expected result.
[https://stackoverflow.com/a/52812734/7638125](https://stackoverflow.com/a/52812734/7638125)

Leo1305Pineda: Only this change has to be tested with this
[https://github.com/ionic-team/pwa-elements/pull/93](https://github.com/ionic-team/pwa-elements/pull/93)

jcesarmobile: If no direction is passed it should use rear (environment) because it’s the default value, your image shows that if no direction is provided (undefined) it uses front camera (user)

Leo1305Pineda: image
So with this cameraModal.facingMode = options.direction === CameraDirection.Front ? 'user' : 'environment';
Fail in 
image

jcesarmobile: Check the proposed change, it’s 
`cameraModal.facingMode = options.direction === CameraDirection.Front ? 'user' : 'environment';`

Leo1305Pineda: image
ok, if I have to make another adjustment

Leo1305Pineda: image
and like this

Leo1305Pineda: image

Leo1305Pineda: image
image
now yes

Leo1305Pineda:
This PR is also required [pwa-elements](https://github.com/ionic-team/pwa-elements/pull/93)

camera/src/web.ts
jcesarmobile: While the last change works as expected, I see unnecessary to create two objects for a simple value check and that approach doesn't align with the style used in the other plugins.
The change I requested was just reversing the order of your initial one line check so environment was the default.
Can you do that?

Aarbel:
@Leo1305Pineda have you tested it on a Microsoft Surface device?

The devices API isn't standard (browser of Microsoft surface doesn't provide all the necessary metadata of devices). 
So it may affect a lot of users if it doesn't work

camera/src/web.ts
Leo1305Pineda: yes

Leo1305Pineda: verification remains like this:
cameraModal.facingMode = options.direction === CameraDirection.Front ? 'user' : 'environment';

Leo1305Pineda:
In pwa-elements there was no entry @Prop () facesMode: FacingMode;
this was already applied in this PR [pwa-elements/pull/93](https://github.com/ionic-team/pwa-elements/pull/93)

jcesarmobile (approved):

jcesarmobile:
Thanks for the pull request!

PR #674, opened 2021-11-01
feat: Disable analytics

saschpe:
Avoid metrics collection by default as this lacks consent required by
the GDPR in Europe. It most likely violates Californian privacy laws as
well. Additionally, metrics collection on dependencies might expose
sensitive information in proprietary projects. Finally, those additional
network requests have an impact on overall build-time, especially for
CocoaPods.

Relates-To: https://github.com/ionic-team/capacitor/issues/5191


thomasvidas:
Closing since if this is a GDPR or CCPA issue, it should be fixed in upstream tooling. See my comment here https://github.com/ionic-team/capacitor/issues/5191#issuecomment-964318230

PR #677, opened 2021-11-02
fix(android): Integer conversion of compileSdk/minSdk/targetSdk

saschpe:
Gradle properties are always parsed as strings from `gradle.properties`.
While Gradle Groovy can utilize the `ext` Gradle plugin directly, Gradle
Kotlin DSL projects have to specify properties in `gradle.properties`.
Convert those to integers as required by Android Gradle Plugin (AGP).
This won't affect projects that still use Gradle Groovy.

Resolves #https://github.com/ionic-team/capacitor/issues/5194


saschpe:
Final ping, I think this is still relevant. Worth rebasing once more?

PR #681, opened 2021-11-04
docs (local-notifications): Specify how to use the default sound for notifications in iOS

dtarnawsky:


jcesarmobile:
Sound is an optional parameter, so it’s better to not pass it than passing an empty string. 
There is a problem in the docs that don’t show which params are optional, I reported it some time ago. But it’s clear if you check the types. Docs are generated from the types, but there is nothing to differentiate optional params from mandatory params. You shouldn’t edit type information from the readme, you should edit the types and run npm run build to generate the readme

https://github.com/ionic-team/capacitor-site/issues/367

Issue #704, opened 2021-11-26
Cannot select photos from Gallery on Samsung devices


Odonno:
### Plugin(s)

"@capacitor/android": "^3.3.1",
"@capacitor/app": "^1.0.6",
"@capacitor/camera": "^1.2.1",

### Capacitor Version

Latest Dependencies:

  @capacitor/cli: 3.3.2
  @capacitor/core: 3.3.2
  @capacitor/android: 3.3.2
  @capacitor/ios: 3.3.2

Installed Dependencies:

  @capacitor/cli: 3.3.1
  @capacitor/ios: 3.3.1
  @capacitor/core: 3.3.1
  @capacitor/android: 3.3.1

### Platform(s)

Android : Samsung Note 8

### Current Behavior

When using `pickImages` function on a Samsung device, the gallery is opened but I cannot select image. I can see a list of images but I can only open the image in fullscreen. There is no way to select an image, even one.

### Expected Behavior

Like any other gallery app, add ability to select multiple images.

Odonno:
Seems like the react-native team already solved a similar issue here: https://github.com/react-native-image-picker/react-native-image-picker/issues/1739

ThaierAlkhateeb:
Hey, 
I have got the same issue.
It shows me in console "{message: 'No images picked'}", but if I choose select image from Google photos it works.

raykin:
Totally same issue on OPPO phone. I captured several screenshots of adb logs. I can upload them if it can help to locate issues.

carlpoole:
I was able to confirm that multi-selecting is working when the Google Photos app is used as the source, but running into issues with the device/manufacturer gallery apps. I was testing the solution proposed #706 but I was still running into issues with the picker that displayed (see: https://github.com/ionic-team/capacitor-plugins/pull/706#issuecomment-981734589) and I was curious if this was the same experience on other devices. I was testing it on a Samsung Galaxy S9+ running Android 10

carlpoole:
In the case with my test on my Galaxy S9+, single tapping an image caused it to immediately try and return that image to the app, instead of allowing me to select more and then confirm. The presence of only a single image causes the `processPickedImages` method to fail at the moment.

To select multiple images I had to press and hold in the device picker, and then it would let me choose more and confirm. This leads me to think we could modify the multi-select code to allow the return of single images and leave it to the users to make sure they are using the device picker correctly to pick multiple images vs. a single image.

N0ctural:
[...]

PR #706, opened 2021-11-26
feat(camera): Support for Samsung Gallery app on pickImages

N0ctural:
Hello,

Fix to correct issue with multiple image selection on samsung devices (https://github.com/ionic-team/capacitor-plugins/issues/704)

closes https://github.com/ionic-team/capacitor-plugins/issues/704


carlpoole:
I am testing this on my Samsung Galaxy 9+ on Android 10 and the picker that shows encounters similar issues to the issue. Single-tapping an item causes the picker to disappear and go back to the app with an error. I can long-press items to multi-select but it isn't super obvious. If I select a different app in "Browse files in other apps" such as the Gallery, it has the same issue as the picker did before when selecting the native Gallery app. 

Screenshot_20211129-091152_Files

N0ctural:
Yes, you're right @carlpoole.  I made a change on multi-select method to avoid the error "No image picked" if only one picture is selected. It's works fine on a OnePlus with Android 9.

dftd:
@N0ctural Thanks for fixing this! Is the code ready to be used or you are still working on something?
I need the fix asap.

carlpoole:
Thanks for the update, I tested this again and it seems to be working great now. On my device it is asking which gallery app to use and if I pick the Samsung gallery app, I have the option of selecting multiple and returning or just choosing one. 

I checked what you mentioned about permissions being necessary for external storage but if I try to pick an image off an SD card it seemed to work fine without requesting permissions. Were you testing that on a specific version of Android/Samsung gallery app?

N0ctural:
> Thanks for the update, I tested this again and it seems to be working great now. On my device it is asking which gallery app to use and if I pick the Samsung gallery app, I have the option of selecting multiple and returning or just choosing one.
> 
> I checked what you mentioned about permissions being necessary for external storage but if I try to pick an image off an SD card it seemed to work fine without requesting permissions. Were you testing that on a specific version of Android/Samsung gallery app?

Good news !
About the request permission It was on Samsung A10 with Android 9. I use Browserstack to test the fix (maybe It's a specific problem with their platform).

carlpoole:
I think it's related to the Android version. It seems to work fine for Android 10+ but so far my test on Android 9 and below I do get a crash for missing External permission.

N0ctural:
> I think it's related to the Android version. It seems to work fine for Android 10+ but so far my test on Android 9 and below I do get a crash for missing External permission.

It's maybe specific behaviour with android version and samsung device. (No problem with that on OnePlus / Android 9)
I don't know if the good solution is to ask for permission on all android version or add a specific patch with something like : 

if(Build.VERSION.SDK_INT <= 9.0 && Build.BRAND.equals("Samsung")) {                
  // ask for permission
}

Or maybe just add checkpermission in the web part.

jcesarmobile (approved):

jcesarmobile:
I've added permission prompt/check, but will continue with the image picking even if denied, in that case it will catch the security exception in devices that require the permissions and reject, or return the images in devices that don't require the permissions

PR #709, opened 2021-11-29
feat(geolocation): Error if Google Play Services are not available

jcesarmobile:
Since the plugin relies on the Fused Location Provider, we should return an error if Google Play Services are not available


carlpoole (approved):
🎉

PR #733, opened 2021-12-09
fix(camera): handle reject of pickImages

Odonno:
When the user reject the upload of images (the "Cancel" button in the dialog), there is no way to detect that rejection event.

Now, I am using the focus event on the window to detect when the web app come back to life. If no resolved event was fired, then we consider the user has rejected the gallery upload.


Odonno:
Closing this PR as it is only working on web browsers, not working on mobile browsers.

PR #743, opened 2021-12-21
feat(android): support for minimizeApp

nothingkid:
I think it makes sense if this small feature will be part of app plugin.

closes https://github.com/ionic-team/capacitor-plugins/issues/130


app/ios/Plugin/AppPlugin.m
```diff
@@ -9,4 +9,5 @@
            CAP_PLUGIN_METHOD(getLaunchUrl, CAPPluginReturnPromise);
            CAP_PLUGIN_METHOD(getState, CAPPluginReturnPromise);
            CAP_PLUGIN_METHOD(removeAllListeners, CAPPluginReturnNone);
+           CAP_PLUGIN_METHOD(minimizeApp, CAPPluginReturnNone);
```
jcesarmobile, line 12: can you move it up before removeAllListeners?

app/ios/Plugin/AppPlugin.swift
```diff
@@ -92,4 +105,8 @@ public class AppPlugin: CAPPlugin {
             ])
         }
     }
+
+    @objc func minimizeApp(_ call: CAPPluginCall) {
+        throw AppPluginError.notImplemented
```
jcesarmobile, line 110: iOS code is not compiling because it should be `case .notImplemented:` instead of `switch .notImplemented:` and also complains about the throw.
You should just use the unimplemented call method like this and you can remove the whole `AppPluginError` enum.
```suggestion
        call.unimplemented()
```

app/src/definitions.ts
```diff
@@ -231,6 +231,11 @@ export interface AppPlugin {
    * @since 1.0.0
    */
   removeAllListeners(): Promise<void>;
+
+  /**
+   * Minimizes application
+   */
+  minimizeApp(): Promise<void>;
```
jcesarmobile, line 238: can you move it up before all the listener related functions?
Also, can you include a line for `@since 1.1.0` as that would be the version where the feature will be included and mention that the feature is only available for Android?

app/android/src/main/java/com/capacitorjs/plugins/app/AppPlugin.java
```diff
@@ -105,6 +105,15 @@ public void getState(PluginCall call) {
         call.resolve(data);
     }
 
+    @PluginMethod
+    public void minimizeApp(PluginCall call) {
```
jcesarmobile, line 109: this is not aligned and will cause lint command to fail, can you run `npm run fmt` command? that will fix all lint issues

jcesarmobile (changes requested):
Made a few code change requests, can you resolve them?

Thanks for the PR!

app/ios/Plugin/AppPlugin.m
jcesarmobile: And the return type should be `CAPPluginReturnPromise` since in the types you said it returns `Promise<void>;`

nothingkid:
Thanks for review. Added changes.

jcesarmobile:
Not all comments are addressed, and you have removed a few android test classes, can you address the missing comments and put those classes back?

jcesarmobile (changes requested):
I've updated my comments with the things that are still missing.
And also the test java classes that were removed should be put back

nothingkid:
Hm. Looks like fmt script removed classes. I will address all comments. I'm sorry that I missed these things

jcesarmobile (approved):

jcesarmobile:
did some minor changes on texts, but other than that looks good, thanks!

PR #753, opened 2022-01-10
Update README.md

AVATAR197:
Capacitor haptics readme.md fix: ImactStyle Heavy description changed.


jcesarmobile (changes requested):
Thanks for the PR, good catch!

But README is generated from the types, so should be changed on https://github.com/ionic-team/capacitor-plugins/blob/main/haptics/src/definitions.ts#L62 and then running `npm run build` should generate the README

Also, we use the exact texts as on [Apple docs](https://developer.apple.com/documentation/uikit/uiimpactfeedbackgenerator/feedbackstyle/heavy), so can you change the text to `A collision between large, heavy user interface elements`

PR #756, opened 2022-01-11
feat(local-notifications): Add weekday scheduling support for Android and iOS

### LINKED ISSUE #129: feat: `weekday` support in LocalNotifications

nkalupahana:
Fixes #129!

The past PR that did this (#601) has gone stale, so I figured I'd do it myself and verify it. Weekdays go from 1 (Sunday) to 7 (Saturday), as dictated by both the iOS and Android standards. I've tested this functionality on both iOS and Android and it all seems to work appropriately.


nkalupahana:
@jcesarmobile linting is fixed; if you could re-approve workflows that would be great :)

jcesarmobile (changes requested):
Can you create an enum for the days, something like this
```
enum Days {
  Sunday = 1,
  Monday,
  Tuesday,
  Wednesday,
  Thursday,
  Friday,
  Saturday
}
```
and assign that enum type to weekday?

That way it's more clear Sunday it's the 1, and also people can use it like `Days.Sunday` instead of using the numbers directly

nkalupahana:
@jcesarmobile all set, let me know if this works

jcesarmobile (changes requested):
Can you add numbers to all days? 

I was hopping the doc generator would be smart enough to generate the missing numbers by incrementing the first one, but looks like it doesn't.

nkalupahana:
@jcesarmobile ✅

nkalupahana:
@jcesarmobile sorry about that, I thought I ran linter but apparently I didn't.

jcesarmobile (approved):
looks good, I scheduled notifications for all week and left it running over the weekend and everything fired when expected

mahen23:
> 

Hello, i want notifications to run only on weekdays(monday - friday), no weekends. How to i set it ?

nkalupahana:
@mahen23 You can schedule 5 separate notifications -- one for each weekday you want the notification to be on.

mahen23:
> @mahen23 You can schedule 5 separate notifications -- one for each weekday you want the notification to be on.

If i set a notification to be triggered at 9am, if i left the minute blank, its never triggered: 

on: {weekday: 4, hour: 9 } => this will never be triggered, be it 9:00 or 09:10 or 09:59
on: {weekday: 4, hour: 9, minute: 4 } => this will be triggered at only 09:04. 

is there a way i can configure notification to be triggered at any time from 09:00 to 09:59 ?

PR #776, opened 2022-01-24
fix(splash-screen): avoid conditional downcast warning

### LINKED ISSUE #761: Conditional downcast from 'UIWindow?' to 'UIWindow' does nothing
https://github.com/ionic-team/capacitor-plugins/blob/3ec2008bc0311c837346ecbdd0afd105149b427a/splash-screen/ios/Plugin/SplashScreen.swift#L120

The line produces a warning while compiling an iOS application in Xcode 13.2.1.

riderx:
closes #761

### DIFF
--- a/splash-screen/ios/Plugin/SplashScreen.swift
+++ b/splash-screen/ios/Plugin/SplashScreen.swift
@@ -117,7 +117,7 @@ import Capacitor
     // Update the bounds for the splash image. This will also be called when
     // the parent view observers fire
     private func updateSplashImageBounds() {
-        var window: UIWindow? = UIApplication.shared.delegate?.window
+        var window: UIWindow? = UIApplication.shared.delegate?.window ?? nil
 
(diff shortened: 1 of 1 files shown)


splash-screen/ios/Plugin/SplashScreen.swift
```diff
@@ -117,7 +117,7 @@ import Capacitor
     // Update the bounds for the splash image. This will also be called when
     // the parent view observers fire
     private func updateSplashImageBounds() {
-        var window: UIWindow? = UIApplication.shared.delegate?.window as? UIWindow
+        var window: UIWindow? = UIApplication.shared.delegate?.window
```
jcesarmobile, line 120: Thanks for the PR, but the code doesn't compile, I think it should look like this:
```suggestion
        var window: UIWindow? = UIApplication.shared.delegate?.window ?? nil
```

jcesarmobile (approved):
looks good, thanks!

Issue #750, opened 2022-01-07
App.exitApp() should return promise


jcesarmobile:
### Plugin(s)
App

### Capacitor Version
any

### Platform(s)
types

### Current Behavior
exitApp returns never

### Expected Behavior
exitApp should return a promise

### Other Technical Details
On android exitApp kills the app, so doesn't matter what it returns, but on iOS and web it rejects the promise and can't be catch as it returns never, if it returns a promise users can catch the unimplemented error

numerized:
if I'm not wrong, you cannot kill an app on iOS

jcesarmobile:
Yeah, you can’t. But at the moment calling exitApp on iOS will return an error that can’t be catch because it doesn’t return a promise, so the issue is about fixing that.

jcesarmobile:
this has been fixed in 4.0.0 beta 0

PR #777, opened 2022-01-24
feat(app): return promise in exitApp

riderx:
document the type change in v2 of the plugin

closes https://github.com/ionic-team/capacitor-plugins/issues/750


jcesarmobile (changes requested):
Thanks for the PR

You also need to change `CAP_PLUGIN_METHOD(exitApp, CAPPluginReturnNone);` to `CAP_PLUGIN_METHOD(exitApp, CAPPluginReturnPromise);` in `AppPlugin.m` file.

We are not merging this yet as it's a breaking change and it's planned for when we start working in Capacitor 4 and version 2 of the plugins.

riderx:
@jcesarmobile thanks for the report. Do we have a doc somewhere do learn oh to code and test ?

jcesarmobile:
we test on the test-app, check its readme
https://github.com/ionic-team/capacitor-testapp#developing-capacitor

riderx:
My issue is more about installing package and using git with Lerna, i see 0 changes do i need a specific VSCode plugin ?

jcesarmobile:
If you run npm install on the root, lerna runs npm install on all plugins, same if you run npm run build, lerna runs it for all.
But you can also go to the plugin folder and run the same commands there, it will be faster as it will only run for that plugin instead of all.

riderx:
@jcesarmobile it seems from the doc i have to install Lerna in global right ? https://lerna.js.org/

jcesarmobile:
Not sure, it’s on the devDependencies, so shouldn’t be needed globally, but if it’s not working then you can give it a try.

riderx:
@jcesarmobile yes with Global install that do way different install flow ! Now i'm able to run all commands !

riderx:
@jcesarmobile i think that good in my side

app/ios/Plugin/AppPlugin.m
```diff
@@ -9,5 +9,5 @@
            CAP_PLUGIN_METHOD(getLaunchUrl, CAPPluginReturnPromise);
            CAP_PLUGIN_METHOD(getState, CAPPluginReturnPromise);
            CAP_PLUGIN_METHOD(minimizeApp, CAPPluginReturnPromise);
-           CAP_PLUGIN_METHOD(removeAllListeners, CAPPluginReturnNone);
+           CAP_PLUGIN_METHOD(removeAllListeners, CAPPluginReturnPromise);
```
jcesarmobile, line 12: You have to change it in `exitApp`, not in `removeAllListeners`
```suggestion
           CAP_PLUGIN_METHOD(removeAllListeners, CAPPluginReturnNone);
```

riderx:
Sorry @jcesarmobile  i was sick yesterday :/ i updated

riderx:
@jcesarmobile good for you ?

jcesarmobile (approved):

PR #780, opened 2022-01-24
Update README.md

BuffMcBigHuge:
The previous method results in the following error:

```
event capacitorDidRegisterForRemoteNotifications not called.  Visit https://capacitorjs.com/docs/apis/push-notifications for more information
```


jcesarmobile:
You might be doing something wrong, the docs are correct.
The changes you include use FCM, but we use APNS as that’s the official way of using push.
We have a separate guide for FCM
https://capacitorjs.com/docs/guides/push-notifications-firebase

PR #781, opened 2022-01-26
Update README.md

ameer:
fix a typo


jcesarmobile (changes requested):
thanks for the PR

the docs are generated from the types, so the typo should be fixed on the types and then run `npm run build` so the docs get re generated
https://github.com/ionic-team/capacitor-plugins/blob/d97898662d0ba037e5f8448990a91de5ec6a4234/toast/src/definitions.ts#L27

PR #793, opened 2022-01-28
Add Electron support to the AppLauncher plugin

jdgjsag67251:
This pull-request implements the AppLauncher plugin for [Electron](https://github.com/capacitor-community/electron).
For security purposes it only allows the HTTP protocol by default. Other protocols can be specified in the `capacitor.config.ts` file like so:
```ts
const config: ElectronCapacitorConfig = {
  appId: 'com.example.test',
  appName: 'Test app',
  webDir: 'www',
  plugins: {
    'app-launcher': {
      electron: {
        protocols: ['http', 'https', 'mailto'],
      },
    },
  },
};
```
Protocol checking can be disabled by setting `protocols` to `false`.


eric-horodyski:

More context:

PR #819, opened 2022-02-11
docs: replace alert with console.log in example section

jonz94:
Because `alert()` only accept one argument, the second argument `value` will be ignored, and thus the `value` will not be shown on screen in the original example.


jcesarmobile (changes requested):
Can you instead replace the alert with a console.log?
There is another issue open about not using alerts in examples

jonz94:
> Can you instead replace the alert with a console.log? There is another issue open about not using alerts in examples

Sure! 

I also checked the example of other plugins, and replace the alert with console.log

jcesarmobile (approved):

jcesarmobile:
thanks!

PR #830, opened 2022-02-22
chore: Update AndroidX dependencies

jcesarmobile:


giralte-ionic (approved):

jcesarmobile:
There seems to be some problem with core 1.7.0 and app compat 1.4.1, so I've downgraded them for now.
Will update them once we have a Capacitor 4 alpha version that we can point to as I think the conflict is caused by using version 3.

IT-MikeS (approved):

PR #857, opened 2022-03-09
feat: Inline Scrolling

theproducer:
This PR adds inline map scrolling support.


IT-MikeS (approved):
With lint fixes this looks good to me!

IT-MikeS (approved):

PR #872, opened 2022-03-14
refactor(keyboard): replace deprecations

### LINKED ISSUE #843: deprecation warnings on keyboard plugin
### Plugin
keyboard

### Description
Handler() in Handler has been deprecated
getDefaultDisplay() in WindowManager has been deprecated
getMetrics(DisplayMetrics) in Display has been deprecated
getStableInsetBottom() in WindowInsets has been deprecated
getSize(Point) in Display has been deprecated
getSystemUiVisibility() in View has been deprecated
SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN in View has been deprecated
toggleSoftInput(int,int) in InputMethodManager has been deprecated

### Platform(s)
android

### Preferred Solution
remove or silence warnings

IT-MikeS:
closes #843


keyboard/android/src/main/java/com/capacitorjs/plugins/keyboard/Keyboard.java
```diff
@@ -60,7 +61,11 @@ public Keyboard(AppCompatActivity activity, boolean resizeOnFullScreen) {
         //calculate density-independent pixels (dp)
         //http://developer.android.com/guide/practices/screens_support.html
         DisplayMetrics dm = new DisplayMetrics();
-        activity.getWindowManager().getDefaultDisplay().getMetrics(dm);
+        if (android.os.Build.VERSION.SDK_INT >= 30) {
```
jcesarmobile, line 64:
```suggestion
        if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
```

use the constant instead of the hardcoded value

keyboard/android/src/main/java/com/capacitorjs/plugins/keyboard/Keyboard.java
```diff
@@ -84,7 +89,11 @@ public void onGlobalLayout() {
                     int resultBottom = r.bottom;
                     int screenHeight;
 
-                    if (Build.VERSION.SDK_INT >= 23) {
+                    if (Build.VERSION.SDK_INT >= 30) {
+                        Insets windowInsets = rootView.getRootWindowInsets().getInsetsIgnoringVisibility(WindowInsets.Type.systemBars());
+                        screenHeight = rootViewHeight;
+                        resultBottom = windowInsets.bottom;
+                    } else if (Build.VERSION.SDK_INT >= 23) {
```
jcesarmobile, line 96:
```suggestion
                    } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
use the constant instead of the hardcoded value

jcesarmobile (changes requested):
Despite the deprecation warnings are being handled on newer Android versions, the compiler will still complain about them since the code is still there. 
Add some `@SuppressWarnings("deprecation")` to the functions using the deprecated code.

giralte-ionic (approved):

giralte-ionic (approved):

giralte-ionic (approved):

keyboard/android/src/main/java/com/capacitorjs/plugins/keyboard/Keyboard.java
```diff
@@ -132,10 +142,7 @@ private int computeUsableHeight() {
 
                 private boolean isOverlays() {
                     final Window window = activity.getWindow();
-                    return (
-                        (window.getDecorView().getSystemUiVisibility() & View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN) ==
-                        View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN
-                    );
+                    return window.getDecorView().getFitsSystemWindows();
```
jcesarmobile, line 145: This is always returning false, even if I overlay the status bar with status bar plugin, I think we can keep the old code for now and use a suppress warning in the funtion

keyboard/android/src/main/java/com/capacitorjs/plugins/keyboard/Keyboard.java
```diff
@@ -55,12 +56,17 @@ public Keyboard(AppCompatActivity activity) {
         this(activity, false);
     }
 
+    @SuppressWarnings("deprecation")
```
jcesarmobile, line 59: Instead of adding this to suppress all warnings, try to split the code that uses deprecated code into new functions and add the suppress warnings to that functions only, so in the future we don't miss other deprecated code that will be hidden by this annotation.

jcesarmobile (changes requested):
There is still a warning left because of the other init method that has @deprecated on the comments but not on the function.
But you can remove the whole init method since we are doing a major release.

jcesarmobile (approved):

Issue #840, opened 2022-02-28
insertImage(ContentResolver,String,String,String) in Media has been deprecated


jcesarmobile:
### Plugin
camera

### Description
insertImage(ContentResolver,String,String,String) in Media has been deprecated

### Platform(s)
android

### Preferred Solution
remove or silence deprecation warning

PR #874, opened 2022-03-15
refactor(camera): Replace deprecated code 

theproducer:
closes: https://github.com/ionic-team/capacitor-plugins/issues/840


jcesarmobile (changes requested):
The conditional code doesn't prevent the deprecation warning, you can add a `@SuppressWarnings("deprecation")` to the function containing the code to suppress it since it's being handled.

jcesarmobile (approved):

PR #877, opened 2022-03-17
chore: set plugins deployment target to 13

IT-MikeS:
addresses WN-362


jcesarmobile (changes requested):
You also have to bump the version on the .podspec file
i.e. https://github.com/ionic-team/capacitor-plugins/blob/main/action-sheet/CapacitorActionSheet.podspec#L14

jcesarmobile (approved):

giralte-ionic (approved):

PR #880, opened 2022-03-18
Make camera prompt cancelable

LaMbA3:
Added ability to make camera prompt cancelable. When set to true, clicking on overlay will dismiss prompt.

### DIFF
--- a/camera/src/definitions.ts
+++ b/camera/src/definitions.ts
@@ -173,6 +173,14 @@ export interface ImageOptions {
    * @since 1.0.0
    */
   promptLabelPicture?: string;
+
+  /**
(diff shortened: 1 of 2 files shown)


jcesarmobile:
Thanks for the PR, but the prompt is not cancelable on purpose.
On iOS and Android it's cancelable, but the plugin will throw an error if canceled.
On web the action sheet used by the plugin doesn't fire any events when canceled, so the plugin won't resolve nor reject if the prompt is canceled, so, to prevent that we made it non cancelable.
When pwa-elements gets updated to emit an event when canceled we would consider changing the behavior of the prompt to match iOS/Android behavior.

PR #885, opened 2022-03-21
refactor(geolocation): replace getCurrentPosition logic

theproducer:
closes: https://github.com/ionic-team/capacitor-plugins/issues/571


jcesarmobile (changes requested):
This should be done for `getCurrentPosition` only, not for `watchPosition`. So the timeout should be documented that it doesn't work for `getCurrentPosition` instead of removing the option as it would still be used for `watchPosition` and will continue working on web and iOS.

I sent a pull request in the past that was not merged because we fixed the problem in a non breaking way, you can use it for reference.

https://github.com/ionic-team/capacitor-plugins/pull/582

theproducer:
> This should be done for `getCurrentPosition` only, not for `watchPosition`. So the timeout should be documented that it doesn't work for `getCurrentPosition` instead of removing the option as it would still be used for `watchPosition` and will continue working on web and iOS.
> 
> I sent a pull request in the past that was not merged because we fixed the problem in a non breaking way, you can use it for reference.
> 
> #582

Ah, got it.  Thanks!

jcesarmobile:
getCurrentPosition calls sendLocation in the implementation class, so sendLocation should contain the new code and requestLocationUpdates should only contain the watchPosition code, so you can remove the getCurrentPosition boolean parameter from it

jcesarmobile (changes requested):
the code is crashing because of `fusedLocationClient` being null, it needs to be initialized like this before it's used `fusedLocationClient = LocationServices.getFusedLocationProviderClient(context);`

jcesarmobile (approved):

piotr-cz:
How about rejecting promise manually after it times out with race condition like [here](https://stackoverflow.com/questions/67924991/navigator-geolocation-getcurrentposition-hangs-on-ios-pwa/67924992#67924992) ?

PR #886, opened 2022-03-21
refactor(status-bar): replace deprecated code 

### LINKED ISSUE #847: deprecation warnings on status-bar plugin
### Plugin
status-bar

### Description
getSystemUiVisibility() in View has been deprecated
SYSTEM_UI_FLAG_LIGHT_STATUS_BAR in View has been deprecated
SYSTEM_UI_FLAG_LIGHT_STATUS_BAR in View has been deprecated
FLAG_TRANSLUCENT_STATUS in LayoutParams has been deprecated
SYSTEM_UI_FLAG_FULLSCREEN in View has been deprecated
SYSTEM_UI_FLAG_VISIBLE in View has been deprecated
SYSTEM_UI_FLAG_LAYOUT_STABLE in View has been deprecated   
SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN in View has been deprecated

### Platform(s)
android

### Preferred Solution
remove or silence deprecation warnings

IT-MikeS:
closes #847


giralte-ionic (approved):

status-bar/android/build.gradle
```diff
@@ -51,6 +51,7 @@ repositories {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation project(':capacitor-android')
+    implementation "androidx.core:core:1.7.0"
```
jcesarmobile, line 54:
```suggestion
    implementation "androidx.core:core:$androidxCoreVersion"
```
use the androidxCoreVersion variable instead of hardcoded 1.7.0

don't need to be documented since it's one of the default variables included in variables.gradle, but need to be declared in the ext section on top of the file

jcesarmobile (changes requested):
Using a boolean for knowing if it's overlaid or not is not very reliable as other plugins could change the status bar appearance and put the variable out of sync.
If it's not possible to know the real value anymore we should keep the old API where it still works and only use the boolean value for where it no longer works and document that it might not be accurate on those versions (WindowCompat docs do something similar with some functions)

Also there are a few of the new functions that Android Studio is warning that "may produce NullPointerException", I added a comment for one of them, but there are more and should be handled.

giralte-ionic (approved):

IT-MikeS:
> I think whenever the `windowInsetsControllerCompat` is null we should log some warning, not sure how common that can be, but at least give some insight to the users if they notice while testing.
> Other than that looks good and works fine.

Its very unlikely as far as I understand however nothing wrong with adding warnings. 👍

jcesarmobile (changes requested):
Use Capacitor's `Logger.warn` instead of default Log class so the logs can be disabled based on the logger preferences.

PR #906, opened 2022-03-30
docs(keyboard): update TS config example w/ enum

eric-horodyski:
Update the Keyboard plugin documentation to use the `KeyboardResize` enumeration for the TypeScript Capacitor configuration example.


jcesarmobile:
Thanks for the PR, the docs are generated from the types and they don't support enums at the moment.
So we should fix it in the doc generator. 
It's being tracked internally, so I'll close the PR.

PR #907, opened 2022-03-30
chore: remove all unused conditional code from plugins

IT-MikeS:
Addresses: WN-347


jcesarmobile (changes requested):
The `@TargetApi(Build.VERSION_CODES.LOLLIPOP_MR1)` was added to specify that the function is only used on LOLLIPOP_MR1, so should not be removed.

just noticed that the code is also used on Android M, so the targetApi is not right, not sure if it should use M instead

jcesarmobile (approved):

PR #919, opened 2022-04-08
feat(push-notifications): Allow to show while in foreground

jcesarmobile:
use the existing presentationOptions object used on iOS to allow to show push notifications while in foreground
if present it will create a channel for displaying push notifications in foreground
if it contains "alert" it will display the push notification
if it contains "sound" it will create the channel with the default sound.
badge is not implemented because the plugin doesn't support badges at the moment for Android.


IT-MikeS (approved):

Lennix:
Maybe add a note that it only works from 26+ (I believe that's Android O)

PR #924, opened 2022-04-13
refactor(push-notifications): Replace deprecated FirebaseInstanceId

vially:
This is a backport of #854 for Capacitor 3.

This is necessary because:
- Capacitor 4 is not yet released
- `com.google.firebase:firebase-messaging:21.0.1` is not compatible with Android 12
- bumping `firebase-messaging` to a newer version (e.g.: using `firebaseMessagingVersion` in `build.gradle`) does not work because version `22.0.0+` dropped the `firebase-iid` dependency ([release notes](https://firebase.google.com/support/release-notes/android#messaging_v22-0-0))

This means that without this change push notifications do not work in Capacitor 3 projects on Android 12 devices.

### DIFF
--- a/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
+++ b/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
@@ -8,6 +8,7 @@ import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.service.notification.StatusBarNotification;
+import androidx.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


brendan-holly-modea:
This is great. We also need a fix for this IID issue.

hoi4:
Would it be possible to release this, @jcesarmobile 🙏🏻

jcesarmobile:
This has been fixed in 4.0.0 beta versions of the plugins and won't be backported to the v1 version of the plugin as it's a breaking change.

The final version will be released this month, but you can start migrating to the beta version, it's pretty much finished and won't change much
Here is the upgrade guide:
https://next.capacitorjs.com/docs/updating/4-0

PR #926, opened 2022-04-14
Capacitor 4

THANAYHORN:


jcesarmobile:
We will merge capacitor-4 branch into main once we release the first beta, closing since we will use our own branch and it will include more changes

PR #927, opened 2022-04-19
docs(push-notifications): Update iOS code & add link to setup guide

wsamoht:
The code in the [Push Notification guide](https://capacitorjs.com/docs/guides/push-notifications-firebase) was recently updated in this [commit](https://github.com/ionic-team/capacitor-site/commit/f46f05d0f5aa5da39f8545364de8addf23f36eb3) clarifying the FCM token is needed in Firebase for sending push notifications to iOS and not the APN token. To reduce confusion, I've updated the code here as well.

I've also added a link to the more detailed guide of setting up push notifications. There seems to be quite a few questions on the Ionic Forum about setup. The most recent is [here](https://forum.ionicframework.com/t/ios-push-notifications-not-being-received-after-capacitor-2-3-and-ionic-5-6-migration/222497) which prompted this PR.

### DIFF
--- a/push-notifications/README.md
+++ b/push-notifications/README.md
@@ -2,6 +2,8 @@
 
 The Push Notifications API provides access to native push notifications.
 
+For a detailed setup guide, see [Push Notifications - Firebase](https://capacitorjs.com/docs/guides/push-notifications-firebase).
+
 ## Install
 
 ```bash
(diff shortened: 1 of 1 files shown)


jcesarmobile:
That guide is for using push notifications with firebase, which is not mandatory to use on iOS but some people wanted to, so the guide was created to document how it could be done without the need of using an additional plugin.

But the default is APNS and that's what the push-notifications plugin documents.

wsamoht:
That makes sense. Would you be open to adding a Firebase section to the iOS setup that mentions the guide? I feel like most developers reaching for push notifications are going to use Firebase for both Android and iOS so there is just one service that they are using. Right now, they have to go searching and might find the guide or end up posting on the forums.

I am thinking something like:

## Firebase setup
By default, the plugin is configured for you to use [Apple's Push Notification service (APNs)](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server). If you would like to use Firebase to send push notifications on iOS, please see the guide [here](https://capacitorjs.com/docs/guides/push-notifications-firebase#ios).

Issue #899, opened 2022-03-28
bug(filesystem): subsequent appendFile calls error on web


jcesarmobile:
### Plugin(s)
filesystem

### Capacitor Version
3.x

### Platform(s)
web

### Current Behavior
calling appendFile multiple times with base64 ends up in broken file

### Expected Behavior
to append correctly as web and android do 

### Code Reproduction
https://github.com/jcesarmobile/issue-5541/tree/append-bug
run 
```
npm install
npm start
```
see First and Next
refresh the browser and you'll get a white screen and an error on the console

jcesarmobile:
this has been fixed in 4.0.0 beta 0

PR #928, opened 2022-04-20
fix(filesystem): web appendFile with base64 data

IT-MikeS:
fixes: #899

### DIFF
--- a/filesystem/src/web.ts
+++ b/filesystem/src/web.ts
@@ -242,7 +242,9 @@ export class FilesystemWeb extends WebPlugin implements FilesystemPlugin {
     if (occupiedEntry !== undefined) {
       if (occupiedEntry.content && this.isBase64String(occupiedEntry.content)) {
         if (!this.isBase64String(data))
-          throw Error('Supplied data does not match the current file data encoding (base64)');
+          throw Error(
+            'Supplied data does not match the current file data encoding (base64)',
+          );
         data = btoa(atob(occupiedEntry.content) + atob(data));
       } else {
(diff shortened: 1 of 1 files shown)


jcesarmobile (changes requested):
I think the code should behave more like on iOS/Android where instead of checking if the existing content is valid base64, it checks if the encoding option is passed. 
Then, if no encoding is passed it means it should be base64, so should validate that it's true (for writeFile too).

At the moment we don't validate and allow to write without encoding on web while on iOS/Android it fails. If we validate before writing we don't need to validate the existing value.

filesystem/src/web.ts
```diff
@@ -240,7 +247,13 @@ export class FilesystemWeb extends WebPlugin implements FilesystemPlugin {
     }
 
     if (occupiedEntry !== undefined) {
-      data = occupiedEntry.content + data;
+      if (occupiedEntry.content && !encoding) {
```
jcesarmobile, line 250: Is it possible that occupiedEntry.content is undefined?

If it's not possible then we shouldn't check

If it's possible then there could be a case where there is no content and no encoding and it goes to the else and allows to write base64 without checking if it's base64

jcesarmobile, line 250: Is it possible that occupiedEntry.content is undefined?

If it's not possible then we shouldn't check

If it's possible then there could be a case where there is no content and no encoding and it goes to the else and allows to write base64 without checking if it's base64

jcesarmobile (approved):

Issue #804, opened 2022-02-07
feat(device): exact model name/identifier


robingenz:
### Plugin

- [x] `@capacitor/device`

### Description

I would like to have access to the exact model name/identifier (e.g. `iPhone14.3` for the iPhone 13 Pro Max, see https://www.theiphonewiki.com/wiki/Models#iPhone).

### Platform(s)

- [x] Android
- [x] iOS

### Preferred Solution

An additional property in the [DeviceInfo](https://capacitorjs.com/docs/apis/device#deviceinfo) interface that contains this identifier.

### Additional Context

See Cordova Device Plugin `device.model` property.

mcfarljw:
I am interested in this as well. We were using the model data from Cordova to target several specific devices.

brandudno:
We are also interested in this feature, we need to know the specific iOS model when debugging issues on devices.

phillip-whitaker-hmcts:
Also interested - can this go in?

PR #929, opened 2022-04-21
feat(device): model now reflects exact model on iOS

IT-MikeS:
closes: #804

This is a feature for the plugin when released along side Capacitor 4 as it's a breaking change.

This work is derived from the original work done by @robingenz on their original PR #809


jcesarmobile (changes requested):
For the simulator it's returning arm64 (on M1 macs), I think the model check should see if the device is a simulator with the `#if targetEnvironment(simulator)` macro and return the old `UIDevice.current.model` instead  or maybe `UIDevice.current.model + " Simulator"`, despite it's not gives the real mode I think it's better to get "iPhone" or "iPhone Simulator" than "arm64".

IT-MikeS:
Good call, I like the idea of falling back to `UIDevice.current.model + " Simulator"` on a simulator, makes it nice and clear.

jcesarmobile (approved):

device/ios/Plugin/DevicePlugin.swift
```diff
@@ -24,6 +24,7 @@ public class DevicePlugin: CAPPlugin {
         let diskFree = implementation.getFreeDiskSize() ?? 0
         let realDiskFree = implementation.getRealFreeDiskSize() ?? 0
         let diskTotal = implementation.getTotalDiskSize() ?? 0
+        let modelName = isSimulator ? ProcessInfo().environment["SIMULATOR_MODEL_IDENTIFIER"]! + " Simulator" : implementation.getModelName()
```
jcesarmobile, line 27: Don’t force unwrap. I thought we had a lint rule to prevent it, but looks like we don’t.
Also, now that we can get the actual name I don’t think the model should contain the “Simulator” part, if users need to know if it’s a simulator they can check the “isVirtual” property

giralte-ionic (approved):

jcesarmobile (changes requested):
The code produces two warnings
1. Will never be executed (line 27)
2. Expression implicitly coerced from 'String?' to 'Any' (line 36)
They should be fixed

jcesarmobile (approved):

PR #959, opened 2022-05-09
chore(ci): change ci to use xcode 13

IT-MikeS:


jcesarmobile (changes requested):
There are a few references to `macos-latest`, I think we should use `macos-12` on everything

jcesarmobile (approved):

Issue #953, opened 2022-05-03
Capacitor-Google Maps issue on Android


notseih:
### Plugin(s)

Capacitor - Google Maps

### Capacitor Version

```
   Capacitor Doctor   

Latest Dependencies:

  @capacitor/cli: 3.5.0
  @capacitor/core: 3.5.0
  @capacitor/android: 3.5.0
  @capacitor/ios: 3.5.0

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 3.4.3
  @capacitor/android: 3.4.3
  @capacitor/core: 3.4.3

[success] Android looking great! 👌
```

### Platform(s)

Android

### Current Behavior

Everything is working as intended on the web view when I launch the web server but on Android I’m having issues. The sidemenu opens like always, but im unable to click on the elements that overlap the map. Same if I try to put an ion-button on top of the map, is shows but its unclikable. If I open the sidemenu, and drag the click on top of the menu entries, i can see the map moving behind.
Here u can see how it works on the web view:
![webview](https://user-images.githubusercontent.com/104075920/166591497-55c85fba-fa18-4200-9899-8f6e74611bd1.gif)

And here on Android:
![android](https://user-images.githubusercontent.com/104075920/166595002-bf0de29a-cb89-4809-9b31-9cd05146c6c4.gif)

### Expected Behavior

 It should work same as on the web view, i can click the button, open the sidemenu and click on every item without the map blocking it.

### Code Reproduction

https://github.com/notseih/pruebaMap

### Other Technical Details

If anyone is going to try it with the repo, remember to put your google maps API Key on the enviroments/enviroment.ts and also in the androidmanifest.xml

KevinToala:
same issue

notseih:
Has anyone figured what's happening or any way to fix it?

RiaanLaubscher:
I can't even see the map on android for ionic react. I know the component is there and probably hidden underneath the webview but I cannot figure out how to show it. I have added `background: none` to literally every single parent component of the `capacitor-google-map` component, but still can't see it.

notseih:
> I can't even see the map on android for ionic react. I know the component is there and probably hidden underneath the webview but I cannot figure out how to show it. I have added `background: none` to literally every single parent component of the `capacitor-google-map` component, but still can't see it.

At least mine is showing... Im sure you tried this but just in case is your ion-content on the .scss looking like this? `ion-content{
    --background: transparent;
}`

[...]

PR #966, opened 2022-05-11
fix(google-maps): Fixing map touch events overriding html elements on Android

theproducer:
This PR ignores map touch events if the map is not the top most layer.

closes: https://ionic-cloud.atlassian.net/browse/WN-436
closes: https://github.com/ionic-team/capacitor-plugins/issues/953

### DIFF
--- a/google-maps/README.md
+++ b/google-maps/README.md
@@ -601,13 +601,14 @@ setOnMyLocationClickListener(callback?: MapListenerCallback | undefined) => Prom
 
 #### CreateMapArgs
 
-| Prop              | Type                                                        |
-| ----------------- | ----------------------------------------------------------- |
-| **`id`**          | <code>string</code>                                         |
-| **`apiKey`**      | <code>string</code>                                         |
-| **`config`**      | <code><a href="#googlemapconfig">GoogleMapConfig</a></code> |
-| **`element`**     | <code>HTMLElement</code>                                    |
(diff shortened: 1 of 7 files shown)


IT-MikeS (approved):

tafelnl:
At least give some credits to the community man. This is not a solution you thought of yourself.

ItsChaceD (approved):

thomasvidas:
Hi @tafelnl! Given the scope the `@capcitor-community` and `@capacitor` plugins both have, wrapping around the native Google Maps SDK, there's bound to be overlap. I even gave the community plugin a shout out on our Google Maps livestream today! I'm really happy that it exists as an alternative. It's a prime example of a great plugin.

However, Ionic has been trying to get a Google Maps plugin working since the Capacitor v1 alpha days as an official plugin, but we've never had the bandwidth until we've scaled the team up recently. Because there's really only one good way that you can embed a native element in a web context, and the fact that the only interface to connect web/native is via the Capacitor plugin API, the code (and solutions) _will_ end up similar. The team is not intentionally copying the community plugin's code or not giving credit. We even reached out to make sure that you didn't feel like you were being trampled on.

I _want_ the community plug-in to exist as an option, it is better at some things and ours is better at others. I'll continue boosting up the community plug-in as an alternative to the official one. But, the community plug-in existing would never have prevented an official one from also existing.

Issue #918, opened 2022-04-07
Keyboard plugin TS declaration file update


saradevh:
Hello,

I was wondering, shouldn't [KeyboardStyle](https://github.com/ionic-team/capacitor-plugins/blob/8e8783622d6e77c38c4aa741a622455302f30486/keyboard/src/definitions.ts#L67) be there instead of hardcoded values? KeyboardStyle is declared in the same file.
`style?: KeyboardStyle`

https://github.com/ionic-team/capacitor-plugins/blob/8e8783622d6e77c38c4aa741a622455302f30486/keyboard/src/definitions.ts#L31

Thanks!

jcesarmobile:
KeyboardStyle values are uppercase, so not really sure if they will work since the config expects lowercase values, also KeyboardStyle.Default  is not a valid value for the config

it could be done by changing the config native code to support Default value and uppercase values

jcesarmobile:
this has been fixed on 4.0.0 beta 0

PR #969, opened 2022-05-12
chore(keyboard): Using KeyboardStyle in plugin TS declaration

theproducer:
closes: https://github.com/ionic-team/capacitor-plugins/issues/918

### DIFF
--- a/keyboard/src/definitions.ts
+++ b/keyboard/src/definitions.ts
@@ -26,7 +26,7 @@ declare module '@capacitor/cli' {
        * Only available on iOS.
        *
        * @since 1.0.0
-       * @example "dark"
+       * @example "DARK"
        */
       style?: KeyboardStyle;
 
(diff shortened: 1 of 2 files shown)


jcesarmobile (changes requested):
Change the example to "DARK" (uppercase) since KeyboardStyles are uppercase
Then run `npm run build` so the docs get generated.

We are going to make the config manual in this PR https://github.com/ionic-team/capacitor-plugins/pull/961, but would be better if we get your PR first so you don't have to manually change the README.

keyboard/src/definitions.ts
```diff
@@ -28,7 +28,7 @@ declare module '@capacitor/cli' {
        * @since 1.0.0
        * @example "dark"
```
jcesarmobile, line 29:
```suggestion
       * @example "DARK"
```

jcesarmobile (approved):

PR #976, opened 2022-05-16
chore(google-maps): target & compile sdk set to 32

IT-MikeS:


jcesarmobile (changes requested):
can you do two separate PRs? one for merging main into capacitor-4 and another for the actual change

PR #991, opened 2022-05-18
chore(google-maps): Add missing swiftlint entry

jcesarmobile:


IT-MikeS:
We have a config js file already for swiftlint:

https://github.com/ionic-team/capacitor-plugins/blob/main/google-maps/swiftlint.config.js

Issue #561, opened 2021-08-23
Migrate to Splash Screen API


jcesarmobile:
### Plugin
splash-screen

### Description
When targeting Android SDK 31, a new Splash Screen API can be used

https://developer.android.com/about/versions/12/splash-screen-migration?hl=es

Some features might not be possible with that API.

### Platform(s)
android

mcfarljw:
I just got a new device with Android 12 (Samsung Galaxy S22) and thought I was going crazy when seeing two splash screens. Does anybody know of any quick and relatively painless ways of disabling the Ionic Capacitor splash screen for SDK 31+ devices while this plugin gets updated?

maxcodefaster:
There seems to be a workaround, at least for the cordova plugin: https://github.com/apache/cordova-plugin-splashscreen/issues/311#issuecomment-938598941

mster81:
Full disclosure, I don't know if this is the correct solution here but doing some testing and adding this to the 'styles.xml' 

<style name="AppTheme.NoActionBarLaunch" parent="AppTheme.NoActionBar">
        <item name="android:background">@drawable/splash</item>
        <item name="android:windowIsTranslucent">false</item>
        <item name="android:colorBackground">#ff0000</item>
    </style>

Allows me to control the first splash, from a background color that is. If you don't want to have the first splash screen ( one where the icon shows up ) then you can set windowistranslucent to true. 

If its an acceptable approach, then baking it into the plugin might be nice to 

1) Have the option to 'show' icon splash 
2) Set icon splash bg-color 

If this wildly incorrect and should not be done this way I apologies :D

bhandaribhumin:
For capacitor project this is the fix :)

![Screenshot 2022-04-29 at 8 05 38 PM](https://user-images.githubusercontent.com/3336047/165966619-4d22853a-da7d-42cb-982c-190040e76bf7.png)

Issue #459, opened 2021-06-03
↕️ Splash Screen image is stretched.


opuu:
### Plugin(s)
Splash Screen

### Platform(s)
Android

### Current Behavior
Image is stretched when gesture based navigation bar is enabled.

Adding this line to config helps but still it is stretched at first 
     `"androidScaleType": "CENTER_CROP",`

![Screenshot_20210604-015059_oone](https://user-images.githubusercontent.com/77793409/120704341-4ab09a80-c4d8-11eb-88ee-267509847d35.png)

### Expected Behavior
Show image without stretching.

DwieDima:
i was able to fix this issue by providing paramters to `capacitor.config.json`:

```

		"SplashScreen": {
			"showSpinner": false
			"launchAutoHide": false,
			"androidScaleType": "CENTER_CROP",
			"showSpinner": false,
			"splashFullScreen": false,
			"splashImmersive": false
		}
```

also had to replace style `NoActionBarLaunch` inside `android/app/src/main/res/values/styles.xml`

```

  <style name="AppTheme.NoActionBarLaunch" parent="AppTheme.NoActionBar">
    <item name="android:background">@drawable/splash</item>
    <item name="android:windowNoTitle">false</item>
    <item name="android:windowActionBar">false</item>
    <item name="android:windowContentOverlay">@null</item>
    <item name="android:windowFullscreen">false</item>
  </style>
```

this solution should be seen more as a workaround, since the splashscreen is not fullscreen and a black bar is visible at the top.
I would appreciate an official fix.

iamart:
It doesn't seem to affect Pixel_3a_API_31_arm64-v8a. Not sure if that's the aspect ratio or API level comes to play.

What I found is, removing `android:theme="@style/AppTheme.NoActionBarLaunch"` from AndroidManifest.xml fixes the issue.

PR #1011, opened 2022-05-31
feat(android): Android 12 Splash Screen API

ItsChaceD:
This PR adds support for the optional usage of the [Android 12 Splash Screen API](https://developer.android.com/guide/topics/ui/splash-screen).

- This will affect all devices running SDK Version 31+ (required)
- Supports Fade Out Animation and Animated Icons
- Backwards compatible with SDK < 31 if opted-in to using new API.

closes https://github.com/ionic-team/capacitor-plugins/issues/459
closes https://github.com/ionic-team/capacitor-plugins/issues/561


splash-screen/src/definitions.ts
```diff
@@ -153,6 +153,23 @@ declare module '@capacitor/cli' {
...
+       * This will override most other configuration options.
+       * Instead, it will use Android's recommended splash screen settings.
+       *
+       * This will only affect Android SDK versions < 31.
+       * There is currently no way to disable the new API in SDK versions >= 31.
+       *
+       * NOTE: You must set the splash screen styles in the `res/values/styles.xml`
+       * For more information, refer to the [Capacitor Splash Screen #Android Documentation](https://capacitorjs.com/docs/apis/splash-screen#android)
+       *
+       * @since 1.3.0
```
jcesarmobile, line 169:
```suggestion
       * @since 4.0.0
```
all plugins are going to be 4.0.0 to match Capacitor 4

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
```diff
@@ -45,12 +54,24 @@ public class SplashScreen {
...
         settings.setShowDuration(config.getLaunchShowDuration());
         settings.setAutoHide(config.isLaunchAutoHide());
+
+        // Method can fail if styles are incorrectly set...
+        // If it fails, log error & fallback to old method
+        try {
+            showWithAndroid12API(activity, settings);
+            return;
+        } catch (Exception e) {
+            e.printStackTrace();
```
jcesarmobile, line 67:
```suggestion
```

Don't print the whole stack trace, it's not very helpful for the users, the warning should be enough.

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
```diff
@@ -10,13 +12,20 @@
 import android.graphics.drawable.Animatable;
 import android.graphics.drawable.Drawable;
 import android.graphics.drawable.LayerDrawable;
+import android.os.Build;
```
jcesarmobile, line 15:
```suggestion
```
It's unused, can be removed

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
```diff
@@ -10,13 +12,20 @@
...
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.ViewTreeObserver;
+import android.view.WindowManager;
 import android.view.animation.LinearInterpolator;
 import android.widget.FrameLayout;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.ProgressBar;
+import androidx.annotation.RequiresApi;
```
jcesarmobile, line 28:
```suggestion
```
It's unused, can be removed.

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
```diff
@@ -59,6 +80,86 @@ public void showOnLaunch(final AppCompatActivity activity) {
...
+                                    new Handler(context.getMainLooper())
+                                        .postDelayed(
+                                            () -> {
+                                                // Splash screen is done... start drawing content.
+                                                isVisible = false;
+                                                content.getViewTreeObserver().removeOnPreDrawListener(this);
+                                            },
+                                            settings.getShowDuration()
+                                        );
+                                }
```
jcesarmobile, line 152:
```suggestion
                                if (settings.isAutoHide()) {
                                    // Start Timer On First Run
                                    if (!isVisible && !isHiding) {
                                        isVisible = true;
                                        new Handler(context.getMainLooper())
                                            .postDelayed(
                                                () -> {
                                                    // Splash screen is done... start drawing content.
                                                    isVisible = false;
                                                     content.getViewTreeObserver().removeOnPreDrawListener(this);
                                                },
                                                settings.getShowDuration()
                                            );
                                        }
                                    }                                
```
isAutoHide means the splash will automatically hide after the getShowDuration time, with current code it hides right away without waiting, I think the code that does the delayed hide should be inside the isAutoHide() check.

jcesarmobile (changes requested):
Added a few comments.

Also I think there is something wrong with the README, there are some style changes that shouldn't be there.
If I run `npm run build` it undoes most of the changes, can you run it too and see if it undoes the changes?

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
```diff
@@ -59,6 +80,81 @@ public void showOnLaunch(final AppCompatActivity activity) {
...
+
+                        fadeAnimator.start();
+
+                        isHiding = true;
+                        isVisible = false;
+                    }
+                );
+
+                // Set Pre Draw Listener & Delay Drawing Until Duration Elapses
+                final View content = activity.findViewById(android.R.id.content);
```
jcesarmobile, line 125: make this `content` variable global so it can be used in hide method instead of needing to pass an activity, since hide is public, adding a parameter would be a breaking change, so better avoid it.

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
```diff
@@ -45,12 +54,24 @@ public class SplashScreen {
...
+            showWithAndroid12API(activity, settings);
+            return;
+        } catch (Exception e) {
+            Logger.warn("Android 12 Splash API failed... using previous method.");
+            this.onPreDrawListener = null;
+        }
+
+        if (config.getLaunchShowDuration() == 0) {
+            return;
+        }
```
jcesarmobile, line 73: This should remain on top, otherwise if users are using `"launchShowDuration": 0`, I briefly see the default android icon (green robot) between the splash and the app.

splash-screen/README.md
```diff
@@ -155,6 +156,34 @@ To use splash screen images named something other than `splash.png`, set `androi
...
+- Add `implementation "androidx.core:core-splashscreen:$coreSplashScreenVersion"` in the dependencies section of `/android/app/build.gradle`.
+
+- In `android/app/src/main/res/values/styles.xml`, edit the theme `parent` attribute on the Applications's MainActivity Theme to `Theme.SplashScreen` and add desired options to the theme.
+
+```xml
+<style name="AppTheme.NoActionBarLaunch" parent="Theme.SplashScreen">
+    <item name="windowSplashScreenAnimatedIcon">@drawable/splash</item>
+    <item name="windowSplashScreenBackground">@android:color/white</item>
+<!--  !!! OPTIONAL BRANDING IMAGE !!! -->
+<!--  <item name="windowSplashScreenBrandingImage">@drawable/NAME</item> -->
```
jcesarmobile, line 174:
```suggestion
    <item name="android:background">@drawable/splash</item>
```
I've been testing multiple things and for me the best experience (to be closer to the old behaviour) is to keep the `android:background` and not use the new properties.
Using this, on Android < 12 the behaviour is almost the same as it was before and using the programmatic show will show a very close splash as the initial one.
`android:background` won't work on Android 12+, but by default it will show the round icon there and then the programmatic show will show the old splash. We should document that.

jcesarmobile (changes requested):
Added a few comments.

Also, all the configuration options that won't work on the new splash should include a note that says something like "Does't work in Android 12+ devices or if using Android 12 Splash Screen API" or similar messaging.

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
```diff
@@ -503,9 +503,10 @@ private void hide(final AppCompatActivity activity, final int fadeOutDuration, b
         // Hide with Android 12 API
         if (null != this.onPreDrawListener) {
             this.isVisible = false;
-            final View content = activity.findViewById(android.R.id.content);
-            content.getViewTreeObserver().removeOnPreDrawListener(this.onPreDrawListener);
             this.onPreDrawListener = null;
+            if (null != content) {
+                content.getViewTreeObserver().removeOnPreDrawListener(this.onPreDrawListener);
```
jcesarmobile, line 508: you have moved this line after the `this.onPreDrawListener = null;`, so as `this.onPreDrawListener` is null, `removeOnPreDrawListener` is not having any effect.

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
```diff
@@ -571,6 +572,16 @@ private void hideDialog(final AppCompatActivity activity, boolean isLaunchSplash
             return;
         }
 
+        // Hide with Android 12 API
+        if (null != this.onPreDrawListener) {
+            this.isVisible = false;
+            this.onPreDrawListener = null;
```
jcesarmobile, line 578: same as the other comment, `this.onPreDrawListener = null;` should be after `removeOnPreDrawListener` or won't have any effect.

jcesarmobile (changes requested):

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
ItsChaceD: Good catch 🤦‍♂️

jcesarmobile (changes requested):
When using this configuration:
```
"plugins": {
    "SplashScreen": {
	  "launchAutoHide": true,
	  "launchShowDuration": 7000
    }
}
```
Then when I call Show with auto hide or Show and then Hide after a timeout, the splash never disappears.

Adding `onPreDrawListener = null;` on line 142 seems to fix the issue, but there might be another way.

jcesarmobile (approved):

IT-MikeS (approved):

moradopablo:
Hi, I have a question: to avoid the stretched image in Android 11 (https://github.com/ionic-team/capacitor-plugins/issues/459), the only solution would be to modify the style.xml file in the Android project. Thank you very much in advance.

PR #1012, opened 2022-05-31
chore(android): deprecation warnings for splash screen

ItsChaceD:
This PR fixes the deprecation warnings on Android for the Splash Screen plugin.


giralte-ionic (approved):

IT-MikeS (approved):

jcesarmobile (changes requested):
Adding `"splashFullScreen": true` or `"splashImmersive": true` in the config file make apps crash on Android 12 when calling Show(); (maybe other versions too).

It's because it should be called on the UI thread.
Should be put inside 
```
activity.runOnUiThread(
            () -> {
            });
```
There is no activity on `buildViews` method, but there is line like this `Activity activity = (Activity) context;` that can be moved before of the if to use it.

Then `"splashImmersive": true` will still crash because the `splashImage` view is not on the window yet, the whole `if (config.isImmersive())` block should be moved after `windowManager.addView(splashImage, params);`.

In older Android versions it doesn't crash, but doesn't work neither, instead of removing the deprecated code, create new `legacyImmersive` and `legacyFullScreen` functions and use those on the Android versions where it wasn't deprecated. 

Same for `setLayerType`, keep the `setDrawingCacheEnabled` for the Android versions where it wasn't deprecated in a new function and use `setLayerType` where it's deprecated.

`"splashFullScreen": true` should hide the status bar
`"splashImmersive": true` should hide the status bar and the navigation buttons (test on a device or simulator that have software buttons for navigation)

jcesarmobile (changes requested):
Added a few comments.
Also, when using `"splashFullScreen": true`, on Android 11+ the status bar is not hidden and it should.

giralte-ionic (approved):

jcesarmobile (approved):

IT-MikeS (approved):

PR #1021, opened 2022-06-02
chore: Rename Storage plugin to Preferences

jcesarmobile:
Didn't rename the default storage group, which is CapacitorStorage to avoid data loss since there was already data loss from cap 2 to cap 3 and we have migrate functions to migrate cap 2 to cap 3 data and having different migrate functions will be messy.


IT-MikeS (approved):
LTGM, however on the point of not changing the `group`, could we not automatically update existing data on construction of the plugin and just convert all the prefixed keys?

jcesarmobile:
On android the group is not a prefix but the preferences name, I don't think you can change that name without doing full data migration and the main point of doing the name change is to make it clearer to people that the plugin shouldn't be used to store big amounts of data, because they are doing it at the moment, doing the data migration ourselves on load could lead to some apps taking to long to launch until the migration is done.
Also, I don't think you can just change the prefix on iOS or web without actually migrating the data since the prefix is part of the key, so to do oldPrefix.userkey1 to newPrefix.userKey1 you have to create the newPrefix.userKey1 with oldPrefix.userkey1 and remove oldPrefix.userkey1 and that would have the same problem as I mentioned for Android.

giralte-ionic (approved):

PR #1034, opened 2022-06-10
Upgrade PushNotificationsPlugin to work with firebase-messaging v23+

bb441db:


hembachrterran:
@jcesarmobile Please check out this commit, no detailed description, but this problem exists in the latest released version.
Retrieving an FCM registration token has been changed in FCM SDK.

PR #1039, opened 2022-06-14
fix(google-maps): onClusterClickHandler not getting marker ids on web

mjwheatley:
bug: The `for` loop iterating the markers for a specified `mapId` will never return a match because the cursor representing the marker ID is being used as the mapId.
bugfix: change line 87 from `if (this.maps[id].markers[id] == marker) {` to `if (this.maps[mapId].markers[id] == marker) {`


jcesarmobile (approved):
good catch, thanks!

Issue #740, opened 2021-12-19
it is not possible to delete active push notifications individually - push-notifications


ahoyahoy:
### Plugin(s)
push-notifications

### Capacitor Version

```
Latest Dependencies:

  @capacitor/cli: 3.3.3
  @capacitor/core: 3.3.3
  @capacitor/android: 3.3.3
  @capacitor/ios: 3.3.3

Installed Dependencies:

  @capacitor/android: 3.3.3
  @capacitor/cli: 3.3.3
  @capacitor/ios: 3.3.3
  @capacitor/core: 3.3.3

[error] Xcode is not installed
```

### Platform(s)
android
ios

### The Problem
```
PushNotifications.removeDeliveredNotifications([msg])
```

This doesn't work at all.

#### Android
Android notification api expects Int message_id. But the system gets id in the form `0:1639932623809889%d78dae59d78dae59`.
This will also cause the id of **each** background message to be stored as `0`

Nothing happens on android when I call a function `removeDeliveredNotifications`
- no message is removed
- no error exception

```
native PushNotifications.removeDeliveredNotifications (#77663661)
    Object
        callbackId: "77663661"
        methodName: "removeDeliveredNotifications"
        options: Array(1)
            0:
                body: "body text"
                data: Object
                groupSummary: false
                id: 0
                title: "Test alert"    
        pluginId: "PushNotifications"

result PushNotifications.removeDeliveredNotifications (#77663661)
    undefined
```

I can't figure out how to capture an incoming background message and edit id to int

- Emulator android Pixel 5 API 28 (Android 9)
- Real android One Plus 7 Pro (Android 11)

#### ios
I don't have an apple dev environment.
Only testflight and real Iphone 8.
And there I get error `Must supply notifications to remove`

jcesarmobile:
Can you provide a sample app where the issue can be reproduced?

MicheleOrdile:
I encounter the same problem, this is a serious bug because it causes an exception in the native code
```
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
        at com.capacitorjs.plugins.pushnotifications.PushNotificationsPlugin.removeDeliveredNotifications(PushNotificationsPlugin.java:163)
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.getcapacitor.PluginHandle.invoke(PluginHandle.java:121) 
[...]

PR #1041, opened 2022-06-14
fix(push-notifications): use id and tag for canceling active notification

theproducer:
Fixes an issue where it was not possible to delete individual notifications on Android.

closes: https://github.com/ionic-team/capacitor-plugins/issues/740

### DIFF
--- a/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
+++ b/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
@@ -134,7 +134,7 @@ public class PushNotificationsPlugin extends Plugin {
                 if (o instanceof JSONObject) {
                     JSObject notif = JSObject.fromJSONObject((JSONObject) o);
                     String tag = notif.getString("tag");
-                    Integer id = notif.getInteger("id", 0);
+                    Integer id = notif.getInteger("id");
 
                     if (tag == null) {
                         notificationManager.cancel(id);
--- a/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
+++ b/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
@@ -98,6 +98,7 @@ public class PushNotificationsPlugin extends Plugin {
                 JSObject jsNotif = new JSObject();
 
                 jsNotif.put("id", notif.getId());
+                jsNotif.put("tag", notif.getTag());
 
                 Notification notification = notif.getNotification();
                 if (notification != null) {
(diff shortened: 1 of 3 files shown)


giralte-ionic (approved):

push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
```diff
@@ -128,13 +129,13 @@ public void getDeliveredNotifications(PluginCall call) {
...
 
-        List<Integer> ids = new ArrayList<>();
+        List<String> tags = new ArrayList<>();
         try {
             for (Object o : notifications.toList()) {
                 if (o instanceof JSONObject) {
                     JSObject notif = JSObject.fromJSONObject((JSONObject) o);
-                    Integer id = notif.getInteger("id");
-                    ids.add(id);
+                    String tag = notif.getString("tag");
```
jcesarmobile, line 137: RemoveDeliveredNotifications removes both push and local notifications, local notifications don't have a tag (at the moment, could be added in the future).
And on local notifications the id is not 0.
So the code should check if the tag is not null and call the `cancel(tag, id)`, and if null call `cancel(id)`.
For making things easier, you can remove the tags/ids list and the for where cancel is called and just call cancel here instead of the `tags.add(tag);`

jcesarmobile (changes requested):
Since getDeliveredNotifications now returns the notification tag, the types should be edited to reflect that, the tag should be optional and documented that it's only returned for Android push notifications (getDeliveredNotifications also returns local notifications, but they don't usually have a tag)

push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
```diff
@@ -128,13 +129,18 @@ public void getDeliveredNotifications(PluginCall call) {
...
 
-        List<Integer> ids = new ArrayList<>();
         try {
             for (Object o : notifications.toList()) {
                 if (o instanceof JSONObject) {
                     JSObject notif = JSObject.fromJSONObject((JSONObject) o);
-                    Integer id = notif.getInteger("id");
-                    ids.add(id);
+                    String tag = notif.getString("tag");
+                    Integer id = notif.getInteger("id", 0);
```
jcesarmobile, line 137: The id is not optional, so you shouldn't need the default value. 
Have you seen it being null?

jcesarmobile (changes requested):
Added a comment.

Also, you have to re run `npm run build` to update the README

push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
theproducer: not really, I removed it.

PR #1051, opened 2022-06-21
Add referrerUrl

MateiNenciu:
As title says, we need an extra parameter from the "appOpenUrl" event result. This parameter is useful for our data department for processing where our app was opened from. 
  For iOS I've made a PR (https://github.com/ionic-team/capacitor/pull/5706) for exposing the referrerUrl from userActivity.


eric-horodyski:

More context:

Issue #1035, opened 2022-06-13
bug?(ios): unexpected error thrown if location services disabled


tafelnl:
## Bug Report

I am reporting this as a bug, as it seems like unexpected behavior and is not documented anywhere. But please correct me if I am wrong (but that then probably means the docs should be updated).

### Plugin(s)
Geolcation

### Capacitor Version

```
  @capacitor/cli: 3.5.1
  @capacitor/core: 3.5.1
  @capacitor/android: 3.5.1
  @capacitor/ios: 3.5.1
```

### Platform(s)
ios

### Current Behavior
When location services are disabled, `checkPermissions` and `requestPermissions` throw the following error: `Location services are not enabled`

links to lines of code where error is thrown:
- [`checkPermissions`](https://github.com/ionic-team/capacitor-plugins/blob/7563040983ad397e28616246e7ed5ffce69727c2/geolocation/ios/Plugin/GeolocationPlugin.swift#L166)
- [`requestPermissions`](https://github.com/ionic-team/capacitor-plugins/blob/7563040983ad397e28616246e7ed5ffce69727c2/geolocation/ios/Plugin/GeolocationPlugin.swift#L194)

### Expected Behavior
- I couldn't find any documentation about this error. (Although when you dive in a bit more, it's pretty obvious why it happens). So likely many developers do not necessarily (explicitly) catch this error in their code. That can result in unexpected behavior. For example, we found out about this, after seeing many of these errors in our sentry dashboard.
- Also, on Android this error does not seem to be thrown in a similar situation (on Android the location permissions can be disabled as well).
- Besides, a similar issue (https://github.com/ionic-team/capacitor/issues/5615) was created for the Capacitor v2 plugin, but it was only labeled with `needs reproduction` and then closed after going stale.

The points above make it seem like it was indeed overlooked by and not intended behavior (again, correct me if I'm wrong).

I think it would make sense to either:
- not throw an error, but return an extra variable/enum type in the return object instead (that _could_ result in breaking behavior, so maybe it's not ideal)
- document the existence of this error (and maybe a simple example of what logical next steps could be in ones code)

### Code Reproduction

- Run a basic Capacitor app with Geolocation plugin installed on a real iOS device
- Disable location services
- Call either `checkPermissions` or `requestPermissions`

Basic reproduction: https://github.com/tafelnl/capacitor-reproduction-geolocation

Ionitron:
[...]

PR #1053, opened 2022-06-22
fix(geolocation): reject checkPermissions / requestPermissions if location services are disabled

theproducer:
This PR will check the Android location services and will throw an error in `checkPermissions` and `requestPermissions` if it's disabled, bringing the behavior inline with iOS.

closes: https://github.com/ionic-team/capacitor-plugins/issues/1035

### DIFF
--- a/geolocation/android/src/main/java/com/capacitorjs/plugins/geolocation/GeolocationPlugin.java
+++ b/geolocation/android/src/main/java/com/capacitorjs/plugins/geolocation/GeolocationPlugin.java
@@ -5,9 +5,7 @@ import android.content.Context;
 import android.location.Location;
 import android.location.LocationManager;
 import android.os.Build;
-
 import androidx.core.location.LocationManagerCompat;
-
 import com.getcapacitor.JSObject;
 import com.getcapacitor.PermissionState;
 import com.getcapacitor.Plugin;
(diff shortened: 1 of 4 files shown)


giralte-ionic (approved):

jcesarmobile (changes requested):
I think we should have a helper method in the implementation class that returns if location is available or not and use that instead of using the LocationManagerCompat class from the plugin class.
Will simplify the code and would be useful for users using the implementation class.

The implementation class already uses isLocationEnabled twice, so would be good if those calls could be replaced with the new helper method too.

geolocation/android/src/main/java/com/capacitorjs/plugins/geolocation/GeolocationPlugin.java
```diff
@@ -51,6 +51,26 @@ protected void handleOnResume() {
...
+        }
+    }
+
+    @Override
+    @PluginMethod
+    public void requestPermissions(PluginCall call) {
+        if (implementation.isLocationServicesEnabled()) {
+            super.requestPermissions(call);
+        } else {
+            call.reject("System location services are not enabled");
```
jcesarmobile, line 70:
```suggestion
            call.reject("Location services are not enabled");
```
Match the current iOS message

jcesarmobile (changes requested):
just a minor comment about the error message, it should match the message used on iOS.

jcesarmobile (approved):

tigohenryschultz:
Location requests are no longer working on Android 31, I wonder if this has anything to do with it:  https://github.com/ionic-team/capacitor-plugins/issues/1363

PR #1086, opened 2022-07-21
Typo fix

BAYiTUPAi:


jcesarmobile (changes requested):
thanks for the PR
README is generated from the types, so should be fixed in the types file and then run npm run build
https://github.com/ionic-team/capacitor-plugins/blob/main/camera/src/definitions.ts#L318

PR #1097, opened 2022-07-28
feat(local-notifications): support subtitle on iOS

p7g:
On iOS you can add a second line of title to the notification like this:

<a href="https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification">demonstration of iOS notification subtitle</a>

The subtitle property of UNMutableNotificationContent is documented here: https://developer.apple.com/documentation/usernotifications/unmutablenotificationcontent/1649873-subtitle

On other platforms it is just ignored. Did I miss anywhere that should be updated to handle a subtitle?


bazuka5801:
@p7g subtitle in apple docs since iOS `15.0` but capacitor uses min iOS version `13.0` try to run your code on iOS 13

eric-horodyski:

More context:

Issue #1111, opened 2022-08-04
bug: `Filesystem.rmdir()` Error: Entry does not exist.


HuangNO1:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.0.1
  @capacitor/core: 4.0.1
  @capacitor/android: 4.0.1
  @capacitor/ios: 4.0.1

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 4.0.1
  @capacitor/core: 4.0.1
  @capacitor/android: 4.0.1

[success] Android looking great! 👌
```

### Current Behavior

When I run `Filesystem.rmdir()` to delete a folder and run `Filesystem.mkdir()` to recreate a new folder. It's not worked.

Code:

```js
const IMAGE_DIR = 'stored-images';


  async saveImage(photo: Photo) {
    // Convert photo to base64 format, required by Filesystem API to save
    const base64Data = await this.readAsBase64(photo);
    const fileName = new Date().getTime() + '.jpeg';
    const savedFile = await Filesystem.writeFile({
      directory: Directory.Documents,
      path: `${IMAGE_DIR}/${fileName}`,
      data: base64Data,
    });

    console.log('saved: ', savedFile);
    return {
      path: `${IMAGE_DIR}/${fileName}`,
      fileName: fileName,
      data: `${base64Data}`,
    };
  }

  async deleteImage(file) {
    try {
      await Filesystem.deleteFile({
        directory: Directory.Documents,
        path: file.path,
      });
    } catch (err) {
      console.log(err);
    }
  }

  private async readAsBase64(photo: Photo) {
    // "hybrid" will detect Cordova or Capacitor
    if (this.platform.is('hybrid')) {
      console.log('is hybrid');
      // Read the file into base64 format
      const file = await Filesystem.readFile({
        path: photo.path,
      });
      return file.data;
    } else {
      // Fetch the photo, read as a blob, then convert to base64 format
      const response = await fetch(photo.webPath);
      const blob = await response.blob();
      return (await this.convertBlobToBase64(blob)) as string;
    }
  }

  convertBlobToBase64 = (blob: Blob) =>
    new Promise((resolve, reject) => {
      const reader = new FileReader();
      reader.onerror = reject;
      reader.onload = () => {
        resolve(reader.result);
      };
      reader.readAsDataURL(blob);
    });

// ------------err func--------------------------------------

  async removeAll() {
[...]

PR #1112, opened 2022-08-05
fix(filesystem): failing to remove folder content on rmdir

jcesarmobile:
readdir now returns an object instead of just the file name, so we need to get the file name from the object to properly delete folder content

closes https://github.com/ionic-team/capacitor-plugins/issues/1111


theproducer (approved):

IT-MikeS (approved):

HuangNO1:
Thanks!And, sorry...What should I do to update the filesystem plugin with npm?

HuangNO1:
I try to change the file's line 393 which in `./node_modules/@capacitor/filesystem/dist/plugin.js`, but it's still not worked.

PR #1118, opened 2022-08-08
fix(push-notifications): properly get the configured icon from manifest

Koenic:
When using the option to display the notification locally on android when the application is in the foreground `applicationInfo.metaData` would always be `null`. This results in it using the wrong icon.

To be honest I'm not familiar enough to know why it fails to retrieve the metaData here. However after some googling the below solution seems to work in my case.


jcesarmobile (approved):
good catch, thanks!

PR #1136, opened 2022-08-17
fix(camera): remove listener methods

jcesarmobile:
The new methods introduced on https://github.com/ionic-team/capacitor-plugins/pull/1125 will be available in next 4.1.0 release, not in 4.0.0, update docs accordingly.

Add missing `removeAllListeners` method in `CameraPlugin.m`, calling it errors at the moment.

Removed `and below` from `iOS 13 and below` error messages since we don't support older versions.

Also noticed that `photoLibraryDidChange` is called every time a change is made in the user's photo library, not only when the limited selection is changed (i.e. if the user takes a new picture or deletes it), so `limitedLibrarySelectionChanged` event is fired for every change and not for selection changes, so maybe the name should be changed to `photoLibraryDidChange` (to match the native name). Current name is misleading and also the description of it is not accurate.
Do we really need that listener? it doesn't provide any useful information to the user.

### DIFF
--- a/camera/README.md
+++ b/camera/README.md
@@ -139,7 +139,7 @@ iOS 14+ Only: Allows the user to update their limited photo library selection.
 
 **Returns:** <code>Promise&lt;<a href="#galleryphotos">GalleryPhotos</a>&gt;</code>
 
-**Since:** 4.0.0
+**Since:** 4.1.0
 
 --------------------
 
@@ -154,7 +154,7 @@ iOS 14+ Only: Return an array of photos selected from the limited photo library.
 
 **Returns:** <code>Promise&lt;<a href="#galleryphotos">GalleryPhotos</a>&gt;</code>
 
-**Since:** 4.0.0
+**Since:** 4.1.0
 
 --------------------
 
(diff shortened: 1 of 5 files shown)


theproducer:
> Also noticed that `photoLibraryDidChange` is called every time a change is made in the user's photo library, not only when the limited selection is changed (i.e. if the user takes a new picture or deletes it)

Ah, this is probably because any new photos added by the app in limited mode is itself added to the selection, thus triggering the change.

Regarding its usefulness, I'd imagine it would be good to know if that selection has changed, since it can be changed from outside the app (in Settings) as well.

jcesarmobile:
No, `photoLibraryDidChange` fires every time there is a change in the photo library according to [Apple docs ](https://developer.apple.com/documentation/photokit/phphotolibrarychangeobserver)
> The PHPhotoLibraryChangeObserver protocol notifies you of changes that occur in the photo library, regardless of whether those changes are made by your app, by a user in the Photos app, or by another app that uses the Photos framework.

So it fires every time a change is made, even if the change is not made from the app itself.
To be clear, where I said "if the user takes a new picture or deletes it", I meant takes a picture from the Camera app and deleted a picture from Photos app, not from the user app. When I go back to the app a few limitedLibrarySelectionChanged events were fired.

Also, `photoLibraryDidChange` provides information about what changed, but we don't use any of that. But I don't think it's possible to just detect if the selection was changed.

theproducer:
I see - I wouldn't be against nixing `photoLibraryDidChange` in that case.

giralte-ionic (approved):

jcesarmobile:
I've removed the listener methods after discussing with @theproducer

IT-MikeS (approved):

giralte-ionic (approved):

theproducer (approved):

PR #1165, opened 2022-09-03
Remove redundant nil coalescing

msrutek-paylocity:
Using the nil-coalescing operator (`??`) here is redundant.

If `UIApplication.shared.delegate?.window` is `nil`, then there's no need to use `??` and set it to `nil`.


jcesarmobile:
As you can see, the code doesn’t compile without it

msrutek-paylocity:
Wow, you are right.

It is actually because the expression 

```
UIApplication.shared.delegate?.window
```

is of type `UIWindow??` - a double optional.  Not at all obvious at first glance.

🤯

Okay, you can disregard this then :)

PR #1189, opened 2022-09-20
feat(local-notifications): support Android 13+ permissions

### LINKED ISSUE #1116: Android 13 (API level 33) introduces a new runtime permission for notifications
## Feature Request

Android 13 (API level 33) introduces a new [runtime permission](https://developer.android.com/guide/topics/permissions/overview#runtime) for sending [non-exempt](https://developer.android.com/about/versions/13/changes/notification-permission#exemptions) notifications from an app: [POST_NOTIFICATIONS](https://developer.android.com/reference/android/Manifest.permission#POST_NOTIFICATIONS).

### Plugin

Currently this plugin doesn't work when we set notifications because user don't see runtime permission popup on Android 13.

### Preferred Solution

- I have implemented this fix for Cordova in this repo [cordova-plugin-local-notification-12](https://github.com/bhandaribhumin/cordova-plugin-local-notification-12)

- I have added few changes to get permission popup [FIX](https://github.com/bhandaribhumin/cordova-plugin-local-notification-12/blob/master/src/android/LocalNotification.java#L214)

- The permission that you need to declare in your app's manifest file appears in the following code snippet:

`<uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>`

### Additional Context

Screenshot 2022-08-08 at 7 09 50 PM

masimplo:
This adds support for explicitly checking and requesting POST_NOTIFICATIONS permissions for Android 13 and above (SDK 33)

Tested it on Android 13 and works as expected. Android < 13 works as previously.

Resolves #1116


jcesarmobile:
Thanks for the PR, but we won't be merging it anytime soon. 
I've verified that local notifications are still working fine on Android 13 as long as the app targets SDK 32, which is the officially supported SDK on Capacitor 4.
So this won't be merged until we start working on Capacitor 5 which will target SDK 33.

IT-MikeS (approved):

theproducer (approved):

giralte-ionic (approved):

giralte-ionic:
@masimplo please update with main so we can merge please and thank you!

local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationsPlugin.java
```diff
@@ -9,21 +11,29 @@
...
 import com.getcapacitor.annotation.Permission;
+import com.getcapacitor.annotation.PermissionCallback;
 import java.util.List;
 import java.util.Map;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
-@CapacitorPlugin(name = "LocalNotifications", permissions = @Permission(strings = {}, alias = "display"))
+@TargetApi(Build.VERSION_CODES.TIRAMISU)
```
jcesarmobile, line 28:
```suggestion
```
we are going to require capacitor 5 for the plugins, which targets sdk 33, so this is not needed

local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationsPlugin.java
```diff
@@ -1,5 +1,7 @@
 package com.capacitorjs.plugins.localnotifications;
 
+import android.Manifest;
+import android.annotation.TargetApi;
```
jcesarmobile, line 4:
```suggestion
```

As commented, the TargetApi is not needed, so can be removed and this import too

local-notifications/README.md
```diff
@@ -10,6 +10,7 @@ npx cap sync
 ```
 
 ## Android
+Android 13 requires a permission check in order to send notifications.  You are required to call `checkPermissions()` and `requestPermissions()` accordingly, when targeting SDK 33.
```
jcesarmobile, line 13:
```suggestion

Android 13 requires a permission check in order to send notifications.  You are required to call `checkPermissions()` and `requestPermissions()` accordingly.

On Android 12 and older it won't show a prompt and just return as granted.
```

Added some spacing between the title and the text
Also removed the note about targeting SDK 33 as that's going to be mandatory.
And added a note about Android 12 and older not prompting since it's a common question why requesting the permission doesn't prompt.

Issue #1158, opened 2022-08-29
@capacitor/camera - pickLimitedLibraryPhotos doesn't return GalleryPhotos like API says


pedromiba:
### Plugin(s)
@capacitor/camera 4.1.0

### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.1.0
  @capacitor/core: 4.1.0
  @capacitor/android: 4.1.0
  @capacitor/ios: 4.1.0

Installed Dependencies:

  @capacitor/android: not installed
  @capacitor/cli: 4.1.0
  @capacitor/ios: 4.1.0
  @capacitor/core: 4.1.0

[success] iOS looking great! 👌
```

### Platform(s)
iOS

### Current Behavior
When the user is prompted for permissions and decides to choose only 2 photos from its library we end up with a limited permission for photos. Then we have the method `pickLimitedLibraryPhotos` to allow the user to change these photos and perhaps add new ones, but there is no way of knowing when the user actually changes them, according to the API it should return `GalleryPhotos`.

### Expected Behavior
When using `pickLimitedLibraryPhotos` it should return the GalleryPhotos just like the method `getLimitedLibraryPhotos`.

### Code Reproduction
[Repository with example of the issue](https://github.com/pedromiba/camera-example)

Look at -> PhotoGallery.tsx -> handleManage

SaroGFX:
I got the same issue, really need a fix for this

PR #1191, opened 2022-09-20
fix(camera): make pickLimitedLibraryPhotos return photos

jcesarmobile:
closes https://github.com/ionic-team/capacitor-plugins/issues/1158

Note: On iOS 14 it's not possible to know when the limited image picker is closed as the method with the completion is only available for iOS 15.
So, for being able to return the pictures, it relies on the `PHPhotoLibraryChangeObserver`. But if the users calls `pickLimitedLibraryPhotos` and then doesn't change the selection, `pickLimitedLibraryPhotos` will not reject nor resolve until some change is made on the user's photo gallery in the future.
Not sure if we want that behavior or if it would be better to just return the pictures on iOS 15+ and document that on iOS 14 it returns an empty array?

### DIFF
--- a/camera/README.md
+++ b/camera/README.md
@@ -134,6 +134,8 @@ pickLimitedLibraryPhotos() => Promise<GalleryPhotos>
 ```
 
 iOS 14+ Only: Allows the user to update their limited photo library selection.
+On iOS 15+ returns all the limited photos after the picker dismissal.
+On iOS 14 or if the user gave full access to the photos it returns an empty array.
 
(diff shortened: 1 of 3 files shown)


giralte-ionic:
I'd recommend the document method for iOS 14

giralte-ionic (approved):

jcesarmobile:
return the pictures only on iOS 15+ and document that on iOS 14 it returns an empty array?

IT-MikeS:
I agree it's better to resolve but return nothing (empty array) with documentation as to why that happens on iOS 14

theproducer (approved):
Tested on 15 and 16, works as expected.

IT-MikeS (approved):

Issue #433, opened 2020-01-13
bug: show time of when the notifications is launched and not when it was scheduled


MayKikuchiThrive:
## Describe the Bug

The problem I am having is that when I schedule a local notification for a future date, the notification will be displayed with the time the local notification was created as the timestamp, not the time the notification was scheduled.

## Platform 

- [x] Android
- [ ] iOS
- [ ] Electron
- [ ] Web

wrzr123:
Question, I'm having the same problem as described. So is it still not resolved? Using latest version of @capacitor/local-notifications, v1.1.0.
I'm using schedule at.

Installed Dependencies:

  @capacitor/cli: 3.4.0
  @capacitor/android: 3.3.4
  @capacitor/ios: 3.3.4
  @capacitor/core: 3.3.4

Splitomania:
Hi, is there any update on this? I'm using 

"@capacitor/android": "^4.0.1",
"@capacitor/app": "^4.0.1",
"@capacitor/core": "^4.0.1",
"@capacitor/local-notifications": "^4.0.1",

and it seems to be still the above mentioned behaviour.

I would really appreciate any info on this.

PR #1198, opened 2022-09-26
fix(local-notifications): Showing correct scheduled notification time

theproducer:
Fixes an issue where `on` and `every` notification schedules did not show the current notification time.

closes: https://github.com/ionic-team/capacitor-plugins/issues/433


IT-MikeS (approved):

jcesarmobile (changes requested):
You don't need to add extra parameters to the Intent. On TimedNotificationPublisher you can get the original notification including the schedule.

But I don't think that's even needed, I think calling `notification.when = System.currentTimeMillis();` in TimedNotificationPublisher should cover all cases, for `on`, `every` and `at`, so we can even remove the current code for handling the `at` and that single line should show the correct time for all of them.

jcesarmobile (approved):

Issue #1138, opened 2022-08-18
Filesystem.writeFile throws Maximum call stack size exceeded error for larger files


beonde:
### Plugin(s)
@capacitor/filesystem@4.0.1

### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.1.0
  @capacitor/core: 4.1.0
  @capacitor/android: 4.1.0
  @capacitor/ios: 4.1.0

Installed Dependencies:

  @capacitor/android: not installed
  @capacitor/ios: not installed
  @capacitor/cli: 4.1.0
  @capacitor/core: 4.1.0
```

### Platform(s)

Web

### Current Behavior

When attempting to write larger images using the Filesystem.writeFile method, the RegEx.test method used in the isBase64String method throws a maximum call stack exceeded error. I have verified that the base64 string provided is valid. The size of the source image is 3.3 MB.

### Expected Behavior

For the valid base64 string to be recognized and saved to browser storage, regardless of size.

### Code Reproduction

Let me know if you need this.

### Other Technical Details

VSCode 1.70.1 on MacOS 11.6.6
Angular 13.3.1 project with Ionic 6.1.15

I haven't added iOS or Android capacitor projects yet. First testing on web and pwa only.

### Additional Context

Relax regex validation of base64 string or eliminate the test completely as in previous builds of the plugin.

```
ERROR Error: Uncaught (in promise): RangeError: Maximum call stack size exceeded
RangeError: Maximum call stack size exceeded
    at RegExp.exec (<anonymous>)
    at RegExp.test (<anonymous>)
    at f.isBase64String (6364.5d7cc5eddf0b3c52.js:1:8220)
    at 6364.5d7cc5eddf0b3c52.js:1:2620
    at Generator.next (<anonymous>)
    at a (main.8e826124c31ed1a7.js:1:1524153)
    at R (main.8e826124c31ed1a7.js:1:1524357)
    at v.invoke (polyfills.b44f3bbb9dc50cef.js:1:7684)
    at Object.onInvoke (main.8e826124c31ed1a7.js:1:1358319)
    at v.invoke (polyfills.b44f3bbb9dc50cef.js:1:7624)
    at X (polyfills.b44f3bbb9dc50cef.js:1:17753)
    at polyfills.b44f3bbb9dc50cef.js:1:18612
    at v.invokeTask (polyfills.b44f3bbb9dc50cef.js:1:8368)
[...]

PR #1202, opened 2022-09-28
fix(filesystem): Avoid max stack size exceeded on base64 check

jcesarmobile:
Some SO comments say this check is a bit lax, but the regular expression hits max stack size exceeded and I can't find a better way that doesn't require regular expressions (tested other expressions and using match instead of test and also get max stack size exceeded).
I've been testing this check and comparing with iOS results and they both return true/false for the same string I've tested, so I think it's ok to use.
Also, we already use the btoa/atob on append call.

closes https://github.com/ionic-team/capacitor-plugins/issues/1138

### DIFF
--- a/filesystem/src/web.ts
+++ b/filesystem/src/web.ts
@@ -636,9 +636,11 @@ export class FilesystemWeb extends WebPlugin implements FilesystemPlugin {
   }
 
   private isBase64String(str: string): boolean {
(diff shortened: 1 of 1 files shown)


IT-MikeS (approved):

theproducer (approved):
Tested, works.

PR #1222, opened 2022-10-11
fix(google-maps): Cast iOS marker id to string

muuvmuuv:
Found out that `setOnMarkerClickListener` returned marker ID's are not equal to those returned by `addMarker`. The issue was with some weird JS behavior for the int returned by Swift. Casting to String on JS side did not fixed it so it was Swift.

Note: I have not tested Android in that particular case but Android seems fine since it uses `JSObject` directly.

May I ask if this is [Hacktober](https://github.com/topics/hacktoberfest) worth? :D 

closes https://github.com/ionic-team/capacitor-plugins/issues/1526


francolondra:
hey ! I detect the same problem in android, I don't have a markerId that I put before.Could you help me?

muuvmuuv:
@francolondra could you share a repl?

francolondra:
I uploaded the same problem.I put one markerId and the setOnMarkerClickListener ,return the markerId name generic
bug-marker

francolondra:
I uploaded the image , I put one markerId of my services but when I wanted to restore the markedId with setOnMarkerClickListener , it put the markerId generic
bug-marker

muuvmuuv:
Hmm I see, looks like the same issue I had, unfortunately you cannot set them yourself, this would make it a lot easier in some cases. I think its this part here. Maybe JSObject cast it wrong or converts bits wrong: https://github.com/ionic-team/capacitor-plugins/blob/main/google-maps/android/src/main/java/com/capacitorjs/plugins/googlemaps/CapacitorGoogleMap.kt#L714

Someone of the Ionic Teams needs to see this one here and take over for Android.

chrisweb:
I can confirm the bug on iOS

but I don't have the bug on android as decribed by @francolondra, for me on android markerId is always something like m1, m2 when using addMarker and also when using setOnMarkerClickListener

I can also confirm that your fix @muuvmuuv solves the problem, well done 👍 , I patched my capacitor google maps using your suggested fix and now it works well, setOnMarkerClickListener finally returns the same markerID as I had when using addMarker / addMarkers which allows me to lookup the marker in my table on get additional data

chrisweb:
@jcesarmobile as soon as you have some time, could you please review this or maybe let another dev know about the fix ;)

@francolondra this is a bit off topic, but I have a solution for your problem so I will share it here, to add data to a marker you don't need extra fields in the marker object itself, it is enough if the markers have a working ID (which currently is not the case on iOS, hence this PR), so now you get your list of markers from the database, something like [{ markerDatabaseId: 1, markerDatabaseName: 'foo' }, { markerDatabaseId: 2, markerDatabaseName: 'bar' }], then you use addMarkers (<https://capacitorjs.com/docs/apis/google-maps#addmarkers>) which returns you a list of mapMarkersIds, so now you iterate through you list from database and for each one you add the mapMarkerId (that you got from addMarkers), something like this:

```
const myMarkersFromDatabase = await fetchMarkersFromDatabase()

const addMarkersResult = await googleMap.addMarkers(googleMapMarkers)

const myFinalMarkersList: IMyCustomMarker[] = []

addMarkersResult.map((googleMapMarkerId, index) => {
	const markerData = myMarkersFromDatabase[index]
	const myMarkerWithMapId = { googleMapMarkerId, ...markerData }
	myFinalMarkersList.push(myMarkerWithMapId)
})
```

now you have an array like this: [{ markerDatabaseId: 1, markerDatabaseName: 'foo', googleMapMarkerId: 'm1' }, { markerDatabaseId: 2, markerDatabaseName: 'bar', googleMapMarkerId: 'm2' }]

and finally somewhere else in your code, if you want to get the data of a marker for example because you are using setOnMarkerClickListener (<https://capacitorjs.com/docs/apis/google-maps#setonmarkerclicklistener>), you do something like this:

```
const markerData= myFinalMarkersList.find(marker => {
    return marker.googleMapMarkerId === markerClickId
})
```

where markerClickId is the Id of the marker you get from setOnMarkerClickListener handler, that's it, you now have a marker object with all your database data, without complexifying the code of the google maps plugin for something that is a specific dev logic of one project

btw, I just noticed, the documentation has a mistake here: <https://capacitorjs.com/docs/apis/google-maps#markerclickcallbackdata>, the value that is being returned is MarkerId and not MapId, if someone can do a PR for that it would be great or else maybe I will have some time to do it soonish

francolondra:
Thanks @chrisweb for your solution but I solved this problem by filtering it for coordinates ,But I understand that is temporal right?, in some release they can solver this?

chrisweb:
@francolondra doing it via the coordinates is clever and simple idea, thx for sharing, hadn't thought about that... you could even do a hash out of both values with something like md5 if you wanted to search only by one value. As to when the fix here will be released, I have no idea, thx to @muuvmuuv we have the fix, now it is up to the Ionic team to validate it and then include it in an upcoming release, crossing fingers it will happen soon ;)

muuvmuuv:
Hm… it's still a mandatory change. I would rather not ping someone from the team…

jcesarmobile (approved):
it was missing a few other casts, I've added them
looks good, thanks!

PR #1224, opened 2022-10-12
docs(camera): fix links about PWA Element

jonz94:
The website link for PWA Element's document should be https://capacitorjs.com/docs/web/pwa-elements, not https://capacitorjs.com/docs/pwa-elements (404 Page Not Found).


jcesarmobile (approved):
good catch, thanks!

PR #1226, opened 2022-10-12
Update GoogleMapConfig

jacinlowe:
Added disableDefaultUI option to remove the google maps navigation ui.  This will allow us to pass that parameter and get a clean map to use.


jcesarmobile:
Thanks for the PR, but I've just merged https://github.com/ionic-team/capacitor-plugins/pull/1233 which allows to use any configuration from `google.maps.MapOptions`, so your change is no longer needed.

Issue #1201, opened 2022-09-27
bug(google-maps): Markers with the same icon have the same size


jcesarmobile:
### Plugin(s)
google-maps

### Capacitor Version
4

### Platform(s)
ios
android

### Current Behavior
if the same iconUrl is used for different markers, but they are configured with different sizes, they will all be displayed the same size

### Expected Behavior
every marker should be displayed with the configured size (as it works on web)

### Code Reproduction
```
const markers: Marker[] = [
        {
          coordinate: {
            lat: 47.6,
            lng: -122.33,
          },
          title: 'Title 1',
          snippet: 'Snippet 1',
          iconUrl: 'assets/icon/pin.png',
          iconSize: {
            width: 30,
            height: 30,
          },
        },
        {
          coordinate: {
            lat: 47.6,
            lng: -122.46,
          },
          title: 'Title 2',
          snippet: 'Snippet 2',
          iconUrl: 'assets/icon/pin.png',
          iconSize: {
            width: 10,
            height: 10,
          },
        },
}]
```

PR #1228, opened 2022-10-13
fix(google-maps): allow to resize icons with same image

jcesarmobile:
Save the image/bitmap without resizing and resize when the icon is going to be used with the configured size

closes https://github.com/ionic-team/capacitor-plugins/issues/1201


theproducer (approved):
Looks good

giralte-ionic (approved):

Issue #1102, opened 2022-07-30
@capacitor/google-maps could only be showed in PC but mobile devices


mannok:
## Bug Report

**It is not related to IonContent Backgroud Color Problem**

### Plugin(s)
@capacitor/google-maps

### Capacitor Version
```
Capacitor Doctor   

Latest Dependencies:

  @capacitor/cli: 4.0.1
  @capacitor/core: 4.0.1
  @capacitor/android: 4.0.1
  @capacitor/ios: 4.0.1

Installed Dependencies:

  @capacitor/cli: 4.0.1
  @capacitor/ios: 4.0.1
  @capacitor/android: 4.0.1
  @capacitor/core: 4.0.1

[success] Android looking great! 👌
```

### Platform(s)
- Android
- Haven't tested in iOS

### Current Behavior
The map could only be showed in PC Chrome. It is not even initialized and displayed in Android App. 

### Expected Behavior
The map can be showed as how PC Chrome displayed

### Code Reproduction
https://github.com/mannok/capacitor-gmap-on-device-issue

### Other Technical Details
PC Chrome: tag was initialized and content was inserted
![image](https://user-images.githubusercontent.com/458220/181870541-a94f0aba-2e8b-4e2b-b4ff-a958efc2484c.png)

Android Remote Debug: seems tag wasn't be modified, `result CapacitorGoogleMaps.create (#69605420)` showing `undefined`
![image](https://user-images.githubusercontent.com/458220/181870626-dec983ef-ec7f-4990-ba21-aec49e276cf3.png)

AndroidManifest.xml
```
<?xml version='1.0' encoding='utf-8'?>
<manifest package="com.vueonic" xmlns:android="http://schemas.android.com/apk/res/android">
    <application android:allowBackup="true" android:icon="@mipmap/ic_launcher" android:label="@string/app_name" android:roundIcon="@mipmap/ic_launcher_round" android:supportsRtl="true" android:theme="@style/AppTheme" android:usesCleartextTraffic="true">
        <activity android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode" android:exported="true" android:label="@string/title_activity_main" android:launchMode="singleTask" android:name="com.vueonic.MainActivity" android:theme="@style/AppTheme.NoActionBarLaunch">
            <intent-filter>
                <action android:name="android.intent.action.MAIN" />
                <category android:name="android.intent.category.LAUNCHER" />
[...]

PR #1252, opened 2022-10-27
fix(google-maps): Retry getting the map size if it's 0

jcesarmobile:
There is a bug on ionic/vue that makes custom elements have 0 size on mounted, and probably some other edge cases that could cause the map to have 0 size.

This PR will retry trying to get the google maps custom element size for 3 seconds every 100ms to solve those problems and warn if it still couldn't get the map size after the 3 seconds.

closes https://github.com/ionic-team/capacitor-plugins/issues/1102

### DIFF
--- a/google-maps/src/map.ts
+++ b/google-maps/src/map.ts
@@ -144,7 +144,7 @@ export class GoogleMap {
     newMap.element = options.element;
     newMap.element.dataset.internalId = options.id;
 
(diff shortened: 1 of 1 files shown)


giralte-ionic (approved):

IT-MikeS (approved):

fryiee:
@jcesarmobile when will this be tagged for release? Thanks

PR #1257, opened 2022-11-03
chore(android): native artifact publish logic

carlpoole:
Similarly to https://github.com/ionic-team/capacitor/pull/6022 this pull request adds the native publication scripts into main and eliminates the need for the `native-publish` branch in the capacitor-plugins repo.

The plugins use a separate ENV variable in the flow `CAP_PLUGIN_PUBLISH` to distinguish from Capacitor so it doesn't trigger the scripts inside the Capacitor core project and cause a gradle error.

- The publish-android job will now attempt to fully publish the plugin all the way to production if the script checks are successful.

- Updated the publication script to also support Kotlin plugins like google-maps.

### DIFF
--- a/.github/workflows/publish-android.yml
+++ b/.github/workflows/publish-android.yml
@@ -7,6 +7,10 @@ on:
         description: 'Specify an optional subset of plugins to publish (space delimited)'
         required: false
         default: ''
+      capacitor-version:
+        description: 'Optional. Specify the version of Capacitor the plugins should depend on. Must be in mathematical notation, eg: [4.0,5.0) for 4.x versions only, or [4.0,) for 4.x versions and higher'
+        required: false
+        default: ''
 jobs:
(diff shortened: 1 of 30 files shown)


jcesarmobile (commented):
Instead of getting latest capacitor version from github, why don't we use `4.+` or similar syntax?

Since we don't release all the plugins at one, not sure if it could cause problems if an user is on latest capacitor version, but has plugins that weren't released on latest release and have a dependency to an older version of capacitor.

carlpoole:
> Instead of getting latest capacitor version from github, why don't we use `4.+` or similar syntax?

I think it does make sense to resolve the latest version, and if devs want fine grained control over which Capacitor version they need they can do it in gradle with an override. Change implemented @jcesarmobile

jcesarmobile (approved):

PR #1259, opened 2022-11-04
fix IllegalArgumentException

j05t:
Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.


jcesarmobile:
This is fixed on version 4.x of the plugin, we are not backporting this kind of changes to older versions nor doing releases of the old versions unless it's because of a security related problem.

PR #1299, opened 2022-11-23
chore(ci): Fix publish-ios.yml 

Steven0351:
Updates publish-ios.yml to use the working-directory of the plugin. Removes unneeded dependency caching and npm install.

### DIFF
--- a/.github/workflows/publish-ios.yml
+++ b/.github/workflows/publish-ios.yml
@@ -16,25 +16,17 @@ jobs:
         plugin: ${{ fromJson(github.event.inputs.plugins) }}
     steps:
       - run: sudo xcode-select --switch /Applications/Xcode_13.3.1.app
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v3
         with:
-          node-version: 14.x
-      - uses: actions/checkout@v2
+          node-version: 16
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


giralte-ionic (approved):

.github/workflows/publish-ios.yml
```diff
@@ -16,25 +16,17 @@ jobs:
         plugin: ${{ fromJson(github.event.inputs.plugins) }}
     steps:
       - run: sudo xcode-select --switch /Applications/Xcode_13.3.1.app
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v3
         with:
-          node-version: 14.x
-      - uses: actions/checkout@v2
+          node-version: 16
```
IT-MikeS, line 21: It shouldn't matter as we aren't doing `npm install` but I'd leave it as `14.x` as the `npm` version that comes with can change the behaviour of peer deps.

jcesarmobile (approved):

.github/workflows/publish-ios.yml
Steven0351: I had thought 14 was EOL

Steven0351: Looks like it still has 5 months of security support left: https://endoflife.date/nodejs

Issue #1283, opened 2022-11-15
No Activity found to handle Intent VIEW


RRGT19:
### Plugin(s)

- "@capacitor/browser": "^4.0.1"

### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.4.0
  @capacitor/core: 4.4.0
  @capacitor/android: 4.4.0
  @capacitor/ios: 4.4.0

Installed Dependencies:

  @capacitor/cli: 4.4.0
  @capacitor/core: 4.4.0
  @capacitor/android: 4.4.0
  @capacitor/ios: 4.4.0

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)

- Android

### Current Behavior
Watching Firebase Crashlytics, I saw users who cannot open my intent to launch a website. Here the logs that I see on Crashlytics:

**RuntimeException**

```
Fatal Exception: java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
at com.getcapacitor.Bridge.lambda$callPluginMethod$0$com-getcapacitor-Bridge(Bridge.java:730)
at com.getcapacitor.Bridge$$ExternalSyntheticLambda5.run(:8)
at android.os.Handler.handleCallback(Handler.java:883)
at android.os.Handler.dispatchMessage(Handler.java:100)
at android.os.Looper.loop(Looper.java:237)
at android.os.HandlerThread.run(HandlerThread.java:67)
```

**InvocationTargetException**

```
Caused by java.lang.reflect.InvocationTargetException:
       at java.lang.reflect.Method.invoke(Method.java)
       at com.getcapacitor.PluginHandle.invoke(PluginHandle.java:125)
       at com.getcapacitor.Bridge.lambda$callPluginMethod$0$com-getcapacitor-Bridge(Bridge.java:721)
       at com.getcapacitor.Bridge$$ExternalSyntheticLambda5.run(:8)
       at android.os.Handler.handleCallback(Handler.java:883)
       at android.os.Handler.dispatchMessage(Handler.java:100)
       at android.os.Looper.loop(Looper.java:237)
       at android.os.HandlerThread.run(HandlerThread.java:67)
```

**ActivityNotFoundException**

```
Caused by android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=https://www.google.com/... (has extras) }
[...]

PR #1323, opened 2022-12-09
fix(browser): Reject if custom tabs are not available

jcesarmobile:
closes https://github.com/ionic-team/capacitor-plugins/issues/1283

### DIFF
--- a/browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserPlugin.java
+++ b/browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserPlugin.java
@@ -1,5 +1,6 @@
 package com.capacitorjs.plugins.browser;
 
+import android.content.ActivityNotFoundException;
 import android.net.Uri;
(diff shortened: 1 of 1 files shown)


giralte-ionic (approved):

IT-MikeS:
Should we add a more robust error message, like "CustomTabsIntent was not found on this device, so the URL could not be displayed" even to just the android logs so developers know why the URL wasn't displayed rather then it just couldn't?

jcesarmobile:
I used the same error message iOS shows at the moment if it couldn’t open the browser, but yeah, I can change to something else.
But talking about CustomTabsIntent is probably confusing as most users won’t know what that is. We don’t even mention “Custom Tabs” anywhere in browser docs.

IT-MikeS:
Perhaps we leave the error sent through `reject()` but add an android log statement that talks about custom tabs missing, just so people have something to search on or ask on forums/discord with.

IT-MikeS (approved):

Issue #1248, opened 2022-10-25
App Foreground -> Notification not-clickable


meriturva:
We have a really simple app that uses the capacitor notification plugin `@capacitor/push-notifications` version 4.1.1

Notifications work correctly when the app is in background, single notification are clickable and the app handles the action correctly.

The main problem is when a notification is received when the app is in foreground. Notification is shown but not clickable.

Our capacitor configuration is:

```
PushNotifications: {
      presentationOptions: ['badge', 'alert', 'sound'],
    },
```

On logcat we have error every time we click on notification:

`0-25 14:51:42.658  2947-2947  NotifActivityStarter    pid 2947 E  onNotificationClicked called for non-clickable notification! 0|xxxx.app|0|null|10465`

Basically, we would like to know how to handle notification clicks when app is in foreground.

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

Please see the Contributing Guide for [how to create a Code Reproduction](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
Ionitron 💙

Ratcha9:
I encounter the same issue with version 4.1.1 on Android. On ios everything is fine.
I've took a look into PushNotificationsPlugin.java. My guess would be that the notification created in `fireNotification(..)` does not contain the key "google.message_id" within the extras bundle, see [here](https://github.com/ionic-team/capacitor-plugins/blob/5bb30005df075ce1679695792e33f398eee298c4/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java#L54).

Ionitron:
It looks like this issue didn't get the information it needed, so I'll close it for now. If I made a mistake, sorry! I am just a bot.

Have a great day!
Ionitron 💙

meriturva:
@Ratcha9 have you any news from your side?

Ratcha9:
> @Ratcha9 have you any news from your side?

I've build a workaround by sending data notifications and creating local notifications in java. 

This Issue shouldn't be closed. I don't see  to provide a reproduction when it comes to notifications from 3rd Party. Some guidance would be appreciated.

Ratcha9:
@meriturva please reopen the issue

meriturva:
[...]

PR #1324, opened 2022-12-10
fix(pushnotifications): App Foreground -> Notification not-clickable

schmop:
### Problem:
When the app is in the foreground, or even active in the background, received push notifications are not clickable at the moment.

### Cause:
I debugged it to be a problem in the [`fireNotification`](https://github.com/ionic-team/capacitor-plugins/blob/c884c38cb2d24105e4667e32ffb6bbe59c97b9b4/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java#L242) method. When creating a local copy of the remote notification (which only happens, if the app is running in back- or foreground), the newly created notification is created without `Intend`.
I am no android developer, excuse me, if I am not firm in the lingo.

### Solution
The proposed changes introduce an intend to the local notification that
* opens the app
* sends the notification data
* dismisses the notification as read.


schmop:
References #1248

marcelbloch:
seems to fix the issue

brilliantinsane:
Will this ever be merged?

junekpavel:
Thanks a lot for saving my day, applied this via [patch-package](https://www.npmjs.com/package/patch-package) and it works like a charm.

AntoscencoVladimir:
Same here ...

trinitiwowka:
Any news?

PR #1327, opened 2022-12-12
fix(keyboard): use view intersection for calculating keyboard height

jcesarmobile:
Thanks @ItsChaceD for the `CoordinateSpace` tip

closes https://github.com/ionic-team/capacitor-plugins/issues/1285

### DIFF
--- a/keyboard/ios/Plugin/Keyboard.m
+++ b/keyboard/ios/Plugin/Keyboard.m
@@ -125,7 +125,9 @@ NSString* UITraitsClassString;
     [hideTimer invalidate];
   }
   CGRect rect = [[notification.userInfo valueForKey:UIKeyboardFrameEndUserInfoKey] CGRectValue];
-  double height = rect.size.height;
+  CGRect webViewAbsolute = [self.webView convertRect:self.webView.frame toCoordinateSpace:self.webView.window.screen.coordinateSpace];
+  CGRect intersection = CGRectIntersection(rect, webViewAbsolute);
+  double height = CGRectIntersection(rect, webViewAbsolute).size.height;
 
   double duration = [[notification.userInfo valueForKey:UIKeyboardAnimationDurationUserInfoKey] doubleValue]+0.2;
   [self setKeyboardHeight:height delay:duration];


ItsChaceD:
I think some logic may need to be added to handle the case where the on screen keyboard is minimized.

Screenshot 2022-12-12 at 9 09 56 AM

Also I noticed that if the keyboard is not overlapping the app at all, the padding height is still set to '3' when it should be '0'.

ItsChaceD (changes requested):
^

jcesarmobile:
That bar appears when using the hardware keyboard, the issue was already present and not related to this change. So if should be evaluated in a separate issue and fixed in a separate PR.
I thought that was a bug on the iPad simulator as before doing the change it was doing a full resize as if the software keyboard appeared but it didn't as the simulator was using the hardware keyboard.
I don't have a hardware keyboard to test and verify if it's how it works on iPadOS or if it's a bug. I tried with two USB keyboard and they don't work on my iPad and I have no bluetooth keyboard to test. 

In which emulator are you testing for getting the keyboard to not overlap the app?
In my tests the keyboard always covers the app by a little even on the smaller size of the app that I'm allowed to resize.
In the 12.9 inch iPad pro I get 26.5 as min height and matches perfectly the overlapped part, and on the 11 inch iPad pro I get 6 as min height and also resizes to the correct position.

ItsChaceD:
> That bar appears when using the hardware keyboard, the issue was already present and not related to this change. So if should be evaluated in a separate issue and fixed in a separate PR. I thought that was a bug on the iPad simulator as before doing the change it was doing a full resize as if the software keyboard appeared but it didn't as the simulator was using the hardware keyboard. I don't have a hardware keyboard to test and verify if it's how it works on iPadOS or if it's a bug. I tried with two USB keyboard and they don't work on my iPad and I have no bluetooth keyboard to test.
> 
> In which emulator are you testing for getting the keyboard to not overlap the app? In my tests the keyboard always covers the app by a little even on the smaller size of the app that I'm allowed to resize. In the 12.9 inch iPad pro I get 26.5 as min height and matches perfectly the overlapped part, and on the 11 inch iPad pro I get 6 as min height and also resizes to the correct position.

Ok that makes sense for the hardware keyboard portion. This could just be an emulator bug.

For the other issue, I am also using the 12.9 inch iPad Pro emulator. Here is an example of what I am talking about. Shouldn't there be 0 padding height instead of having 3?
Screenshot 2022-12-13 at 9 05 44 AM

jcesarmobile:
Ah, yeah, also looks like a bug on the keyboard x, y values when two stage manager windows are present, which makes the CGRectIntersection to have incorrect values.
Will try switching to doing some math with the screen height, keyboard height and WebView height instead of using the CGRectIntersection

ItsChaceD (approved):

giralte-ionic (approved):

staff0rd:
This is [breaking](https://github.com/ionic-team/capacitor-plugins/issues/1413)

Issue #1325, opened 2022-12-10
`Camera.pickImages(...)` ignores `limited` permission state


robingenz:
### Plugin(s)

- `@capacitor/camera@4.1.4`

### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 4.6.1
  @capacitor/core: 4.6.1
  @capacitor/android: 4.6.1
  @capacitor/ios: 4.6.1

Installed Dependencies:

  @capacitor/cli: 4.6.0
  @capacitor/core: 4.6.0
  @capacitor/android: 4.6.0
  @capacitor/ios: 4.6.0

[success] iOS looking great! 👌
[success] Android looking great! 👌

```

### Platform(s)

- iOS

### Current Behavior

If the user only allows access to specific images in the app's permissions, all images on the device will still appear in Photo Picker (`Camera.pickImages(...)`).

|Restricted access to specific photos|All photos selectable|
|----|----|
|![Simulator Screen Shot - iPhone 13 - 2022-12-10 at 22 31 18](https://user-images.githubusercontent.com/13857929/206876228-39a655a5-3da4-4257-b29c-4ba50902803e.png)|![Simulator Screen Shot - iPhone 13 - 2022-12-10 at 22 34 44](https://user-images.githubusercontent.com/13857929/206876254-c95e9ed2-3d9f-49c6-8b77-18d499f480ee.png)|

### Expected Behavior

I expect the Photo Picker to display only the images for which the user has given permission.

### Code Reproduction

I will provide the reproducible example shortly.
Maybe I just misunderstood something.

### Other Technical Details

iOS Version: 15.5

### Additional Context

I have not much experience with the photo picker on iOS. So far I thought that iOS handles this and restricts access. But apparently the developer has some freedom here.

Issue #1297, opened 2022-11-22
Camera plugin for iOS always prompts for permissions, confusing behavior when limited library is selected


chudooder:
### Plugin(s)

Observed on:

```
@capacitor/camera: 4.1.4
```

### Capacitor Version

```
Latest Dependencies:

  @capacitor/cli: 4.5.0
  @capacitor/core: 4.5.0
  @capacitor/android: 4.5.0
  @capacitor/ios: 4.5.0

Installed Dependencies:

  @capacitor/cli: 3.3.3
  @capacitor/core: 4.3.0
  @capacitor/android: 4.3.0
  @capacitor/ios: 4.3.0
```

### Platform(s)

Observed on iOS 15.6 - iPhone X

### Current Behavior

After a fresh install of my app, the user clicks a button that calls `Camera.getPhoto`. The user is prompted to either select a photo or take a picture (expected due to `CameraSource.Prompt`). After clicking "Select a photo", the user is prompted to provide either limited access, or to allow access to all photos. If the user picks a limited library to share to the app, the photo picker that comes up after this will show the entire photo library regardless. This is causing confusion as the user expects to only see the limited set of photos they selected.

### Expected Behavior

Since `PHPicker` doesn't require an explicit permissions grant to share photos with the app, the user should not see the initial permissions prompt to select a limited photo library. Instead, calling `getPhoto` should only prompt for permissions when necessary, and instead immediately open `PHPicker` in any other case.

### Code Reproduction

Called `Camera.getPhoto` with the following options:

```ts
const cameraPhoto = await Camera.getPhoto({
  quality: 90,
  saveToGallery: true,
  resultType: CameraResultType.DataUrl,
  source: CameraSource.Prompt,
  webUseInput: true,
});
```

### Additional Context

https://developer.apple.com/documentation/photokit/delivering_an_enhanced_privacy_experience_in_your_photos_app

jcesarmobile:
While the `PHPickerViewController` doesn't need the permission to appear, the permission is needed for getting the image information (exif). So that's why the plugin requests the permission.

If the user grants the limited permission, then only the photos that were picked will include the exif information.
Sadly the `PHPickerViewController` doesn't provide a way of only showing the picked images, and also Apple doesn't provide a different picker for only the selected pictures.

[...]

PR #1328, opened 2022-12-12
feat(camera): add option to exclude Exif data on iOS

robingenz:
The `pickImages` function should only ask for permission if it is really necessary. 
This is the case on iOS only when you need the Exif data (see [comment](https://github.com/ionic-team/capacitor-plugins/issues/1325#issuecomment-1346182920)). My client does not need this data and therefore wants to improve the UX by not asking permission for data that is not used.

Related to #1297 and #1325

### DIFF
--- a/camera/README.md
+++ b/camera/README.md
@@ -249,14 +249,15 @@ Request camera and photo album permissions
 
 #### GalleryImageOptions
 
-| Prop                     | Type                                   | Description                                                                                | Default                     | Since |
-| ------------------------ | -------------------------------------- | ------------------------------------------------------------------------------------------ | --------------------------- | ----- |
-| **`quality`**            | <code>number</code>                    | The quality of image to return as JPEG, from 0-100                                         |                             | 1.2.0 |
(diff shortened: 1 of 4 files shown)


robingenz:
The [Capacitor File Picker](https://github.com/capawesome-team/capacitor-file-picker) plugin now supports this (see [`pickImages(...)`](https://github.com/capawesome-team/capacitor-file-picker#pickimages)).

mesqueeb:
@robingenz since file picker and camera roll picker are separate, it's still useful to have this feature in the camera roll picker.

robingenz:
@mesqueeb Yes, it would be nice if this was also possible with the plugin. I reopened the pr.

kentora:
This makes for much less confusion for the end users, and therefore would result in less 1 star reviews from 'paranoid' type users, complaining we can see pictures they did not agree to us seeing. I sure hope this can get merged

kentora:
A suggestion might be to add the option to exclude exif data when using getPhoto, as this can also redirect to showing the picker...

fanckush:
any updates on this PR? it's gotten stale but I think it's very valuable

eric-horodyski:

More context:

PR #1331, opened 2022-12-14
Fix(camera): bug in SDK 33 where opening the gallery when taking a picture fails due to permissions

modus-jose:
## Problem
With the release of Android 13, SDK 33 the permission for accesing media files changed.
https://developer.android.com/about/versions/13/behavior-changes-13

When using the camera plugin 
It gives the option to open the gallery which fails with a no permissions error in Android 13.

Even when adding permissions in an ionic app, this plugin still doesn't has the permission to read media files

## Solution
Detect if the app is using the sdk 33, and grant/check the proper permissions


ItsChaceD (approved):

jcesarmobile (changes requested):
the alias should be a simpler string, maybe call it `media`?

The new alias should be added to the types so it can be requested independently
https://github.com/ionic-team/capacitor-plugins/blob/main/camera/src/definitions.ts#L5-L10

And also should be documented in the README

markemer:
Address @jcesarmobile's comments and conflicts and this should be gtg.

jcesarmobile:
This was addressed by the linked PR from chace

Issue #1309, opened 2022-11-29
Google maps clustering don't work in 4.3.1 version


xdelvalle:
### Plugin(s)

@capacitor/google-maps

### Capacitor Version

```
Latest Dependencies:

  @capacitor/cli: 4.5.0
  @capacitor/core: 4.5.0
  @capacitor/android: 4.5.0
  @capacitor/ios: 4.5.0

Installed Dependencies:

  @capacitor/cli: 4.5.0
  @capacitor/android: 4.5.0
  @capacitor/core: 4.5.0
  @capacitor/ios: 4.5.0

```

### Platform(s)

Android

### Current Behavior

When enableClustering() is called, the markers are clustered, but when zoom in, is not possible to see individual markers again (always clustered)

### Expected Behavior

When zoom in, the individual markers are displayed.

### Code Reproduction

await this.addCustomMarkers();
await this.map.enableClustering();

```
  async addCustomMarkers() {
    for (const element of this.elements) {
      const marker: Marker = {
        coordinate: {
          lat: element.latitude,
          lng: element.longitude
        },
        title: element.name,
        snippet: element.name,
        iconUrl: 'assets/icon/markers/marker.png',
      };
      this.markers.push(marker);
    }
    await this.map.addMarkers(this.markers);
  }
```

### Other Technical Details

"@capacitor/android": "^4.5.0",
"@capacitor/app": "^4.1.1",
"@capacitor/core": "^4.5.0",
"@capacitor/geolocation": "^4.1.0",
**"@capacitor/google-maps": "4.2.1",**
"@capacitor/haptics": "4.1.0",
"@capacitor/ios": "^4.5.0",
"@capacitor/keyboard": "^4.1.0",
"@capacitor/status-bar": "^4.1.0",

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

Please see the Contributing Guide for [how to create a Code Reproduction](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
Ionitron 💙

xdelvalle:
Minimal Github project

https://github.com/xdelvalle/test-maps.git

jcesarmobile:
In your sample app, the markers are not getting clustered, but that's an issue too.
If I reload the app then they get clustered and don'g get unclustered if I zoom out.

[...]

PR #1383, opened 2023-01-19
fix(google-maps): Update Clusters when moving map

ItsChaceD:
This PR addresses #1309, where (on android) clustering does not update when the map is moved, rotated, or zoomed in/out. It also allows modification of the minimum cluster size so behavior can be matched across all platforms.

### DIFF
--- a/google-maps/README.md
+++ b/google-maps/README.md
@@ -270,7 +270,7 @@ export default MyMap;
 <docgen-index>
 
 * [`create(...)`](#create)
-* [`enableClustering()`](#enableclustering)
+* [`enableClustering(...)`](#enableclustering)
 * [`disableClustering()`](#disableclustering)
 * [`addMarker(...)`](#addmarker)
 * [`addMarkers(...)`](#addmarkers)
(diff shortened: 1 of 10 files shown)


IT-MikeS (approved):

jcesarmobile (commented):
types say 4 is the default for minClusterSize, but the default is only being set to 4 for web.
Is it 4 the default native value? or should it be set to 4 there too if the minClusterSize option is not passed?

This should be two separate pull requests, one for the feat and another for the fix, or at least turn this into a feat PR.

ItsChaceD:
> types say 4 is the default for minClusterSize, but the default is only being set to 4 for web. Is it 4 the default native value on Google Maps SDK? or should it be set to 4 there too if the minClusterSize option is not passed?
> 
> This should be two separate pull requests, one for the feat and another for the fix, or at least turn this into a feat PR.

Yes, you are right, they should be separate PRs. I split this into #1398 and #1399 

And yes, 4 is the default native value for the Google Maps SDK on Android & iOS. I think this is different on web, so I explicitly set it to 4 if it is undefined.

PR #914, opened 2022-04-04
fix(local-notifications): Add FLAG_MUTABLE to pending intents for SDK 31

jcesarmobile:
closes https://github.com/ionic-team/capacitor-plugins/issues/689

PR #1391, opened 2023-01-26
fix: Add FLAG_MUTABLE to pending intents for SDK 31 support

tafelnl:
Backport of #913, #914 and #1173

I know normally you don't do backports. Especially since @jcesarmobile mentioned adding support for SDK 31 itself, would be a feature, which won't be shipped to Capacitor v3 anymore.

But please allow me to explain why I think fixes like these would still make sense to me:

(I am assuming targeting SDK 31 is required by Google Play Store, which at the moment, it is)

I can agree that adding explicit support for SDK 31 to Capacitor v3 can/should be considered a feature in itself. But if you won't be shipping fixes like this one, Capacitor v3 is essentially useless and completely unsupported as a whole for new projects. Okay, that would still be fair. But for large _existing_ projects, that means, no native updates can be done, unless you would upgrade to Capacitor v4. The migration of v3 to v4 is pretty painless, for which I applaud your team. But for enterprise-sized projects, it's still a major thing to do, and can't be done in a few days, week or even months. But during that time, I think it's fair to say you would still need to be able to ship new native updates.

Even if - after this speech ;) - you still don't agree with me on this, I think it's fair to explicitly mention in the docs, that Capacitor v3 only supports up to SDK 30. Because otherwise someone might increase the targetSdkVersion himself, and run into unexpected runtime errors (I had this myself).

### DIFF
--- a/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationManager.java
+++ b/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationManager.java
@@ -254,12 +254,11 @@ public class LocalNotificationManager {
         // Open intent
         Intent intent = buildIntent(localNotification, DEFAULT_PRESS_ACTION);
 
-        PendingIntent pendingIntent = PendingIntent.getActivity(
-            context,
-            localNotification.getId(),
-            intent,
-            PendingIntent.FLAG_CANCEL_CURRENT
-        );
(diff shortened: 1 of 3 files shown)


jcesarmobile:
[as you can see](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#branches) in the CONTRIBUTING.md file

3.x branch will only get security fixes (if needed).

tafelnl:
For those reading along: Capacitor has added a support policy recently: https://github.com/ionic-team/capacitor-docs/pull/187

PR #1399, opened 2023-01-30
feat(google-maps): Add minClusterSize setting for enableClustering

ItsChaceD:
This PR allows modification of the minimum cluster size so behavior can be matched across all platforms. (Default minClusterSize is 4 on the Android & iOS SDKs)

### DIFF
--- a/google-maps/src/implementation.ts
+++ b/google-maps/src/implementation.ts
@@ -107,11 +107,6 @@ export interface OnScrollArgs {
 
 export interface EnableClusteringArgs {
   id: string;
-  /**
-   * The minimum number of markers that can be clustered together.
-   *
-   * @default 4
-   */
   minClusterSize?: number;
 }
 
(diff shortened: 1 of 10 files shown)


google-maps/src/implementation.ts
```diff
@@ -105,13 +105,23 @@ export interface OnScrollArgs {
   };
 }
 
+export interface EnableClusteringArgs {
+  id: string;
+  /**
+   * The minimum number of markers that can be clustered together.
+   *
+   * @default 4
+   */
```
jcesarmobile, line 114: The way google-maps plugin is built won't show this information to the users nor appear on the README.
For it to be visible it should be moved to the `enableClustering` method definition in `map.ts` file.

ItsChaceD: Ok, I moved it for this PR. However, I'm also going to create a separate PR to copy the docs on the method definitions to the GoogleMapsInterface since the README is missing descriptions for all of the methods.

jcesarmobile (approved):

giralte-ionic (approved):

giralte-ionic (approved):

PR #1409, opened 2023-02-07
Avoid crashing by catching and rejecting the call.

boldtrn:
There is an issue when using preferences. Right now the preferences plugin assume that they are always Strings. 

The plugin has no way to pass more information on which type it should fetch. When you are getting a preference that is for example Long, the app will crash.

Wrapping the JS in a try/catch does not catch the exception. This change catches class cast exceptions to avoid crashing the app if the types don't match.


eric-horodyski:

More context:

PR #1416, opened 2023-02-10
Add no-op Keyboard web plugin

cornflakes:
When using the Keyboard plugin, we get the following log in the console, for an app using Capacitor in a desktop web browser:
image
`Error: Uncaught (in promise): Error: "Keyboard" plugin is not implemented on web`

Purpose of this PR is to silence this with a no-op keyboard web plugin.


jcesarmobile:
Thanks for the PR, but we discussed it internally and we don't want no-op implementations for plugins.

Issue #1347, opened 2022-12-26
add - googleMaps : getMapTypeId


rbalet:
### Plugin

[google-maps](https://www.npmjs.com/package/@capacitor/google-maps)

### Description

Add the [google getMapTypeId() method](https://developers.google.com/maps/documentation/javascript/reference/map#Map.getMapTypeId) to this library

rbalet:
I've open other issues and would be glad to do some pr, but I'm not sure I understand how exactly the `CapacitorGoogleMaps`. 

Can I just simply add Same logics as the [@angular/google-maps](https://github.com/angular/components/blob/main/src/google-maps/google-map/google-map.ts) library or do I need to take care of something else ? 

Here for example I would have added the following code inside the [map.ts](https://github.com/ionic-team/capacitor-plugins/blob/main/google-maps/src/map.ts) file

```typescript
  /**
   * See
   * https://developers.google.com/maps/documentation/javascript/reference/map#Map.getMapTypeId
   */
  getMapTypeId(): google.maps.MapTypeId | string | undefined {
    return CapacitorGoogleMaps.getMapTypeId();
  }
```

and then this, inside the [impkementation.ts](https://github.com/ionic-team/capacitor-plugins/blob/main/google-maps/src/implementation.ts) file.

```typescript
export interface CapacitorGoogleMapsPlugin extends Plugin {
// ...
getMapTypeId(): google.maps.MapTypeId | string | undefined
}
```

But then I do not understand why this is working as is..

PR #1420, opened 2023-02-15
feat(google-maps): Add getMapType method

jcesarmobile:
I've called it `getMapType` instead of the proposed `getMapTypeId` (google maps js SDK name) since we have a `setMapType` function and the native methods in google maps SDK are called `getMapType`.

Also returns our existing `MapType`  instead of the proposed `google.maps.MapTypeId` since they don't match.

closes https://github.com/ionic-team/capacitor-plugins/issues/1347


IT-MikeS (approved):

giralte-ionic (approved):

rbalet:
Hi @jcesarmobile  and thx for this one.

Could you enlighten me on the reason why you aren't using the same naming convention as the google maps js SDK one ?
It would made the think easier to understand, don't you think ?

PR #1423, opened 2023-02-18
feat(push-notifications): allow more control over foreground notifications

### LINKED ISSUE #1368: Use default notification channel for foreground push notifications
### Plugin
"@capacitor/push-notifications": "4.1.2",

### Description

If [presentationOptions](https://capacitorjs.com/docs/apis/push-notifications#push-notifications-appearance-in-foreground) are given in the capacitor config, the push notification plugin automatically [creates a notification channel](https://github.com/ionic-team/capacitor-plugins/blob/3c90fea63e2c24deacc38ba9fd8471c2cc0bde56/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/NotificationChannelManager.java#L155-L157) to use for those foreground notifications, with a hard-coded name and description ("Push Notifications Foreground" / "Push notifications in foreground"). This is quite unexpected, especially if the app is internationalized and/or specifies its own channels.

### Platform(s)

Android

### Preferred Solution

Push notifications in foreground should hade the same behavior as other push notifications, and use the "regular" [default notification channel](https://stackoverflow.com/questions/46047343/firebase-how-to-set-default-notification-channel-in-android-app) if no channel is specified on the notification.

An issue with this solution is however that a notification channel might be needed in order to use the "sound" option for `presentationOptions`.

### Alternatives

An alternative solution would be to make it possible to specify a `channelId` to use for this in settings and let the user be responsible for creating the channel themselves if set.

Another option would be to document this behavior so that developers at least knows that this is happening when using `presentationOptions` and instead handle foreground notifications themselves if they do not want the current behavior.

### LINKED ISSUE #1388: Custom notification sound is not played while the app is in foreground
### Plugin(s)

@capacitor/push-notifications": "^4.0.0"

### Capacitor Version

```
Latest Dependencies:

  @capacitor/cli: 4.6.2
  @capacitor/core: 4.6.2
  @capacitor/android: 4.6.2
  @capacitor/ios: 4.6.2

Installed Dependencies:

  @capacitor/cli: 4.6.1
  @capacitor/ios: 4.6.1
  @capacitor/core: 4.6.1
  @capacitor/android: 4.6.1
```

### Platform(s)

Android

### Current Behavior

I created an android notification channel and a custom notification sound for it. The issue is that the custom sound only plays while the user has the app closed or in background.

### Expected Behavior

The custom notification sound should be played even if the app is in foreground.

### Code Reproduction

```            
await PushNotifications.createChannel(
            {id: 'orders', name: 'orders', importance: 3, sound: 'order_sound'}
      );
```

RaddishIoW:
Adds options for customising the android foreground notifications, while maintaining default behaviour if the settings are not specified.
Also addresses issue with clicking notifications - see #1324 

Fixes #1368
Fixes #1388

### DIFF
--- a/push-notifications/README.md
+++ b/push-notifications/README.md
@@ -56,9 +56,13 @@ Android Studio has an icon generator you can use to create your Push Notificatio
 
 You can configure the way the push notifications are displayed when the app is in foreground.
 
-| Prop                      | Type                              | Description                                                                                                                                                                                                                                                                                                                                                                                          | Since |
-| ------------------------- | --------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----- |
-| **`presentationOptions`** | <code>PresentationOption[]</code> | This is an array of strings you can combine. Possible values in the array are: - `badge`: badge count on the app icon is updated (default value) - `sound`: the device will ring/vibrate when the push notification is received - `alert`: the push notification is displayed in a native dialog An empty array can be provided if none of the options are desired. badge is only available for iOS. | 1.0.0 |
+| Prop                                      | Type                                        | Description                                                                                                                                                                                                                                                                                                                                                                                          | Since |
+| ----------------------------------------- | ------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----- |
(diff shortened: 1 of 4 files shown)


RaddishIoW:
This PR would enable app developers to gain more control over the way that notifications are presented while the app is in the foreground, while maintaining the current behaviour by default.

The PR adds options to the plugin config allowing the app developer to:
1. Specify options (e.g. sound) for the default, built in, foreground notification channel, or
2. Specify a channel ID for a channel which they have created themselves for the display of notifications in the foreground, or
3. Use the channel ID specified in the original notification payload to display the notification in the foreground.

RaddishIoW:
Does the Capacitor team welcome PRs from the community? Doesn't seem like they merge anything not created by an Ionic team member, from what I can see...

I've created a separate plugin which is a drop-in replacement for the "official" one (forked from it), which includes these features and fixes, in case anyone else comes across this:
[@haylltd/capacitor-push-notifications](https://www.npmjs.com/package/@haylltd/capacitor-push-notifications)

PR #1447, opened 2023-03-02
chore(status-bar): set default statusbar animation to fade on iOS

IT-MikeS:


jcesarmobile (changes requested):
we should deprecate slide and also indicate that fade is the default in the types

jcesarmobile (approved):

markemer (approved):

PR #1448, opened 2023-03-03
chore(google-maps)(android): update gradle and kotlin dependencies

ItsChaceD:
Update all AndroidX, Gradle, and Kotlin dependencies for the `google-maps` plugin

### DIFF
--- a/google-maps/android/build.gradle
+++ b/google-maps/android/build.gradle
@@ -1,19 +1,19 @@
 ext {
     capacitorVersion = System.getenv('CAPACITOR_VERSION')
     junitVersion = project.hasProperty('junitVersion') ? rootProject.ext.junitVersion : '4.13.2'
-    androidxAppCompatVersion = project.hasProperty('androidxAppCompatVersion') ? rootProject.ext.androidxAppCompatVersion : '1.4.2'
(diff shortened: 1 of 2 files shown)


google-maps/android/build.gradle
```diff
@@ -22,7 +22,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.2.1'
+        classpath 'com.android.tools.build:gradle:7.4.2'
```
jcesarmobile, line 25:
```suggestion
        classpath 'com.android.tools.build:gradle:7.4.1'
```
for now we are using 7.4.1 since that's what latest Android Studio recommends

jcesarmobile (approved):
Wops, I was wrong, there was a new android studio release yesterday that recommends gradle plugin 7.4.2
anyway, we went with 7.2.1 in capacitor android, so it's probably better to stick with the same version everywhere.
Also, we will probably end up updating to gradle 8 before the final release.

ItsChaceD:
> Wops, I was wrong, there was a new android studio release yesterday that recommends gradle plugin 7.4.2 anyway, we went with 7.2.1 in capacitor android, so it's probably better to stick with the same version everywhere. Also, we will probably end up updating to gradle 8 before the final release.

No problem, sounds good to me

theproducer (approved):

jcesarmobile (approved):

PR #1449, opened 2023-03-03
chore(google-maps): update maps dependencies

ItsChaceD:
Update the Google Maps Utils dependencies for the `google-maps` plugin on Android


jcesarmobile (approved):

theproducer (changes requested):
Looks like the Google Maps Utils version is a major version change, and with it are some changes to the `ClusterItem` interface that we need to fix in `CapacitorGoogleMapMarker`.

ItsChaceD:
> Looks like the Google Maps Utils version is a major version change, and with it are some changes to the `ClusterItem` interface that we need to fix in `CapacitorGoogleMapMarker`.

We will have to review/merge #1474to fix this

ItsChaceD:
> changes look good, but looks like something broke with the change, so we should create a new issue about it first and fix it (if can be fixed without making the dependency update first)

We will have to review/merge #1474 to fix this

markemer (approved):

PR #1458, opened 2023-03-07
chore: update lerna

IT-MikeS:


markemer (changes requested):
```
        with:
          path: ${{ steps.npm-cache-dir.outputs.dir }}
          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
          restore-keys: |
            ${{ runner.os }}-node-
```

Needs `${{ hashFiles('**/package-lock.json') }}` on the restore key line as well.

.github/workflows/ci.yml
```diff
@@ -19,15 +19,21 @@ jobs:
         uses: styfle/cancel-workflow-action@ce177499ccf9fd2aded3b0426c97e5434c2e8a73
```
markemer, line 19: We can also probably get rid of this, and add that "github concurrency cancel running jobs" stuff. It's kinda funny that we all gravitated toward this one action - I used it before too.

PR #1463, opened 2023-03-07
feat(push-notifications): support Android 13+ permissions

theproducer:
Implements support for notifications permissions in Android 13 (API 33):
https://developer.android.com/develop/ui/views/notifications/notification-permission

### DIFF
--- a/push-notifications/README.md
+++ b/push-notifications/README.md
@@ -29,6 +29,8 @@ func application(_ application: UIApplication, didFailToRegisterForRemoteNotific
 
 The Push Notification API uses [Firebase Cloud Messaging](https://firebase.google.com/docs/cloud-messaging) SDK for handling notifications.  See [Set up a Firebase Cloud Messaging client app on Android](https://firebase.google.com/docs/cloud-messaging/android/client) and follow the instructions for creating a Firebase project and registering your application.  There is no need to add the Firebase SDK to your app or edit your app manifest - the Push Notifications provides that for you.  All that is required is your Firebase project's `google-services.json` file added to the module (app-level) directory of your app.
 
+Android 13 requires a permission check in order to receive push notifications.  You are required to call `checkPermissions()` and `requestPermissions()` accordingly, when targeting SDK 33.
+
 ### Variables
 
 This plugin will use the following project variables (defined in your app's `variables.gradle` file):
@@ -303,7 +305,7 @@ checkPermissions() => Promise<PermissionStatus>
 
 Check permission to receive push notifications.
 
-On Android the status is always granted because you can always
+On Android 12 and below the status is always granted because you can always
 receive push notifications. If you need to check if the user allows
 to display notifications, use local-notifications plugin.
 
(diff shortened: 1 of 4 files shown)


giralte-ionic (approved):

push-notifications/android/src/main/AndroidManifest.xml
```diff
@@ -8,4 +8,5 @@
             </intent-filter>
         </service>
     </application>
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
```
jcesarmobile, line 11:
```suggestion
```
For push notifications the plugin doesn't need to include the permission because firebase-messaging is already adding it

push-notifications/README.md
```diff
@@ -303,7 +305,7 @@ checkPermissions() => Promise<PermissionStatus>
 
 Check permission to receive push notifications.
 
-On Android the status is always granted because you can always
+On Android 12 and below the status is always granted because you can always
```
jcesarmobile, line 308: this texts comes from the types, should be edited there and then run `npm run build` to propagate it to the README
https://github.com/ionic-team/capacitor-plugins/blob/main/push-notifications/src/definitions.ts#L93-L117

push-notifications/README.md
```diff
@@ -322,7 +324,7 @@ requestPermissions() => Promise<PermissionStatus>
 
 Request permission to receive push notifications.
 
-On Android it doesn't prompt for permission because you can always
+On Android 12 and below it doesn't prompt for permission because you can always
```
jcesarmobile, line 327: this texts comes from the types, should be edited there and then run `npm run build` to propagate it to the README
https://github.com/ionic-team/capacitor-plugins/blob/main/push-notifications/src/definitions.ts#L93-L117

jcesarmobile (changes requested):

push-notifications/README.md
theproducer: Ah, thanks for catching that - I was in markdown mode going through and changing everything.

typefox09:
Do we need to install the Capacitor local-notifications plugin to get these updates? Or will this also be merged to the Capacitor Push-notifications package also? We're only sending server-side notifications

jcesarmobile (approved):

danbunkr:
Is there a plan to release this PR to the `push-notifications` plugin in Capacitor 4?

jcesarmobile:
No, this is not going to be ported to Capacitor 4, to use the notification permission it's required to target SDK 33, and Capacitor 4 targets SDK 32.
Targeting SDK 33 is supported in Capacitor 5.

Changing the target SDK is a breaking change, so there is no way of releasing this in Capacitor 4.

push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
```diff
@@ -68,6 +75,30 @@ protected void handleOnNewIntent(Intent data) {
         }
     }
 
+    @PluginMethod
```
tafelnl, line 78: Shouldn't these methods have the `@Override` annotation?

PR #1464, opened 2023-03-07
chore(android): fix local-notification deprecation warning

markemer:
- Replace getParcelableExtra(String) with getParcelableExtra(String, Class<T>)

### DIFF
--- a/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/TimedNotificationPublisher.java
+++ b/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/TimedNotificationPublisher.java
@@ -29,13 +29,7 @@ public class TimedNotificationPublisher extends BroadcastReceiver {
     public void onReceive(Context context, Intent intent) {
         NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
 
-        Notification notification;
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
-            notification = intent.getParcelableExtra(NOTIFICATION_KEY, Notification.class);
-        } else {
-            notification = intent.getParcelableExtra(NOTIFICATION_KEY);
-        }
-
+        Notification notification = getNotification(intent, NOTIFICATION_KEY);
         notification.when = System.currentTimeMillis();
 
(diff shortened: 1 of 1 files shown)


IT-MikeS (changes requested):
Could you separate the old call out into its own function and suppress it there

markemer:
> 

Yep - I'll do that on the other one too.

jcesarmobile (changes requested):
Mike meant to move just the `intent.getParcelableExtra(NOTIFICATION_KEY);` line to a new function so only that code is suppressed by  the `@SuppressWarnings("deprecation")`.

markemer:
> Mike meant to move just the `intent.getParcelableExtra(NOTIFICATION_KEY);` line to a new function so only that code is suppressed by the `@SuppressWarnings("deprecation")`.

I know - but isn't this a little cleaner? And we could just pull out the old code when we hit 33 and higher, which will be lord only knows when. I'm cool with just pulling out the stuff to stop the warnings too, if you think that's better.

jcesarmobile:
Since the `@SuppressWarnings("deprecation")` will hide any future deprecations in the whole method, if we put code in there and it gets deprecated we won't notice, so that's why we only put the deprecated code

I think it's also possible to use inline suppress warnings, that might be another option

markemer:
> Since the `@SuppressWarnings("deprecation")` will hide any future deprecations in the whole method, if we put code in there and it gets deprecated we won't notice, so that's why we only put the deprecated code
> 
> I think it's also possible to use inline suppress warnings, that might be another option

That's a good point - that could become a black hole of deprecated calls.

markemer:
Took @jcesarmobile's advice and went with @IT-MikeS's first idea to avoid accidentally hiding deprecations in the future.

IT-MikeS (approved):

jcesarmobile (changes requested):
There is another deprecation warning that should be handled:
`local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationsPlugin.java:139: warning: [deprecation] get(String) in BaseBundle has been deprecated
                        extras.put(key, notification.extras.get(key));`

markemer:
> There is another deprecation warning that should be handled: `local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationsPlugin.java:139: warning: [deprecation] get(String) in BaseBundle has been deprecated extras.put(key, notification.extras.get(key));`

Damn it - I'll open a new one and reference it here.

PR #1474, opened 2023-03-09
feat(google-maps): add zIndex to markers

ItsChaceD:
This PR adds the `zIndex` property to Markers. It is also required to update the Maps Util Verion #1449


theproducer (approved):

markemer:
What's making this fail on CI, seems like something related to Java?

theproducer:
It's complaining because in this version of the Google Maps Utils, `getZIndex` doesn't exist to be overridden.  We can probably ignore it so that we can get this merged in for the sake of https://github.com/ionic-team/capacitor-plugins/pull/1449, which will then include the correct version of the Utils.

jcesarmobile (changes requested):
I've made some request changes.

But the zIndex doesn't seem to be working in Android, not sure if some code changes are still missing, but I've created two markers with the same coordinates and different zIndex and it shows both, doesn't matter if the one on top has a lower zIndex, while on web and iOS it gets hidden when the zIndex is smaller.

Also, on web if no zIndex is provided, it's "undefined", if the docs say it's 0, then I think we should default it to 0.

We should probably move the dependency updates PR changes into this PR so build pass

PR #1475, opened 2023-03-09
chore(camera): replace Android 13 deprecations

ItsChaceD:
This PR fixes the Android 13 deprecations in the `camera` plugin


camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
```diff
@@ -372,7 +373,12 @@ public void processPickedImages(PluginCall call, ActivityResult result) {
                     } else if (data.getExtras() != null) {
                         Bundle bundle = data.getExtras();
                         if (bundle.keySet().contains("selectedItems")) {
-                            ArrayList<Parcelable> fileUris = bundle.getParcelableArrayList("selectedItems");
+                            ArrayList<Parcelable> fileUris;
+                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                                fileUris = bundle.getParcelableArrayList("selectedItems", Parcelable.class);
+                            } else {
+                                fileUris = bundle.getParcelableArrayList("selectedItems");
+                            }
```
IT-MikeS, line 381: Should you split out the `fileUris = bundle.getParcelableArrayList("selectedItems");` into is own function and only suppress that just incase `fileUris = bundle.getParcelableArrayList("selectedItems", Parcelable.class);` also gets deprecated at some point. That way it won't get missed in Lint

IT-MikeS (changes requested):

camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
markemer: Yeah - I'm just starting to get used to how android does this - I almost want to load up `InternalUtils` with all the version logic. there is so so much.

jcesarmobile (approved):

IT-MikeS (approved):

PR #1478, opened 2023-03-11
fix(push-notification): notification not dispatching click when received in foreground

ftheomunhoz:
Push notifications are missing the intent to allow them to be clickable. Without that, the event `pushNotificationActionPerformed` will never trigger.

### DIFF
--- a/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
+++ b/push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
@@ -2,6 +2,7 @@ package com.capacitorjs.plugins.pushnotifications;
 
 import android.app.Notification;
 import android.app.NotificationManager;
+import android.app.PendingIntent;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ApplicationInfo;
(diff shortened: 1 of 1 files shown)


push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
```diff
@@ -239,12 +240,22 @@ public void fireNotification(RemoteMessage remoteMessage) {
                     if (bundle != null && bundle.getInt("com.google.firebase.messaging.default_notification_icon") != 0) {
                         pushIcon = bundle.getInt("com.google.firebase.messaging.default_notification_icon");
                     }
+                    Intent intent = new Intent(getContext(), getActivity().getClass());
+                    intent.putExtras(remoteMessage.toIntent().getExtras());
+                    PendingIntent pendingIntent = PendingIntent.getActivity(
+                        getContext(),
+                        0,
```
ajnrajeev, line 247: instead of 0, consider adding a unique id

push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
```diff
@@ -239,12 +240,22 @@ public void fireNotification(RemoteMessage remoteMessage) {
...
                         getContext(),
                         NotificationChannelManager.FOREGROUND_NOTIFICATION_CHANNEL_ID
                     )
                         .setSmallIcon(pushIcon)
                         .setContentTitle(title)
+                        .setAutoCancel(true)
+                        .setContentIntent(pendingIntent)
                         .setContentText(body)
                         .setPriority(NotificationCompat.PRIORITY_DEFAULT);
                     notificationManager.notify(0, builder.build());
```
ajnrajeev, line 261: the 0 here will potentially overwrite the last notification.

ErikBrendel:
@ftheomunhoz I would very much like this change, as for me, the notification tap event is also missing 👍

abennouna:
wow I literally didn't see this PR when I submitted this https://github.com/ionic-team/capacitor/discussions/6439

push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/PushNotificationsPlugin.java
ErikBrendel: But this is a separate issue, right? This 0 (and the newly added one) result in the problem/feature that only one push notification can be shown. But the change in this MR still only adds value to capacitor and doesn't break any existing features, so maybe it would be good to merge this anyway, and deal with the other thing separately?

tanujlight:
I am also facing the same issue, is this PR going to merge or not

raymondmuller:
Tried this today and it works 👍 
Using pnpm patch now to automatically apply these changes, but would be great to see this merged asap.

sawaca96:
Waiting for PR merge...👏

phpMagpie:
Facing some issue with no intent on foreground notifications.  Confirm that this patch adds that missing functionality.  Any idea when this can be worked on and merged?

tanuj-g:
any new update on this?

trinitiwowka:
Any news?

tanuj-g:
any update on this, waiting for this from last 6 months

PR #1479, opened 2023-03-12
fix `.cut()` and `.copy()` error

tachibana-shin:
Since `this.readdir` is currently returning `Promise<ReaddirResult[]>` so we need to fix this or else there will be a path that looks like `from/[object Object]` which gives the 2 functions error ` .copy()` and `.cut()`

> These functions are currently failing in my new app i hope to see a patched release asap ♥

### DIFF
--- a/filesystem/src/web.ts
+++ b/filesystem/src/web.ts
@@ -612,8 +612,8 @@ export class FilesystemWeb extends WebPlugin implements FilesystemPlugin {
           // Move item from the from directory to the to directory
           await this._copy(
             {
-              from: `${from}/${filename}`,
-              to: `${to}/${filename}`,
+              from: `${from}/${filename.name}`,
+              to: `${to}/${filename.name}`,
               directory: fromDirectory,
               toDirectory,
             },


jcesarmobile:
can you provide a sample app that reproduces the issue?

tachibana-shin:
> 

you can run this code to see the error:

```ts
import { Filesystem, Directory, Encoding } from "@capacitor/filesystem"

await Filesystem.mkdir({
  path: "current/src",
  recursive: true,
  directory: Directory.External
})
await Filesystem.writeFile({
  path: "current/src/index.ts",
  data: "const a = 34",
  encoding: Encoding.UTF8,
  directory: Directory.External
})

await Filesystem.copy({
  from: "current/src",
  to: "current/src2",
  directory: Directory.External,
  toDirectory: Directory.External
})
```

jcesarmobile:
can you provide a full sample app in a github repository or zip file?

tachibana-shin:
> can you provide a full sample app in a github repository or zip file?

Demo: https://stackblitz.com/edit/vitejs-vite-37psud?file=src%2Fmain.ts&terminal=dev

If you finished watching it tell me i need to delete it to clean up my stackbliz account

> This patch has been tested and works

tachibana-shin:
@jcesarmobile  anyway this error comes from a commit of your own: https://github.com/ionic-team/capacitor-plugins/commit/0a9f43dffd3815f600c35ed4528c017644fdb55e you patched `.rmdir()` in https://github.com /ionic-team/capacitor-plugins/commit/ae451aa08beb2138ecebdfcdd26101660aa00fde but you forgot `._copy()`

jcesarmobile:
stackblitz is not a valid reproduction neither, we need a project where we can test the changes

tachibana-shin:
> 

open the preview in a new page and press f122 you will see the error

jcesarmobile:
I know there is an error, I say I don't believe you, but we require a sample project for being able to verify that the fix fixes the issue, and for that we need a local copy of the project where we can install the modified version of the plugin

please, provide a sample app in github or zip file

jcesarmobile:
Thanks for the sample app
I noticed that after your changes, copy was still not working correctly as it was trying to write the file without encoding, which will result in writing it as base64, but since the original file was UTF8 it was failing as it's not valid base64.
I've added a check for detecting the encoding and passing UTF8 when needed

PR #1481, opened 2023-03-13
chore(local-notifications): remove deprecated call

markemer:
Move Deprecated `Bundle.get` call to it's own function and update.


jcesarmobile (changes requested):
`getString` is available since SDK 21, so there is no need of keeping the old code, can be safely replaced with getString for all versions

markemer:
> `getString` is available since SDK 21, so there is no need of keeping the old code, can be safely replaced with getString for all versions

Awesome! Yeah, I'll do that.

jcesarmobile (approved):

PR #1494, opened 2023-03-17
Update the MarkerClickCallbackData-doc

Raqhael:
The Documentation doesnt show the interface inheritance and its not clearly that you get informations like markerId & Co (from MarkerCallbackData) aswell.


jcesarmobile:
thanks for the PR, but sadly, this is automatically generated, so on next regeneration will be deleted, there are several issues with the google-maps docs, but we can't patch them manually, we should fix the doc generator to properly generate extends 

Issue #1408, opened 2023-02-07
Crash upon retrieving the limited library photos list (with 0 images selected)


pedromiba:
### Plugin(s)
@capacitor/camera

### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 4.6.3
  @capacitor/core: 4.6.3
  @capacitor/android: 4.6.3
  @capacitor/ios: 4.6.3

Installed Dependencies:

  @capacitor/cli: 4.6.2
  @capacitor/core: 4.6.2
  @capacitor/android: 4.6.2
  @capacitor/ios: 4.6.2

[success] iOS looking great! 👌
```

### Platform(s)
iOS

### Current Behavior
When selecting limited library permissions and giving access to 0 photos. 
If you attempt to retrieve the limited library photos list it will crash.

### Expected Behavior
It should just return an empty array.

### Code Reproduction
``` 
const limitedLibraryPhotos: GalleryPhotos = await Camera.getLimitedLibraryPhotos();
```

### Other Technical Details
Nothing to add

### Additional Context
![Error](https://user-images.githubusercontent.com/47518029/217211844-d355591d-1868-4e9b-bdef-194e99c358b8.png)

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

Please see the Contributing Guide for [how to create a Code Reproduction](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
Ionitron 💙

pedromiba:
As requested, here is the code to reproduce the reported issue:
[Github link](https://github.com/pedromiba/camera-plugin-issue)

Thanks in advance!

PR #1495, opened 2023-03-20
fix(camera): prevent iOS crash with 0 limited images selected

jcesarmobile:
closes https://github.com/ionic-team/capacitor-plugins/issues/1408

### DIFF
--- a/camera/ios/Plugin/CameraPlugin.swift
+++ b/camera/ios/Plugin/CameraPlugin.swift
@@ -109,19 +109,20 @@ public class CameraPlugin: CAPPlugin {
                         options.deliveryMode = .highQualityFormat
 
                         let group = DispatchGroup()
+                        if assets.count > 0 {
+                            for index in 0...(assets.count - 1) {
+                                let asset = assets.object(at: index)
+                                let fullSize = CGSize(width: asset.pixelWidth, height: asset.pixelHeight)
 
-                        for index in 0...(assets.count - 1) {
-                            let asset = assets.object(at: index)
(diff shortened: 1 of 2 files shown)


camera/ios/Plugin/CameraPlugin.swift
```diff
@@ -109,19 +109,20 @@ public class CameraPlugin: CAPPlugin {
...
-                            group.enter()
-                            imageManager.requestImage(for: asset, targetSize: fullSize, contentMode: .default, options: options) { image, _ in
-                                guard let image = image else {
+                        if assets.count > 0 {
+                            for index in 0...(assets.count - 1) {
+                                let asset = assets.object(at: index)
+                                let fullSize = CGSize(width: asset.pixelWidth, height: asset.pixelHeight)
+
+                                group.enter()
+                                imageManager.requestImage(for: asset, targetSize: fullSize, contentMode: .default, options: options) { image, _ in
```
markemer, line 118: We're ignoring the `info` dictionary of the `requestImage` handler, which could give us an error in the `PHImageErrorKey` key, if there is an error. I think in case of error, image is nil, so we don't have to worry about this code not working, but it might make sense to add it, or add a ticket to add it later.

jcesarmobile: you can create a feature request ticket for returning an error in case `PHImageErrorKey` contains something, but won't be doing this in this fix PR as is not related

markemer: Yeah - more a suggestion of future work, really. I'll create a ticket.

PR #1497, opened 2023-03-21
feat(device): allowing to get region from the device

bsalesc:
This PR will allow applications get the region/country from the device, allowing them select the right experience depending on which country the device is from.

On android it will get from the language but on iOS there is a defined field for it. Web will throw an error saying that isn't supported.

## Android
Screenshot 2023-03-21 at 15 48 10

## iOS
Screenshot 2023-03-21 at 15 48 45

### DIFF
--- a/device/README.md
+++ b/device/README.md
@@ -103,6 +103,20 @@ Get the device's current language locale code.
 
 --------------------
 
+### getRegionCode()
+
+```typescript
+getRegionCode() => Promise<GetRegionCodeResult>
+```
+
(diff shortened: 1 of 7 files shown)


jcesarmobile:
Thanks for opening this PR 🙏

Ionitron:
It looks like this issue didn't get the information it needed, so I'll close it for now. If I made a mistake, sorry! I am just a bot.

Have a great day!
Ionitron 💙

Issue #1100, opened 2022-07-28
Delete the Push FCM Token to get a new one


JustDoItSascha:
### Plugin
@capacitor/push-notifications

### Description
It would be nice if there is a option to delete the current token and get a new one from FCM. Because we an delete the token from the user server side, but there are circumstances where this could fail and then a user gets notifications even when he is logged out. So the option to get a fresh fcm token would be nice.

josuelmm:
+1

josuelmm:
Good day, in the plugin
https://www.npmjs.com/package/cordova-plugin-fcm-with-dependecy-updated
Exists a function for regenerate a new token. 
Is it possible to do the same with this plugin?

JustDoItSascha:
@josuelmm Did you try the deleteInstanceId() function in the plugin you posted? Because we did and it doesn't seem to work. The token from this plugin is also always the same.

josuelmm:
The function "deleteInstanceId" does not exist inside the plugin's definitions.d.ts

JustDoItSascha:
I mean from the plugin you posted, there is a `FCM.deleteInstanceId()` function. I think that is the function to regenerate the token you meant?

josuelmm:
Yes but it is an old plugin.

josuelmm:
Any help?

josuelmm:
I just made the modification in the source code and it works for me. I created a delete function and then run register again and it generated another token for me.

JustDoItSascha:
Awesome! Could you share modifications? I'm totally lost. Thanks!

josuelmm:
Check this =>
https://github.com/ionic-team/capacitor-plugins/pull/1110

sasos90:
Could this be merged to the original repo? I think this is a must have! :)

PR #1110, opened 2022-08-04
Refresh Token Push Notifications

josuelmm:
Refresh Token Push Notifications

PR #1498, opened 2023-03-21
feat(push-notifications): add unregister functions

markemer:
Closes: #1100, Related to PR #1110 which this supersedes. We can close that PR when this merges.


markemer:
> The new method needs to be added to the plugin interface in the definitions.ts file https://github.com/ionic-team/capacitor-plugins/blob/main/push-notifications/src/definitions.ts#L31
> 
> And also run `npm run build` so the method appears in the plugin docs
> 
> And would be good if the test app gets updated with a new button on the push notifications page to call the unregister method

Duh, yeah - I knew a forgot something - I was so preoccupied with thinking I forgot something on the native side, I forgot all the typescript.

giralte-ionic (approved):

push-notifications/ios/Plugin/PushNotificationsPlugin.swift
```diff
@@ -47,6 +47,16 @@ public class PushNotificationsPlugin: CAPPlugin {
...
     }
 
+    /**
+     * Unregister for remote notifications
+     */
+    @objc func unregister(_ call: CAPPluginCall) {
+        DispatchQueue.main.async {
+            UIApplication.shared.unregisterForRemoteNotifications()
+        }
+        call.resolve()
```
jcesarmobile, line 57: The resolve should be inside the `DispatchQueue.main.async` block

markemer: Sure, makes sense. Should `register`'s `resolve` also be in that block? I just copied the pattern.

jcesarmobile: yeah, it should, good catch!

jcesarmobile (approved):

Issue #1435, opened 2023-02-28
Device plugin: add Android SDK_INT


e6c31d:
### Plugin
Device plugin

### Description
We'd like to get the Android [`Build.VERSION.SDK_INT`](https://developer.android.com/reference/android/os/Build.VERSION#SDK_INT) in our Capacitor app. We can't use [`DeviceInfo.osVersion`](https://capacitorjs.com/docs/apis/device#deviceinfo) because it returns the `Build.VERSION.RELEASE` value, which is useless for comparing versions.

### Platform(s)
Android

### Preferred Solution
Add a new Android-only property to `DeviceInfo` that contains the value obtained from `SDK_INT`.

### Alternatives
[cordova-plugin-device](https://github.com/apache/cordova-plugin-device) has an Android-only property `sdkVersion`.

### Additional Context
When we raise the minimum Android version that our native Android app supports, we would like to show a message to users whose devices we no longer support. We can't use `Build.VERSION.RELEASE` because its documentation says:

> The user-visible version string. E.g., "1.0" or "3.4b5" or "bananas". This field is an opaque string. Do not assume that its value has any particular structure or that values of RELEASE from different releases can be somehow ordered.

PR #1505, opened 2023-03-27
feat(device): Add numerical SDK / OS Version

theproducer:
This PR makes available the Android SDK version / iOS version as an number.

closes: https://github.com/ionic-team/capacitor-plugins/issues/1435


giralte-ionic (approved):

device/ios/Plugin/DevicePlugin.swift
```diff
@@ -39,6 +39,7 @@ public class DevicePlugin: CAPPlugin {
             "model": modelName,
             "operatingSystem": "ios",
             "osVersion": UIDevice.current.systemVersion,
+            "iOSVersion": Double(UIDevice.current.systemVersion) ?? 0,
```
markemer, line 42: If the iOS Version is something like 16.4.1 this will return 0. We should probably return something more meaningful in that case. Like 16.41 or just 16 or 1641 as an integer.

markemer (changes requested):
Come up with a way to handle 3 part version numbers.

device/ios/Plugin/DevicePlugin.swift
markemer: One idea:
```
let number = "16.2.1"
let majorVersion = number.split(separator: ".")
let majorVersionInt = Int(majorVersion[0])
``

theproducer:
> If the iOS Version is something like **16.4.1** this will return 0...

Ah, good point!  One of the suggestions you left above would work - I'll also leave a note in the docs about it too.

markemer (approved):

IT-MikeS (approved):

Issue #1510, opened 2023-03-30
Device plugin: iOSVersion is inconsistent for 16.0 vs. 16.0.1 vs. 16.0.10


e6c31d:
### Plugin(s)
Device plugin

### Capacitor Version
This code was just committed recently and not released yet, but I thought I'd report the issue now so there would be no breaking change after release.
https://github.com/ionic-team/capacitor-plugins/blob/2ceb2965f3572068f22bad3e6d79c0902ec5c116/device/ios/Plugin/Device.swift#L82-L88

```
On iOS 16.0: 160
On iOS 16.0.1: 1601
On iOS 16.0.10 (hypothetical): 16010
```

### Platform(s)
iOS

### Current Behavior
The problem is that iOSVersion simply removes the dots and joins the integers together. This breaks comparison and sorting. For example, 16.0.1 would sort higher than 16.1 because 1601 > 161. It also ignores the fact that minor and patch versions may have more than one digit (It happened before, see: [iOS 4.2.**10**](https://en.wikipedia.org/wiki/IOS_version_history#4.2.10)).

### Expected Behavior
I suggest two changes:
1- Set minor and patch versions to 0 if they don't exist in systemVersion.
2- Pad minor and patch versions to two digits.

```
On iOS 16.0: 160000
On iOS 16.0.1: 160001
On iOS 16.0.10 (hypothetical): 160010
```

PR #1514, opened 2023-03-30
fix(device): Improving iOSVersion consistency

theproducer:
closes: https://github.com/ionic-team/capacitor-plugins/issues/1510


markemer (approved):
Honestly I think most people will use the string version, but this will work for `>` `<` aficionados.

IT-MikeS (approved):

PR #1524, opened 2023-04-05
docs(camera): update information about format on web

### LINKED ISSUE #1148: Camera giving me png when docs say it should be jpg
Description of the problem:

The docs say I should get a jpg back from capacitor: 

https://capacitor.ionicframework.com/docs/apis/camera/

> // The url starting with 'data:image/jpeg;base64,' and the base64 encoded string representation of the image, if using CameraResultType.DataUrl.
dataUrl ?: string;

When testing in the browser, I'm getting a png (e.g. data:image/png;base64,iVBO...). I need consistent or controllable behaviour of the image format.

Affected platform
- [ ] Android
- [ ] iOS
- [ ] electron
- [x] web

OS of the development machine
- [x] Windows
- [ ] macOS
- [ ] linux

Other information:

Capacitor version: 1.1.0

node version: 10.16.0

npm version: 6.9.0

Steps to reproduce:

Take a picture:

```
      const { Camera } = Plugins;

      const image = await Camera.getPhoto({
        quality: 90,
        allowEditing: false,
        resultType: CameraResultType.DataUrl,
        source: CameraSource.Prompt
      });
```

Put a breakpoint on the result, you'll see the data url is png.

ItsChaceD:
This PR clarifies that the web only supports `png` or `gif` and not `jpeg` files.
closes #1148


giralte-ionic (approved):

IT-MikeS (approved):

camera/src/definitions.ts
```diff
@@ -269,15 +270,15 @@ export interface GalleryPhoto {
...
+   * Web supports png and gif.
    *
    * @since 1.2.0
    */
   format: string;
 }
 export interface GalleryImageOptions {
   /**
-   * The quality of image to return as JPEG, from 0-100
+   * The quality of image to return, from 0-100
```
jcesarmobile, line 281: quality is only supported on android and ios, where it's always jpeg, so no need to delete it, but maybe add a note about it not being supported on web.

jcesarmobile (changes requested):
I've been testing and looks like I was wrong on my comment or things have changed since then, but I'm getting jpeg on chrome, while on safari and firefox I get png. 

So maybe mention that the format is dependent on the browser used?

jcesarmobile (changes requested):
For format, where it says "Web supports jpeg and png", it should also mention that depends on the browser.

jcesarmobile (approved):

PR #1535, opened 2023-04-14
fix(google-maps): Cast iOS marker id to string

jcesarmobile:


markemer (approved):

markemer:
This is around the Google Maps Utils too, I'd really love to update that.

PR #1537, opened 2023-04-15
Add Preferences configuration to specify iOS app group identifier

davidisaaclee:
Previously, the Preferences plugin would always use `UserDefaults.standard`. This PR adds an `iosAppSuite` configuration option to use an App Group's UserDefaults as described in https://developer.apple.com/documentation/foundation/userdefaults/1409957-init

### DIFF
--- a/preferences/ios/Plugin/Preferences.swift
+++ b/preferences/ios/Plugin/Preferences.swift
@@ -6,9 +6,11 @@ public struct PreferencesConfiguration {
     }
 
     let group: Group
+    let suiteName: String?
 
-    public init(for group: Group = .named("CapacitorStorage")) {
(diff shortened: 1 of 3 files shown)


preferences/ios/Plugin/Preferences.swift
```diff
@@ -6,17 +6,19 @@ public struct PreferencesConfiguration {
...
+        self.suiteName = suiteName
     }
 }
 
 public class Preferences {
     private let configuration: PreferencesConfiguration
 
     private var defaults: UserDefaults {
-        return UserDefaults.standard
+        return UserDefaults(suiteName: suiteName)!
```
davidisaaclee, line 21:
```suggestion
        return UserDefaults(suiteName: configuration.suiteName)!
```

eric-horodyski:

More context:

PR #1538, opened 2023-04-15
feat(statusbar): Navigation bar color on setBackgroundColor

qliqdev:
set Navigation Bar color also in Android


jcesarmobile:
Thanks for the PR, but the status bar plugin should only handle the status bar color. People might want to have different colors on the status bar and on the navigation bar.
There are 3rd party plugins for the navigation bar such as https://www.npmjs.com/package/@hugotomazi/capacitor-navigation-bar.
They can also be styled from the styles file if the user don't plan to change them programmatically and avoid using a plugin for that.

PR #1555, opened 2023-04-22
chore: fix variable naming

flogy:
The variable name of `firebaseMessagingVersion` seems to be wrong in the docs. Removed the `$` to avoid confusion.

### DIFF
--- a/push-notifications/README.md
+++ b/push-notifications/README.md
@@ -35,7 +35,7 @@ Android 13 requires a permission check in order to receive push notifications.  
 
 This plugin will use the following project variables (defined in your app's `variables.gradle` file):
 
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Thanks for the PR
We actually did that on purpose, variables are used with the `$`, but since users would be declaring the variables and not using them, then it makes sense to remove the `$` as for declaring them you don't have to use the `$`.

I was doing the changes on all other plugin READMEs and looks like I failed to get your branch because I had one with the same name, so when I tried to push to your branch I actually pushed to main with the changes
https://github.com/ionic-team/capacitor-plugins/commit/8c6deab6609daf9048fd3df989a18e1dd9c48b70

PR #1564, opened 2023-04-29
fixed typo from meow to now

michaelnjuguna:


jcesarmobile:
thanks for the PR, but I think it was done like that on purpose, not a typo

https://www.urbandictionary.com/define.php?term=Right%20Meow

Issue #1557, opened 2023-04-25
[@capacitor/google-maps] updateRender method in Android does not respect width and height changes


AE1NS:
### Plugin(s)
```
@capacitor/google-maps: 4.5.0
```

### Capacitor Version

```
@capacitor/cli: 4.7.0
@capacitor/core: 4.7.0
@capacitor/ios: 4.7.0
@capacitor/android: 4.7.0
```

### Platform(s)
Android

### Current Behavior
When I update the map container dimensions (width, height, x, y) and call 'updateMapBounds', the native map just get repositioned but not resized.

### Expected Behavior
The native map should also update its widht and height attributes.

### Reproduction
- Resize the html map wrapper, i.e. set a new height or width, with dev tools on the remote android device
- Call `updateMapBounds` on the plugin instance map object or alternatively call `window.dispatchEvent(new CustomEvent('resize'))`
- Now the map only updates it x and y values correctly but the size keeps in the initial size.

### Other Technical Details
Only the x and y values are updated on the map view in the implementation:
https://github.com/ionic-team/capacitor-plugins/blob/c39d84fbae464fc8984e26f970e85fd1862ff2ac/google-maps/android/src/main/java/com/capacitorjs/plugins/googlemaps/CapacitorGoogleMap.kt#L119-L123

AE1NS:
Replacing it with this, works:
```
CoroutineScope(Dispatchers.Main).launch {
        val bridge = delegate.bridge
        val layoutParams =
                FrameLayout.LayoutParams(
                        getScaledPixels(bridge, config.width),
                        getScaledPixels(bridge, config.height),
                )
        layoutParams.leftMargin = getScaledPixels(bridge, config.x)
        layoutParams.topMargin = getScaledPixels(bridge, config.y)
        this@CapacitorGoogleMap.mapView.layoutParams = layoutParams
}
```
But I dont know if it has any side effects.

AE1NS:
Or this:
```
CoroutineScope(Dispatchers.Main).launch {
        val mapRect = getScaledRect(delegate.bridge, updatedBounds)
        this@CapacitorGoogleMap.mapView.x = mapRect.left
        this@CapacitorGoogleMap.mapView.y = mapRect.top
        this@CapacitorGoogleMap.mapView.layoutParams.width = getScaledPixels(delegate.bridge, config.width)
[...]

PR #1569, opened 2023-05-03
fix(google-maps): Update dimensions in updateRender method

AE1NS:
This PR updates the width and height of the mapview in the updateRender method for Android, additionally to the x and y values.

Closes https://github.com/ionic-team/capacitor-plugins/issues/1557


AE1NS:
I updated the PR and wrapped a condition around it. So the layout gets only changed, when width or height has changed. This will at least add the future to update the dimension, allthough its a little bit laggy when used in combination with a draggable bottom sheet (live update of map dimension).

AE1NS:
Closed in favor of https://github.com/ionic-team/capacitor-plugins/issues/1594

PR #1576, opened 2023-05-04
chore(android): update android publishing for 5.0 and Gradle 8

carlpoole:


Steven0351:
Does this also need the `artifact("$buildDir/outputs/aar/${project.getName()}-release.aar")` update instead of using the `components` reference like capacitor did?

scripts/android/publish-module.gradle
```diff
@@ -33,7 +34,7 @@ afterEvaluate {
 
                 // Two artifacts, the `aar` (or `jar`) and the sources
                 if (project.plugins.findPlugin("com.android.library")) {
-                    from components.release
+                    artifact("$buildDir/outputs/aar/${project.getName()}-release.aar")
```
carlpoole, line 37: @Steven0351 Indeed :)

Steven0351: I totally missed this and I was even actively looking for it 😆

Steven0351 (approved):

scripts/android/publish-module.gradle
carlpoole: I blame the way Github ordered all the changes, although its just A-Z dir name. But still

IT-MikeS (approved):

PR #1577, opened 2023-05-04
fix(android): add appCompat libraries for maven releases

ItsChaceD:
This PR adds the `androidx.appcompat:appcompat` library to the plugins where it is missing (`filesystem` and `geolocation`) so that `assembleRelease` does not throw errors related to resource linking.

### DIFF
--- a/filesystem/android/build.gradle
+++ b/filesystem/android/build.gradle
@@ -1,6 +1,7 @@
 ext {
     capacitorVersion = System.getenv('CAPACITOR_VERSION')
     junitVersion = project.hasProperty('junitVersion') ? rootProject.ext.junitVersion : '4.13.2'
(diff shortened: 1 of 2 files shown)


filesystem/android/build.gradle
```diff
@@ -68,6 +69,7 @@ dependencies {
         implementation project(':capacitor-android')
     }
 
+implementation "androidx.appcompat:appcompat:$androidxAppCompatVersion"
```
carlpoole, line 72:
```suggestion
    implementation "androidx.appcompat:appcompat:$androidxAppCompatVersion"
```

carlpoole (approved):

IT-MikeS (approved):

Issue #1586, opened 2023-05-09
[@capacitor/google-maps] @googlemaps/js-api-loader: Refactor to importLibrary method


AE1NS:
### Plugin
@capacitor/google-maps: 5.0.0

### Description
Instead of the `load` method, the new `importLibrary` method should be used to import the google maps instance. This prevents interferences when the developer wants to use @googlemaps/js-api-loader in the main project with different libraries, i.e. to use the geometry API of google.

### Platform(s)
Web

### Preferred Solution
Instead of setting up the needed libraries in the constructor of the `loader` and calling the `.load()` method, its possible to leave the libraries in the loader empty and use the `.importLibrary()` method with the parameter `maps`. The result of the method will be the MapsLibrary.

### Additional Context
https://github.com/ionic-team/capacitor/discussions/6576

AE1NS:
Moved https://github.com/ionic-team/capacitor-google-maps/issues/22

PR #1587, opened 2023-05-09
feat(google-maps): @googlemaps/js-api-loader: Refactor to importLibrary method

AE1NS:
Closes https://github.com/ionic-team/capacitor-plugins/issues/1586

### DIFF
--- a/google-maps/package.json
+++ b/google-maps/package.json
@@ -93,8 +93,8 @@
     "access": "public"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


AE1NS:
This is a minor but effective adjustment. What do you think about it @capacitor devs? We are already using this in production for months and its the only way how we can use other google apis like geolocation or directions in parallel.

AE1NS:
Moved https://github.com/ionic-team/capacitor-google-maps/pull/21

PR #1592, opened 2023-05-16
Fix LocalNotificationSchema.schedule link to Schedule Type definition

Camotubi:
## Change

* Fix schedule link in LocalNotificationSchema to link to the Schedule type definition instead of the schedule method.

image


jcesarmobile:
Thanks for the PR, but that is automatically generated from the plugin types, so will be overridden when `npm run build` is run again.
This should be fixed in https://github.com/ionic-team/capacitor-docgen, which is what generates the plugin docs

Issue #1594, opened 2023-05-17
[@capacitor/google-maps] Update the size of the map


AE1NS:
### Plugin
@capacitor/google-maps

### Description
Currently only the x and y position of a map gets updated by the native updateRender methods. There should be the possibility to update also the size of the map, when the html container changed its size.

### Platform(s)
Android
iOS
Web

### Preferred Solution
There could be a ResizeObserver that listen to size changes of the map wrapper element and update the native map size.

### Additional Context
I created a demo for this topic:
https://github.com/AE1NS/capacitor-maps-resizing
When you click the 'draggable sheet' button, you will have a map combined with a draggable sheet at the bottom. As you see, the map size will not update when moving the sheet.
With https://github.com/ionic-team/capacitor-plugins/pull/1595 the feature is added to update the size of the map when moving the sheet in Android and iOS.

Issue #997, opened 2022-05-22
Google Map gets hidden after route change


warjohn123:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 3.5.1
  @capacitor/core: 3.5.1
  @capacitor/android: 3.5.1
  @capacitor/ios: 3.5.1

Installed Dependencies:

  @capacitor/android: 3.4.3
  @capacitor/ios: 3.4.3
  @capacitor/core: 3.4.3
  @capacitor/cli: 3.4.3

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)
```
iOS and Android
```

### Current Behavior
```
Awesome plugin. However I have an issue rendering the map. At first it works fine, but after changing routes and goes back to the route where map is shown, it won't load anymore. Btw, the map is shown in a tab
```

### Expected Behavior
```
Should not get hidden after route change
```

### Code Reproduction
```
const mapElement = document.getElementById('my-map');
    const mapConfig = {
      center: {
        lat: 33.6,
        lng: -117.9,
      },
      zoom: 8,
      androidLiteMode: false,
    };
    const mapOptions = {
      id: 'map',
      apiKey: MAP_APIKEY,
      config: mapConfig,
      element: mapElement,
    };
    const map = await GoogleMap.create(mapOptions);
```

muuvmuuv:
`setTimeout` works but feels very very wrong, especially if you have a lot of places, the tiles from Gmaps have to be loaded each time and you need some functionality to "destroy" the map properly. Our app crashed after several page back and forth transitions, I guess due to memory, even with `ngOnDestory` to destroy the existing map.

matteoerrera:
bump

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

Please see the Contributing Guide for [how to create a Code Reproduction](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
Ionitron 💙

AE1NS:
I reproduced this issue with Ionic on iOS.

Other than in Android, where the map is added natively behind the Webview, iOS can access the webview items and append UIViews.
iOS workflow:
- Search all subviews of webview for target container (capacitor-google-maps of type WKChildScrollView)
[...]

PR #1595, opened 2023-05-17
feat(google-maps): Update map size in updateRender method if it has changed

AE1NS:
In this PR I added the feature to update the map size when the html wrapper container changes its size. Its implemented with the help of a ResizeObserver which then calls the native `updateRender` methods.

Closes https://github.com/ionic-team/capacitor-plugins/issues/1594

### DIFF
--- a/google-maps/package.json
+++ b/google-maps/package.json
@@ -63,6 +63,7 @@
     "@ionic/eslint-config": "^0.3.0",
     "@ionic/prettier-config": "^1.0.1",
     "@ionic/swiftlint-config": "^1.1.2",
+    "@types/resize-observer-browser": "^0.1.7",
     "@types/supercluster": "^7.1.0",
     "downlevel-dts": "^0.7.0",
     "eslint": "^7.11.0",
--- a/google-maps/package.json
+++ b/google-maps/package.json
@@ -62,7 +62,6 @@
     "@capacitor/ios": "^5.0.0",
     "@ionic/eslint-config": "^0.3.0",
     "@ionic/prettier-config": "^1.0.1",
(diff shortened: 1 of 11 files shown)


AE1NS:
I extended the PR to fix https://github.com/ionic-team/capacitor-plugins/issues/997 as its also fixed with the help of the ResizeObserver I implemented here.

Its an adjustment where the map wasnt displayed, i.e. after an ionic back navigation. With the help of the ResizeObserver, a 'hidden' or 'show' state of the element can be tracked and a native method in iOS can fix the issue. Unfortunately I had to add a setTimeout of 700ms to get it work, which seems to be related to the page transition, but Im not sure, otherwise the map would still be not visible.

theproducer (approved):

markemer (approved):

google-maps/package.json
```diff
@@ -62,6 +62,8 @@
     "@capacitor/ios": "^5.0.0",
     "@ionic/eslint-config": "^0.3.0",
     "@ionic/prettier-config": "^1.0.1",
+    "@ionic/swiftlint-config": "^1.1.2",
```
IT-MikeS, line 65: This is not needed as the swift lint config was changed to use just references from the `swiftlint.config.js` file.

IT-MikeS (changes requested):

google-maps/package.json
AE1NS: Did something go wrong with the last merge commit @theproducer? I didnt touched this line.

theproducer: Possibly.  I'll remove it.

IT-MikeS (approved):

PR #1599, opened 2023-05-20
docs(google-maps): fix typo

jonz94:
According to:

https://github.com/ionic-team/capacitor-plugins/blob/dc5acf07d60293d53dbb33dbfb61620fc05572c6/google-maps/android/build.gradle#L93

The `kotlin_version` variable is being used as the version of `org.jetbrains.kotlin:kotlin-stdlib`, not `org.jetbrains.kotlin:kotlin-stdlib-jdk7`.


jcesarmobile (approved):
Thanks for the PR!

I removed the `-jdk7`, but looks like it was added back on a following PR that had merge conflicts and nobody noticed the change. Good catch!

Issue #1604, opened 2023-05-23
[@capacitor/google-maps] Data layer support


AE1NS:
### Plugin
@capacitor/google-maps

### Description
Google maps has a data layer feature, where it is possible to i.e. add a geoJson data object. They call it 'map feature'.
https://developers.google.com/maps/documentation/javascript/datalayer

### Platform(s)
Android
iOS
Web

### Preferred Solution
The following methods could be implemented in the first implementation:
- a method to add features
- a method to remove a feature
- a method to get the bounds of a feature

AE1NS:
Moved https://github.com/ionic-team/capacitor-google-maps/issues/20

PR #1605, opened 2023-05-23
feat(google-maps): Implemented map feature functions (i.e. GeoJSON support)

AE1NS:
This PR adds support (first implementation scoped to geoJson) for map features.

#### addFeatures(type: FeatureType, data: any, idPropertyName: string, styles: FeatureStyles): string[]
```
await map.addFeatures(
    FeatureType.GeoJSON,
    {
        type: 'Feature',
        geometry: GeoJSONData,
        properties: {
            id: 'my-identifier',
        },
    },
    'id',
    {
        ['my-identifier']: {
            strokeColor: '#123456',
            strokeOpacity: 0.8,
            strokeWeight: 1,
            fillColor: '#654321',
            fillOpacity: 0.1,
            geodesic: true,
        }
    }
)
```

#### getFeatureBounds(featureId: string)
`map.getFeatureBounds('my-identifier')`

#### removeFeature(featureId: string)
`map.removeFeature('my-identifier')`

Closes https://github.com/ionic-team/capacitor-plugins/issues/1604

### DIFF
--- a/google-maps/README.md
+++ b/google-maps/README.md
@@ -296,6 +296,9 @@ export default MyMap;
 * [`removeCircles(...)`](#removecircles)
 * [`addPolylines(...)`](#addpolylines)
 * [`removePolylines(...)`](#removepolylines)
+* [`addFeatures(...)`](#addfeatures)
+* [`getFeatureBounds(...)`](#getfeaturebounds)
+* [`removeFeature(...)`](#removefeature)
 * [`destroy()`](#destroy)
 * [`setCamera(...)`](#setcamera)
 * [`getMapType()`](#getmaptype)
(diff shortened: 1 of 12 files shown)


theproducer (changes requested):
I've added a page in the test app for testing out GeoJSON support, and grabbed a sample object from Google Map's documentation site.  It works on web, but doesnt work on iOS or Android (on Android it hard crashes).  From looking at it briefly it looks like its due a missing `idPropertyName` within the features.

If that's actually the issue I would rework it so that supplying a id property isn't needed, or, have it throw an appropriate error.

AE1NS:
I see. Seems that I only took the case with an id into account, because I only use it with styles :) You are absolutely correct, this should be an optional parameter.

Did you mean that you are going to rework it, or advised me to do so? I would be glad about support.

AE1NS:
I just pushed some more changes to make the _idPropertyName_ and _styles_ parameter optional. Looks good for me in Android. In iOS I just made the code adjustments but I am not able to test it right now. Would be great if you could have another look.

AE1NS:
Well, I think with the commit in August I broke it completely. Should be fixed now.

AE1NS:
Moved https://github.com/ionic-team/capacitor-google-maps/pull/19

PR #1614, opened 2023-05-23
fix(motion): example code fix to work on ios/android

dtarnawsky:
This resolves [https://github.com/ionic-team/capacitor-docs/issues/91](https://github.com/ionic-team/capacitor-docs/issues/91)

The DeviceMotionEvent.requestPermission API is only present on Safari and iOS. It also needs to be typecast to any in Typescript.

### DIFF
--- a/motion/README.md
+++ b/motion/README.md
@@ -16,7 +16,7 @@ permission before using this API. To request permission, prompt the user for
 permission on any user-initiated action (such as a button click):
 
 ```typescript
-import { Capacitor, PluginListenerHandle } from '@capacitor/core';
+import { PluginListenerHandle } from '@capacitor/core';
 import { Motion } from '@capacitor/motion';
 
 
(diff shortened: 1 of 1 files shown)


jcesarmobile (commented):
In the capacitor test app we do it like 
```
if (DeviceOrientationEvent !== undefined && typeof DeviceOrientationEvent.requestPermission === 'function')
```
That way it will still work on javascript as the cast to any code will not work in javascript.
While the block of code is marked as typescript, all other plugins code is javascript "compliant" too.

But we use that code to show or hide a button that's what will actually request the permission, and since this example code comes from the test app, it doesn't include the checks, because you can only get the button click if the button was displayed.

Also, while you have added a comment that says `// This should return a result of 'granted'`, the user could also have rejected the permission, so would be good to check the actual status and do something if not granted.

The test app page for reference:
https://github.com/ionic-team/capacitor-testapp/blob/main/src/pages/Motion.tsx

giralte-ionic (approved):

Issue #1617, opened 2023-05-25
[@capacitor/google-maps] Refactor options and extend a method to update the map options


AE1NS:
### Plugin
@capacitor/google-maps

### Description
This feature request has 2 topics

#### Refactor options
- As styles also exists in `google.maps.MapOptions` it could be removed from `GoogleMapsConfig`. I think you added it here, to show it in the documentation, but I would rather write a comment above the class, which properties from `google.maps.MapOptions` are supported
- To prevent overhead, the `MapType` class could be deleted, because its documented in `google.maps.MapOptions`
- The following methods could be removed and implemented within the options object (this would require to have a `update` method, see below):
   - getMapType
    - setMapType
    - enableIndoorMaps
    - enableTrafficLayer
    - enableAccessibilityElements
    - enableCurrentLocation
    - setPadding
- There should be some additional options in `GoogleMapsConfig` like:
    - isCompassEnabled
    - isRotateGesturesEnabled
    - isTiltGesturesEnabled
    - isToolbarEnabled
    - isZoomGesturesEnabled
- Some more properties of `google.maps.MapOptions` should be handled by the native code like:
    - gestureHandling
    - restriction

#### Add an update method for map options
The above defined changes should also be appliable by an update method, to change them after the map instantiation.

### Platform(s)
Android
iOS
Web

djabif:
Adding to this issue, I think it would be really nice to be able to update the map options as @AE1NS suggests.

In my case I need to set [setMyLocationButtonEnabled](https://developers.google.com/maps/documentation/android-sdk/reference/com/google/android/libraries/maps/UiSettings#public-void-setmylocationbuttonenabled-boolean-enabled)  to false. 

My problem is that if I set [enableCurrentLocation](https://capacitorjs.com/docs/apis/google-maps#enablecurrentlocation) to true, on Android the default geolocation button is shown, even if I set `disableDefaultUI: true` in the map config options. On iOS it works fine. 

<img width="437" alt="Screenshot 2023-08-27 at 22 05 53" src="https://github.com/ionic-team/capacitor-plugins/assets/1735814/45be1974-512a-424e-9c06-8059f2f34746">

matbravob:
The "disableDefaultUI" is only available for web.  For iOS and Android only the config options declared on GoogleMapConfig are available.

ngmiduc:
Are there any updates on this?

AE1NS:
[...]

PR #1618, opened 2023-05-25
﻿feat(google-maps): Refactored options and added feature to update the map options

AE1NS:
This refactorings simplify the map options handling and therefore remove many methods, that can now be set in the instantiation of the map but also be updated with a new method.

Closes https://github.com/ionic-team/capacitor-plugins/issues/1617

### DIFF
--- a/google-maps/ios/Plugin/CapacitorGoogleMapsPlugin.m
+++ b/google-maps/ios/Plugin/CapacitorGoogleMapsPlugin.m
@@ -7,6 +7,7 @@ CAP_PLUGIN(CapacitorGoogleMapsPlugin, "CapacitorGoogleMaps",
    CAP_PLUGIN_METHOD(create, CAPPluginReturnPromise);
    CAP_PLUGIN_METHOD(enableTouch, CAPPluginReturnPromise);
    CAP_PLUGIN_METHOD(disableTouch, CAPPluginReturnPromise);
+   CAP_PLUGIN_METHOD(update, CAPPluginReturnPromise);
    CAP_PLUGIN_METHOD(addMarker, CAPPluginReturnPromise);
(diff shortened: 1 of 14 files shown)


AE1NS:
I also fixed an issue with the padding values in Android, that were used as DP value. I used the `getScaledPixels` to convert the DP values to screen PX. Now its working fine.

google-maps/ios/Plugin/CapacitorGoogleMapsPlugin.m
```diff
@@ -19,13 +20,6 @@
    CAP_PLUGIN_METHOD(disableClustering, CAPPluginReturnPromise);
    CAP_PLUGIN_METHOD(destroy, CAPPluginReturnPromise);
    CAP_PLUGIN_METHOD(setCamera, CAPPluginReturnPromise);
-   CAP_PLUGIN_METHOD(getMapType, CAPPluginReturnPromise);
```
markemer, line 22: This is a breaking change - these should be kept, perhaps marked as deprecated.

markemer (changes requested):

google-maps/ios/Plugin/CapacitorGoogleMapsPlugin.m
AE1NS: https://github.com/ionic-team/capacitor-plugins/pull/1618/commits/ac752524925fd5ecd4f535a6294d3c3ebe31e647#diff-e887271a6e743973df748b0b83f29f5250ad28c088c82e1aacadfbb4ffc3d1b4R460
The method still exists in the map.ts with the same return type and is marked as deprecetd there. It was just removed from the native implementations because its replaced by the new solution.

AE1NS:
Moved https://github.com/ionic-team/capacitor-google-maps/pull/26

PR #1629, opened 2023-06-01
fix(push-notification): Notification not dispatching click when received in foreground

AndrewKravchenko:
Based on https://github.com/ionic-team/capacitor-plugins/pull/1478

Added generateUniqueId method to prevent overwriting the last notification


tafelnl:
I think you've got more chance of getting this merged, if you only make this about the `uniqueId` part. PRs should be as small as possible, and with the smallest scope possible

Issue #1633, opened 2023-06-02
Mapview overlays Webview at height:100%;


jonasphp:
### Plugin(s)
- Google Maps Plugin

### Capacitor Version
Latest Dependencies:

  @capacitor/cli: 5.0.4
  @capacitor/core: 5.0.4
  @capacitor/android: 5.0.4
  @capacitor/ios: 5.0.4

Installed Dependencies:

  @capacitor/core: 5.0.3
  @capacitor/ios: 5.0.3
  @capacitor/cli: 5.0.3
  @capacitor/android: 5.0.3

### Platform(s)
iOS

### Current Behavior
On height 100% the mapview overlays webview and makes it impossible to get dom elements to overlay. 

### Expected Behavior
DOM elements should be able to overlay the mapview.

ngmiduc:
I am also having this issue. As a workaround it use the height 99,99%.

Good that this will get fixed soon !

PR #1638, opened 2023-06-03
fix(google-maps): attach native map view to correct scroll view

theproducer:
Fixes an issue on iOS were the native map view would attach to the bridge webview scrollview instead of the intended `div` when the map container `div` is set to a width and height of 100%.

closes: https://github.com/ionic-team/capacitor-plugins/issues/1633


IT-MikeS (approved):

giralte-ionic (approved):

IT-MikeS (approved):

google-maps/ios/Plugin/Map.swift
```diff
@@ -99,25 +103,39 @@ public class Map {
...
+                    if isScrollView && !isBridgeScrollView {
                         (item as? UIScrollView)?.isScrollEnabled = true
 
-                        let isWidthEqual = round(Double(item.bounds.width)) == self.config.width
-                        let isHeightEqual = round(Double(item.bounds.height)) == self.config.height
+                        let height = Double((item as? UIScrollView)?.contentSize.height ?? 0)
+                        let width = Double((item as? UIScrollView)?.contentSize.width ?? 0)
+                        let actualHeight = round(height / 2)
+
+                        print("\(self.config.width) x \(self.config.height)")
```
markemer, line 119: Are these leftovers, or things we want logged?

NoMoreTickets:
This pull request is core-critical for anyone trying to use iOS to do anything at all important, including having a full-screen map. For the love of all that is good and righteous, can someone please merge this PR now and publish to NPM?

If this can't be done, can someone please write clear and comprehensive, end-to-end instructions on how to use a branch of capacitor plugins for one particular plugin, or provide a way to do this if there isn't one. I was able to get it to work and do what I wanted to, but not without a day and a half of rereading directions and experimentation.

google-maps/ios/Plugin/Map.swift
theproducer: They were just for debugging - our existing test app seems to be working fine with new code, and I've directed a few Discord users to test out this branch, so I think we are good to go now.

PR #1640, opened 2023-06-05
fix(splash-screen): crash if splash screen is missing

lubomirblazekcz:
If you choose not to use legacy splash screen below Android 12, the app crashes. This fixes the problem.

### DIFF
--- a/splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
+++ b/splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreen.java
@@ -339,13 +339,13 @@ public class SplashScreen {
                     }
                     imageView.setScaleType(config.getScaleType());
                     imageView.setImageDrawable(splash);
+
+                    splashImage.setFitsSystemWindows(true);
+                    
+                    if (config.getBackgroundColor() != null) {
+                        splashImage.setBackgroundColor(config.getBackgroundColor());
+                    }
                 }
-            }
-
-            splashImage.setFitsSystemWindows(true);
-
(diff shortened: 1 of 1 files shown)


jcesarmobile:
can you provide a sample app that crashes?

lubomirblazekcz:
You can reproduce it in vanilla capacitor app with splash-screen plugin. Just remove all splash images in `src/main/res` and splash reference in styles.xml

Since Android 12 and higher doesn't include any image spash screen, I decided to remove legacy splash screens completely in my app, but that caused this crash - `setFitsSystemWindows` was setting on `splashImage` that was `null`

jcesarmobile:
can you provide that sample app?

lubomirblazekcz:
Sure here is a sample app - https://github.com/lubomirblazekcz/capacitor-sample-app

And crash message that happens on API 29

```
FATAL EXCEPTION: main
Process: com.example.app, PID: 4508
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.example.app/com.example.app.MainActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.setFitsSystemWindows(boolean)' on a null object reference
```

jcesarmobile:
Thanks for the PR and the sample app.
Applying your changes would produce that the background color and setFitsSystemWindows would not be called in certain cases. I've sent a PR that would prevent the crash without affecting the plugin functionality.
https://github.com/ionic-team/capacitor-plugins/pull/1642

Anyway, if you want to remove the legacy splash screen entirely, you should set the launch activity's theme parent to `Theme.SplashScreen`, when not using that it will fallback to the old splash screen and that's why it crash, you are not using the new splash screen on Android 11 and older.

PR #1642, opened 2023-06-05
fix(splash-screen): Avoid crash when splash resources not found

jcesarmobile:
related to https://github.com/ionic-team/capacitor-plugins/pull/1640


IT-MikeS:
Works, but should we throw an error or maybe just a Log warning that the resources weren't found?

IT-MikeS (approved):

Issue #1644, opened 2023-06-07
[@capacitor/google-maps] ZIndex for marker in Android not set


AE1NS:
### Plugin(s)
@capacitor/google-maps

### Platform(s)
Android

### Current Behavior
The Android implementation has the zIndex not applied to the markerOptions. In web and iOS its set.

### Expected Behavior
Marker option 'zIndex' should also be set in Android.

### Code Reproduction
Web: https://github.com/ionic-team/capacitor-plugins/blob/main/google-maps/src/web.ts#L647
iOS: https://github.com/ionic-team/capacitor-plugins/blob/main/google-maps/ios/Plugin/Map.swift#L571
Android: https://github.com/ionic-team/capacitor-plugins/blob/main/google-maps/android/src/main/java/com/capacitorjs/plugins/googlemaps/CapacitorGoogleMap.kt#L755

jcesarmobile:
Thanks for the issue, we were already aware of it and tracking it internally
It was spotted here https://github.com/ionic-team/capacitor-plugins/pull/1474#pullrequestreview-1341696167

PR #1645, opened 2023-06-07
fix(google-maps): Apply zIndex for marker in Android and default to 0

AE1NS:
Closes https://github.com/ionic-team/capacitor-plugins/issues/1644


AE1NS:
Yes of course I can update it. I also was wondering about the nullable float.

jcesarmobile (approved):
thanks!

PR #1646, opened 2023-06-07
feat(google-maps): Update Google-Maps-Utils-iOS to 4.2.0

markemer:


markemer:
If anyone could explain why the heck the web build is failing even though I didn't touch it in the slightest, I'd very much appreciate it, because I'm confused.

markemer:
I mean, this isn't a breaking change exactly - This will solve people's workaround problems, but yeah - it'll require a `pod update` - not sure how to handle this. I mean, it shouldn't wait until Cap 6, but it can - Maps 8 is out and could be rolled into the change.

jcesarmobile:
Since the dependency is ~> 4.1 it should already pick 4.2 if users delete their Podfile.lock and do pod install (or npx cap update). New apps will also pick 4.2 automatically.
But since the Podfile.lock is committed now, it’s a breaking change for existing apps because it would cause issues.

giralte-ionic (approved):

markemer:
Closing this for now, re-evaluating maps for Cap 6

PR #1657, opened 2023-06-15
feat: use Firebase SDK for creating foreground notifications

### LINKED ISSUE #1368: Use default notification channel for foreground push notifications

### LINKED ISSUE #1388: Custom notification sound is not played while the app is in foreground

### LINKED ISSUE #1651: feat: make foreground notifications behave consistent across Android and iOS
### Plugin
push-notifications

### Description
Out of the box iOS foreground notifications work just fine. Android foreground notifications are for some reason not handled by the Firebase SDK itself. So every library using Firebase SDK has to implement their own logic. Shame on Firebase. However, that doesn't mean it shouldn't work consistently across iOS and Android.

I think the current implementation is too naive, not consistent with iOS and certainly not yielding expected results among developers.

Actually, I think it can even be marked as a bug, because the current implementation it just doesn't make a lot of sense to a lot of people. But I'll let that up to you.

### Platform(s)
Android

### Preferred Solution
I've just submitted a bunch of PRs that I think should be merged to make foreground notifications work like they should:

- https://github.com/ionic-team/capacitor-plugins/pull/1647
- https://github.com/ionic-team/capacitor-plugins/pull/1648
  - https://github.com/ionic-team/capacitor-plugins/pull/1504 (related documentation PR by someone else)
- https://github.com/ionic-team/capacitor-plugins/pull/1649
- https://github.com/ionic-team/capacitor-plugins/pull/1650

If you're going to merge these, probably some merge conflicts will arise. I will be more than happy to resolve them in due course.

Some of them have overlap with other open PRs, which would become obsolete when merging my PRs.

- https://github.com/ionic-team/capacitor-plugins/pull/1629 (obsolete because of 1650, the `uniqueId` part of this PR might still be helpful though)
- https://github.com/ionic-team/capacitor-plugins/pull/1478 (obsolete because of 1650)
- https://github.com/ionic-team/capacitor-plugins/pull/1324 (obsolete because of 1650)
- https://github.com/ionic-team/capacitor-plugins/pull/1423 (obsolete because of 1650 and 1649)

tafelnl:
Upon further inspection of the Firebase SDK, I came up with a way better solution than the combination of my previous PRs to solve #1651

Basically it does what the Firebase SDK itself does internally to show background notifications. I only utilized methods marked as public!

This will:

_Open issues:_
Fix #1651
Fix #1368
Fix #1388

_My 'old' PRs:_
Close #1650
Close #1649
Close #1648
Close #1647

_PRs by others:_
Close #1629
Close #1478
Close #1324
Close #1423


jahead:
Hey @markemer do we have a ETA when this might be release? it's currently blocking a app release?

jahead:
Hey not to spam here, but it seems our testers got it wrong; we think they were triggering a background notification display without realising it. 

This pr doesn't completely mimic the display notification class that the official firebase native Android SDK uses to display background notification. 

One biggest change is it doesn't handle the property `imageUrl`
Which is a step that happens before they create the notificationInfo. 

To add that feature though. It would be a bit more of a code change. So we have decided to fork the package so we customise the fire notification method. 

It's unfortunate since we mostly only need to modify the one method.

If the capacitor team is interested in accepting a pr to bring the foreground in line with how the background is handled I'm sure our client would be happy to share back. 

I will say though this did get us mostly to where we needed to be

push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/NotificationChannelManager.java
```diff
@@ -140,35 +137,4 @@ public void listChannels(PluginCall call) {
             call.unavailable();
         }
     }
-
-    /**
-     * Create notification channel
-     */
-    public void createForegroundNotificationChannel() {
```
theproducer, line 147: @jcesarmobile Do we still need this, or [is adding a default channel id ](https://firebase.google.com/docs/cloud-messaging/android/client#manifest)to the manifest enough?

theproducer (approved):

push-notifications/android/src/main/java/com/capacitorjs/plugins/pushnotifications/NotificationChannelManager.java
tafelnl: As far as I am aware, it's not needed anymore. The `CommonNotificationBuilder.getOrCreateChannel(` will take care of upserting the channel.

The Firebase SDK will execute similar logic as what's explain in #1649:

It will derive the `channelId` to be used for the foreground in the following order:

1. Firstly it will check if the incoming notification has a `channelId` set
2. Then it will check for a possible given value in the `AndroidManifest.xml`<sup>1</sup>
3. Lastly it will use the fallback `channelId` that the Firebase SDK provides for us. This channel will be created by the Firebase SDK upon receiving the first push message<sup>2</sup>

<sup>1</sup> From the [Firebase docs](https://firebase.google.com/docs/cloud-messaging/android/client#:~:text=(Optional)%20From%20Android,Manage%20notification%20channels):

> From Android 8.0 (API level 26) and higher, notification channels are supported and recommended. FCM provides a default notification channel with basic settings. If you prefer to create and use your own default channel, set `default_notification_channel_id` to the ID of your notification channel object as shown; FCM will use this value whenever incoming messages do not explicitly set a notification channel.

(see [README](https://github.com/firebase/quickstart-android/tree/master/messaging#set-default-channel) also)

<sup>2</sup> This is similar to how `react-native-push-notification` is handling this: https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationConfig.java#LL69C19-L69C50

I think this makes way more sense than the current solution. It will also automatically inherit the channel settings set for the incoming notification. So that's more expected behaviour for the developer

---

It would make sense to explain all this to the user of course. I did add some stuff to the docs in my other PR (https://github.com/ionic-team/capacitor-plugins/pull/1649/files#diff-09c58e1cc87af661e7a9d0081a9360efcd3da00a53911c3f6b998a66168ea170). Honestly, I don't know why I didn't add that to this PR. Probably just forgot about it. I can add it if you want?

theproducer: Yes, I would add that note about the default notification channel in case users needed to define and use their own.  Thank yoU!

tafelnl: Added! I tried my best to explain it as clear as possible.

ItsChaceD (approved):

PR #1672, opened 2023-06-30
fix(google-maps): Add missing listeners to removeAllMapListeners()

NikitaZelenskis:
Some listeners are missing from `removeAllMapListeners()` which results in them being called even after calling `removeAllMapListeners()`.


jcesarmobile (approved):
good catch, thanks!

PR #1678, opened 2023-07-06
fix(local-notifications): make requestPermissions resolve if granted

jcesarmobile:
like https://github.com/ionic-team/capacitor-plugins/pull/1677 but for local notifications plugin


markemer (approved):
Still have to resolve the permissions if granted, huh? TIL.

ItsChaceD (approved):

PR #1680, opened 2023-07-09
Fixing Schedule interface link in docs

jlucartc:
Link was pointing to `schedule()` method instead of Schedule interface.


jcesarmobile:
thanks for the PR, but this is auto generated, so it will be reverted on next build, should be fixed on the doc generator 

PR #1683, opened 2023-07-13
chore: update lerna and fix full-publish

IT-MikeS:


jcesarmobile (commented):
In capacitor's lerna.json we have `"useWorkspaces": true,`

https://github.com/ionic-team/capacitor/blob/main/lerna.json#L2

Since this PR seems to change to use that, shouldn't we add that line here too?

IT-MikeS:
> In capacitor's lerna.json we have `"useWorkspaces": true,`
> 
> https://github.com/ionic-team/capacitor/blob/main/lerna.json#L2
> 
> Since this PR seems to change to use that, shouldn't we add that line here too?

It was set as default in 7.x of lerna so its removed now, also did the removal in my Cap PR

PR #1684, opened 2023-07-14
fix(browser): anchor events links

Jorgagu:
Fix anchors links to events because there need a hyphen at the end to point to the good section of the markdown file


jcesarmobile:
Thanks for the PR, but that's auto generated, so it will be reverted on next doc generation run.
Should be fixed in the doc generator

PR #1688, opened 2023-07-16
docs(geolocation): use more specific return type for watchPosition

manaust:


jcesarmobile:
Thanks for the pull requests, but method documentation is generated from the types automatically, so this change will be reverted on next generation.
It should be fixed on the doc generator instead

Issue #1693, opened 2023-07-19
Get app specific language


philet:
### Plugin
@capacitor/device

### Description
We have an app that supports the three languages German, Italian and French (but not English):

<img width="629" alt="Bildschirmfoto 2023-07-19 um 10 36 27" src="https://github.com/ionic-team/capacitor-plugins/assets/3964328/7cc26b60-1148-4843-abfa-8574ab915d9f">

A user has set the order of preferred languages on his device to: English, Italian, German (see left screenshot below).

Therefore he would like to see our app in **Italian** (see app specific language settings in right screenshot below).

<img width="300" alt="iOS language settings" src="https://github.com/ionic-team/capacitor-plugins/assets/3964328/c5e274d5-d480-47bf-96f3-07e17c41d381">

<img width="300" alt="App specific language settings" src="https://github.com/ionic-team/capacitor-plugins/assets/3964328/b5af04f6-d853-4f85-92f5-55f940264bb2">

On iOS @capacitor/device getLanguageCode() returns 'en' in this case. (On Android 'it' is returned.)

On iOS there is no other method on @capacitor/device which we can call to get the preferred language 'it'.

Sample project illustrating the problem:
https://github.com/philet/capacitor-app-language-example

### Platform(s)
iOS
(Android, Web for API parity.)

### Preferred Solution
Add a method 'getAppLanguageCode' to @capacitor/device.

The following method returns the correct application language code ('it' in this case):

#### iOS

```
    public func getAppLanguageCode() -> String {
      
        // According to https://developer.apple.com/news/?id=u2cfuj88, the current
        // app language is returned by 'Bundle.main.preferredLocalizations.first'.
        // Fallback to device language if app language is not defined.
      
        let appLanguageWithFallbackToDeviceLanguage =
            Bundle.main.preferredLocalizations.first ?? Locale.preferredLanguages[0]
        return String(appLanguageWithFallbackToDeviceLanguage.prefix(2))
    }
```

#### Android

On Android @capacitor/device getLanguageCode() already returns the app language. Therefore the 'getAppLanguageCode' implementation could just call 'getLanguageCode':

```
@PluginMethod
public void getAppLanguageCode(PluginCall call) {
    getLanguageCode(call);
}
```

### Alternatives
-

### Additional Context
In addition to adding `getAppLanguageCode`, the existing `getLanguageTag` method should probably also be supplemented by a new `getAppLanguageTag` method.

jcesarmobile:
[...]

PR #1694, opened 2023-07-19
feat(app): Add 'getAppLanguageCode' and 'getAppLanguageTag'

philet:
Solves https://github.com/ionic-team/capacitor-plugins/issues/1693


app/src/definitions.ts
```diff
@@ -130,6 +130,24 @@ export type URLOpenListener = (event: URLOpenListenerEvent) => void;
...
+export interface GetLanguageCodeResult {
+  /**
+   * Two character language code.
+   *
+   * @since 5.1.0
+   */
+  value: string;
+}
+
+export interface LanguageTag {
```
IT-MikeS, line 142: And this one to `AppLanguageTag` for the same reason as above.

IT-MikeS (changes requested):

jcesarmobile:
I've addressed the changes requested by Mike and added two buttons on the test app for testing
https://github.com/ionic-team/capacitor-testapp/pull/628

But I've noticed that `getAppLanguageTag()` is also returning a 2 language code on iOS instead of the documented "IETF BCP 47 language tag".
`Bundle.main.preferredLocalizations.first` will return a matching code on `CFBundleLocalizations` if there is one, which is usually a 2 letter code, or if there isn't one that matches, then it will return the `CFBundleDevelopmentRegion`, which is a 2 or 3 letter ISO code. So I don't think the `getAppLanguageTag` method is needed, and also, I don't think the `getAppLanguageCode` should return a 2 letter code as it could also be a 3 letter ISO code.
I think it should just be `getAppLanguage()` and return the whole value of `Bundle.main.preferredLocalizations.first`

IT-MikeS (changes requested):
Please see jcesarmobile's comment above.

ThatOneCalculator:
Any update on this?

reslear:
@jcesarmobile will be good

eric-horodyski:
Thanks for opening this PR 🙏

reslear:
still actual

PR #1737, opened 2023-08-15
Improve logging when permissions are not granted

IDisposable:
Trying to diagnose why permissions aren't working right (and thus the plugin crashes), would help to know what `alias` is being used, so add that to the logged message on failure


eric-horodyski:

More context:

IDisposable:
Any particular reason you didn't just merge this one-line change for 12 months?

Issue #1726, opened 2023-08-10
[google-maps] MapType.Normal doesn't work on web


ngmiduc:
### Plugin(s)
 `@capacitor/google-maps": "^5.3.0"`

### Platform(s)
Web

### Current Behavior
When changing map types on web, the type `MapType.Normal` doesn't work on web. It works on native devices but not on web sites.

### Expected Behavior
The map type on the web platform can be set to normal.

### Code Reproduction
https://stackblitz.com/edit/angular-yz24zb?file=src%2Fapp%2Fexample.component.ts

1. Set google_api_key in  `example.component.ts` line21:  `const apiKey = 'API KEY';`

jcesarmobile:
please, can you provide a github repository or zip file containing a full sample app?
stackblitz urls are not valid reproductions as we use the reproductions to verify the fix

ngmiduc:
> please, can you provide a github repository or zip file containing a full sample app? stackblitz urls are not valid reproductions as we use the reproductions to verify the fix

I think you can just download the code and files from stackblitz.
Here the files:
[angular-yz24zb.zip](https://github.com/ionic-team/capacitor-plugins/files/12359127/angular-yz24zb.zip)

jcesarmobile:
thanks, I'm not familiar with stackbliz and didn't see the download button

PR #1738, opened 2023-08-16
fix(google-maps): problems on get/setMapType on web

jcesarmobile:
`setMapType` is not working for `MapType.Normal` because `Normal` doesn't exist on google maps and we replace it with `roadmap`, but the check where we compare with `Normal` is not working because we compare after the lowercase. 
So changed `setMapType` to compare the `MapType.Normal` with the non lowercased value.

Also noticed that `getMapType` was returning the google maps values, which are lowercased, so they don't match with the types and return undefined, so changed them to return our values with the first letter capitalized.

closes https://github.com/ionic-team/capacitor-plugins/issues/1726


ItsChaceD (approved):
Looks good to me!

_Note for anyone else testing on the sample app provided in the issue, I had to use [ncu -u](https://www.npmjs.com/package/npm-check-updates) to force update all of the dependencies and get the app running._

IT-MikeS (approved):

PR #1740, opened 2023-08-17
Android: Specify export behavior for broadcast receives

saschpe:
Required for Android 14 support. See https://developer.android .com/about/versions/14/behavior-changes-14


jcesarmobile:
Thanks for the pull request, but for the Network plugin that code is only reachable on Android M and older, so the change is not needed.
And for Share plugin, while it fixes the initial app crash if the plugin is present, it doesn't fix the crash when you try to use the plugin, so I'm working in a separate PR that fixes both problems.

jcesarmobile:
If you don't want that we "messed with" (make changes) in your pull requests there is a checkbox you can uncheck (or check, don't remember exactly) to not give us push access to your pull request, in that case I would have asked you to rebase instead of doing it myself.
But the rebase is a simple click for me on the GitHub UI and makes the tests run, if you rebase the tests don't run and I wouldn't have seen lint failing, so I wouldn't have tell you to run fmt command neither.

We are currently working on releasing Capacitor 6 and modifying the share plugin to work on Android 14 is one of our internal tasks, so we can't wait too long for users to make the requested changes, specially since I've already been looking into the required changes for share to work and they are slightly complex.

Issue #1708, opened 2023-06-29
feat: Make 'id' optional when creating a Google Map - Native Static and Dynamic Maps (without id) are charged 0.00 USD


amymarsh4work:
### Description
According to [Google's Billing documentation for maps](https://developers.google.com/maps/billing-and-pricing/pricing#maps-product-skus), we would NOT accrue charges when using the Google Maps SDKs for Native Static and Dynamic maps, IF we created our maps without an ID. Currently, the 'id' property is required for the CreateMapArgs type. Making the id optional could result in lowering costs.

### Platform(s)
Only applies to Android and iOS. When a dynamic map is created with the Maps JavaScript API, it results in charges regardless.

### Preferred Solution
Being able to call this code without an id:
async createMap() {
   this.googleMap = await GoogleMap.create({
      id: 'myFavMap', // <== this is optional
      element: this.mapRef.nativeElement,
      apiKey: environment.mapsApiKey,
      config: this.config
    });
}

### Alternatives
I tried setting id to empty string, and tested. This is what happens on the platforms:
Browser: the map gets created
Android: ‘Error: Missing or invalid map id.’ and the map is not created
iOS: No error. Logs say the map is created, but I don’t see the map. I wasn’t seeing the map before either, so this is likely due to a separate problem.

We have also implemented our map features with @angular/google-maps. It works, but we would be charged, since it wraps the Maps API for JavaScript.

### Additional Context
Our app has relatively basic needs for maps functionality, but map creation is somewhat frequent. Example: displaying a map on detail pages of all buildings, bus stops, parking lots, dining locations at a large university. A couple features require multiple markers and drawing polylines. The dynamic maps functionality has met all of our requirements. Our use-case seems like it would be pretty common.

markemer:
I moved it - everything should be cool now. I think this is a new-ish feature, I don't remember it being possible to avoid charges before.

PR #1750, opened 2023-08-22
feat(google-maps): Separate mapId for Google Maps Cloud IDs

theproducer:
This PR adds support for a separate optional `mapId` that is used for cloud based map styling and features:
https://developers.google.com/maps/documentation/get-map-id

closes: https://github.com/ionic-team/capacitor-plugins/issues/1708

### DIFF
--- a/google-maps/android/src/main/java/com/capacitorjs/plugins/googlemaps/CapacitorGoogleMapsPlugin.kt
+++ b/google-maps/android/src/main/java/com/capacitorjs/plugins/googlemaps/CapacitorGoogleMapsPlugin.kt
@@ -11,6 +11,7 @@ import com.getcapacitor.annotation.CapacitorPlugin
 import com.getcapacitor.annotation.Permission
 import com.getcapacitor.annotation.PermissionCallback
 import com.google.android.gms.maps.MapsInitializer
+import com.google.android.gms.maps.OnMapsSdkInitializedCallback
 import com.google.android.gms.maps.model.LatLng
 import com.google.android.gms.maps.model.LatLngBounds
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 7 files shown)


IT-MikeS (approved):

jcesarmobile (commented):
It looks good, but should we use different config options for Android and iOS?
the `mapId` is unique per platform, so having a single `mapId` option will mean users will have to detect the platform and pass the id for the specific platform, a single `mapId` can't be used, it will error if you try to use a javascript map id in Android (on iOS it surprisingly works, but it shouldn't according to the docs)

giralte-ionic:
I agree with Julio that the platform should be part of the params passed here to avoid the problem. I also don't believe we should assume that iOS will always accept a javascript platform id.

giralte-ionic (approved):

google-maps/android/src/main/java/com/capacitorjs/plugins/googlemaps/CapacitorGoogleMapsPlugin.kt
```diff
@@ -90,6 +92,13 @@ class CapacitorGoogleMapsPlugin : Plugin() {
         )
     }
 
+    override fun onMapsSdkInitialized(renderer: MapsInitializer.Renderer) {
+        when (renderer) {
+            MapsInitializer.Renderer.LATEST -> Log.d("Capacitor Google Maps", "Latest Google Maps renderer enabled")
+            MapsInitializer.Renderer.LEGACY -> Log.d("Capacitor Google Maps", "Legacy Google Maps renderer enabled - Cloud based map styling and advanced drawing not available")
```
jcesarmobile, line 98: Use `Logger.debug` instead of android Log class so it can be globally disabled

jcesarmobile (approved):

PR #1757, opened 2023-08-24
getRawGoogleMapInstance

ThomasWestrelin:
This PR adds a getRawGoogleMapInstance method that exposes the underlying google.maps.Map instance when on web platforms.

This is especially useful when one wants to access specific features that are only available in the Google Map JS SDK or that haven't been implemented by this plugin. This provides an escape hatch that allows users to access all JS SDK features when those are only needed on web platforms, the only current workaround would be conditionally using this plugin to instead use the raw JS SDK on web platforms.

 

Context

We are migrating an existing app to Capacitor and are using advanced features from the JS SDK that cannot be implemented with the current interface. We don't need those advanced features in the native version of our app se we are fine to have them work only on the web.

### DIFF
--- a/google-maps/README.md
+++ b/google-maps/README.md
@@ -301,6 +301,7 @@ export default MyMap;
 * [`setCamera(...)`](#setcamera)
 * [`getMapType()`](#getmaptype)
 * [`setMapType(...)`](#setmaptype)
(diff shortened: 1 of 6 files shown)


eric-horodyski:

More context:

PR #1762, opened 2023-08-25
chore: force typescript 4.1

jcesarmobile:
Despite we have `"typescript": "~4.1.5"` in all plugin's package.json, 5.2.2 is being installed in the root and causing lint issues because typescript 5.2 removed something that we don't use but something we use might be using.

This PR adds `"typescript": "~4.1.5"` entry in the root package.json so it uses that version for all packages and lint gets fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
     "nx": "^16.3.1",
     "prettier": "~2.3.0",
     "prompts": "^2.3.2",
-    "glob": "^10.3.3"
+    "glob": "^10.3.3",
+    "typescript": "~4.1.5"
   },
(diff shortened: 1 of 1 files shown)


markemer:
Wait, but doesn't capacitor itself use Typescript 5?

jcesarmobile:
capacitor 5 was updated to use typescript 5.0.x, but capacitor plugins weren't updated

markemer:
> capacitor 5 was updated to use typescript 5.0.x, but capacitor plugins weren't updated

Cool, I forget we can mix like that.

markemer (approved):

giralte-ionic (approved):

PR #1777, opened 2023-09-03
feat(push-notifications): Helper for opening native notification settings

MegaMaddin:
Add a helper for opening the native notification settings, e.g. for the case when user declined push permissions and need to be routed to the native settings to enable them again.

I'm aware that there are external plugins for achieving a similar result. I just think that it would benefit users, when not being enforced to install another plugin to be able to cover most of a frontend push notification journey.


MegaMaddin:
Could someone from the core team please have a look? @jcesarmobile

MegaMaddin:
@jcesarmobile Q: what is the proper process of getting PRs reviewed? I've searched for this in the contribution guides, but wasn't able to find anything. Some pointers would be appreciated.

eric-horodyski:
Thanks for opening this PR 🙏

PR #1785, opened 2023-09-08
Broken links fixed in push-notifications README.md

bartan34:
Mistyped hyperlinks fixed in push-notifications documentation API list


jcesarmobile:
thanks for the PR, but those links are auto generated, so should be fixed in the library that generates the links, changing them manually will just result in them being overwritten next time the readme is generated

PR #1802, opened 2023-09-20
chore: fix ci latest publish trying to publish all plugins

IT-MikeS:
Current npm script tries to publish each plugin even if they aren't needed. This PR fixes that by using Lerna with provenance turned on via `.npmrc`

See https://github.com/lerna/lerna/issues/3657#issuecomment-1535284598 for example/info

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 package-lock=false
+provenance=true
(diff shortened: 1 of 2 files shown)


giralte-ionic (approved):

theproducer (approved):

jcesarmobile:
since main branch is for capacitor 6, shouldn't the alpha/beta/etc. scripts be fixed too?

jcesarmobile (approved):

jcesarmobile (approved):

PR #1803, opened 2023-09-20
chore: fix ci latest publish trying to publish all plugins (5.x)

IT-MikeS:
See: https://github.com/ionic-team/capacitor-plugins/pull/1802

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 package-lock=false
+provenance=true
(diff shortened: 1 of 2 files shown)


giralte-ionic (approved):

jcesarmobile (approved):

jcesarmobile:
When I requested changes for alpha/beta/rc I mean on main branch only since we are not releasing alpha/beta/rc of capacitor 5

IT-MikeS:
> When I requested changes for alpha/beta/rc I mean on main branch only since we are not releasing alpha/beta/rc of capacitor 5

Figured we might as well make them match, just in case

PR #1805, opened 2023-09-21
improve documentation of status-bar

cprodhomme:
i add an example of the usage `backgroundcoloroptions` because it will help to know that your value need to include `#` symbol

Capture d’écran 2023-09-21 à 12 58 29

### DIFF
--- a/status-bar/README.md
+++ b/status-bar/README.md
@@ -192,9 +192,9 @@ This method is only supported on Android.
 
 #### BackgroundColorOptions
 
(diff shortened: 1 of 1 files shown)


jcesarmobile:
thanks for the pull request, but that information is generated from the definitions.ts file, so we can't make manual changes like this as they will be overwritten on next build.
The proper way of doing this would be to add an `@example "#ffffffff"` in the definitions.ts, but just tested and the docs generator doesn't support it on interfaces.

Issue #1667, opened 2023-06-22
[help]: google maps remote icon markers on main thread


ngmiduc:
### Plugin(s)
`"@capacitor/google-maps": "^5.1.2"`

### Capacitor Version
```
"@capacitor/app": "^5.0.0",
"@capacitor/core": "^5.0.0",
"@capacitor/google-maps": "^5.1.2"
```

### Platform(s)
IOS

### Current Behavior
I have integrated the capacitor google maps plugin into our Ionic application. When I add markers with remote icon, it throws a warning inside xcode. Should this be addressed?

```
2023-06-22 14:15:22.889158-0400[47947:1893573] [CFNetwork] Synchronous URL loading of https://3776657.fs1.hubspotusercontent-na1.net/hub/3776657/hubfs/ionic-icon-3.png should not occur on this application's main thread as it may lead to UI unresponsiveness. Please switch to an asynchronous networking API such as URLSession.
2023-06-22 14:15:22.976128-0400[47947:1893573] [CFNetwork] Synchronous
```

### Expected Behavior
Should this warning not exist?

### Additional Context
Running in IOS / Xcode emulator with an Iphone 14, IOS

Ionitron:
This issue needs more information before it can be addressed.
In particular, the reporter needs to provide a minimal sample app that demonstrates the issue.
If no sample app is provided within 15 days, the issue will be closed.

Please see the Contributing Guide for [how to create a Sample App](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
Ionitron 💙

ngmiduc:
Reproduction:
[issue-google-maps-remote-marker.zip](https://github.com/ionic-team/capacitor-plugins/files/12426047/issue-google-maps-remote-marker.zip)


Steps:
1. install npm packages and ionic/cli
2. add google maps key to `src/app/home/home.page.ts::19`
3. run `ionic capacitor build ios`
4. build xcode app to an emulator device
5. run app
6. see warning in emulator debug logs

The IOS app will freeze until all remote markers are loaded.

Error:

`App[7023:75878] [CFNetwork] Synchronous URL loading of https://png.pngtree.com/png-vector/20190115/ourmid/pngtree-vector-location-icon-png-image_317888.jpg should not occur on this application's main thread as it may lead to UI unresponsiveness. Please switch to an asynchronous networking API such as URLSession.`

PR #1818, opened 2023-09-27
fix(google-maps): use URLSession for remote iconUrl on iOS

jcesarmobile:
closes https://github.com/ionic-team/capacitor-plugins/issues/1667


markemer (approved):
This is a good change. `Data` is mainly only good for file urls and even apple says:

> As this method runs synchronously and blocks the calling thread until it finishes, don’t invoke it from the main thread.

I'm actually a little surprised this ever worked.

IT-MikeS (approved):

PR #1833, opened 2023-10-06
feat!(camera): Use Android Photo Picker instead of ACTION_PICK intent

jcesarmobile:


carlpoole:
> Setting it as breaking because the component change might cause unknown behaviour changes. Also because it no longer requires permissions for picking pictures from the gallery on Android.
> 
> I have not tested the changes in a fragment, so not sure if somebody with a sample app that uses fragments could test it and make sure this code is correct https://github.com/ionic-team/capacitor-plugins/pull/1833/files#diff-40c8474750f8a1fbb312900f646eca6a86067853007cdfe292b428bf0e48bddfR322-R334 I made it similar to how `fragment.registerForActivityResult` works.

@jcesarmobile tested this in the Portals Ecommerce App and it seems to be working as expected.

image

ItsChaceD (approved):

IT-MikeS (approved):

bhandaribhuminpfizer:
Are there any plans to merge this PR or any tentative date?

webfletch:
> Setting it as breaking because the component change might cause unknown behaviour changes. Also because it no longer requires permissions for picking pictures from the gallery on Android.
> 

This does cause unknown behaviour changes and various differences to how it works depending on the Android version and device being used, including breaking the functionality for selecting multiple images on some devices, as outlined in the following issue: #2199 

Due to the amount of uncertainty around how different devices/versions would be affected by this change, I think this would have benefitted from having an option for "UseLegacyMode" to allow people to not be forced to use the Android Photo Picker and continue using the previous image picker. 

Due to the issues outlined above it means that we cannot upgrade to this version of Capacitor until we change to use a different image picker that provides a consistent interface and functionality across different Android versions.

jcesarmobile:
Sadly [Google is requiring](https://support.google.com/googleplay/android-developer/answer/14115180) to use the Photo Picker for apps whose main purpose isn't handling images, so there was no option for us to keep backward compatible code as it could lead to app rejections if the code was detected even if not used by all users.

And sadly the Photo Picker is half backed, same thing happened when they forced to use the Android 12 Splash Screen API.

You can try updating the `androidxActivityVersion` variable in `variables.gradle` to `1.10.0-alpha02` and see if it solves your problems, it supposedly has fixed several problems with 3rd party pickers.

If you still find issues you can [report them to Google](https://source.android.com/docs/setup/contribute/report-bugs)

The deadline to not use the Photo Picker already expired, they started asking for additional information when requesting the photos permission and you have to send the information before October 31, 2024.
You can ask for an extension and if they grant it you could still use the old API until January 22, 2025, but then you'll have to use the Photo Picker.

webfletch:
@jcesarmobile Thank you for the detailed explanation as to the reasons you were forced to make this change. 

We tried `1.10.0-alpha02` and it resolved some issues (i.e. multiple select worked on Android 9 again) but still didn't fix the strange issue of a completely different photo picker being used on a Pixel 4 (Android 12) emulator.

As you mentioned it seems that the Photo Picker does need a lot of work before it is fit for purpose on all devices, there are already a lot of reported issues with Google I found from following your link.

Our App is one that has been approved as having a main purpose of handling images so we are not forced to use the Photo Picker, therefore, I think the best solution for us for the time being is going to be to use a forked version of Capacitor/Camera where we have reverted back the changes forcing the use of Photo Picker, then just wait until they have brought out the relevant updates to fix the backwards compatibility issues or wait until our userbase have updated their devices to newer versions.

PR #1840, opened 2023-10-11
feat(local-notifications): Adding permissions for using SCHEDULE_EXACT_ALARM in Android 14

theproducer:
As a result of [changes in Android 14](https://developer.android.com/about/versions/14/changes/schedule-exact-alarms), this PR makes using exact alarms for scheduling local notifications optional, and provides functions for checking and requesting permissions for using exact alarms.


IT-MikeS (approved):

jcesarmobile (changes requested):
The plugin already handles the exact notifications being optional, the only difference was that on Android 13 is granted by default if the permission is present in the `AndroidManifest.xml` and on Android 14 is denied by default even if the permission is present, so the two methods for requesting the permission and checking its value are needed, but not the extra `exact` parameter. If the user adds the permission to the manifest I think that's enough indication that the want to use exact notifications without having to pass an extra parameter.

`SCHEDULE_EXACT_ALARM` is not a runtime permission, so it shouldn't be added to the alias, having permissions in the alias that are optional will result in errors for users that don't have the permission in the manifest.

Since it's not a runtime permission and the user won't be prompted but will be redirected to the notification settings page, not sure if the naming should be different to not confuse users, something like `checkExactNotificationSetting` and `changeExactNotificationSetting`

Don't use `android.util.Log` class, use our `Logger` wrapper so it can be disabled globally.

 On Android 14 there is a new permission called `USE_EXACT_ALARM`, I think we should document it as an alternative to `SCHEDULE_EXACT_ALARM` but making it clear that is not an option for all apps.
 https://developer.android.com/reference/android/Manifest.permission#USE_EXACT_ALARM

local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationsPlugin.java
```diff
@@ -221,13 +227,34 @@ public void requestPermissions(PluginCall call) {
         }
     }
 
+    @PluginMethod
+    public void changeExactNotificationSetting(PluginCall call) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            startActivityForResult(call, new Intent(ACTION_REQUEST_SCHEDULE_EXACT_ALARM), "alarmPermissionsCallback");
+        }
```
jcesarmobile, line 234: if `< Build.VERSION_CODES.S` it should call `checkExactNotificationSetting(call);` so it resolves and returns as granted on older versions, as it's now it will hang in previous versions.

jcesarmobile (changes requested):
Requested some changes.

Also, the docs say `Only available on Android.`, but then we have "dummy" implementations where we just return `granted`. If we do that, then we shouldn't say that it's only available on Android, we should explain that on iOS and web it won't take you to settings but just return granted.
When something is only available on Android the other platforms should reject with the `unimplemented` method.
On iOS there are "time sensitive notifications" that we don't support at the moment, so not sure if returning granted can be confusing as some users might thing that the exact notifications can be related to the "time sensitive notifications", so maybe it's better to reject with `unimplemented`.

Docs should also mention that on Android < S it won't open settings.

local-notifications/src/definitions.ts
```diff
@@ -155,6 +155,24 @@ export interface LocalNotificationsPlugin {
    */
   requestPermissions(): Promise<PermissionStatus>;
 
+  /**
+   * Direct user to the application settings screen to configure exact alarms.
+   *
+   * Only available on Android >= S.
```
jcesarmobile, line 161: We usually use the version number (12) instead of the version letter (S), except on push notifications plugin, but there we also indicate the SDK number. 
I think the version number is more helpful for users.

jcesarmobile (changes requested):
The `changeExactNotificationSetting()` method is opening the general notifications settings page, not the specific app notification settings page, it should open the app page, to do that, the creation of the intent to `ACTION_REQUEST_SCHEDULE_EXACT_ALARM` should include a param with the app package.

When calling `changeExactNotificationSetting()` when it's granted and changing it to denied, the app restarts. I don't think the restart can be prevented, but we should document that behavior, and also [according to the docs](https://developer.android.com/reference/android/app/AlarmManager#ACTION_SCHEDULE_EXACT_ALARM_PERMISSION_STATE_CHANGED ) all the notifications that used exact will be deleted (also verified by scheduling a notification and changed the setting and it didn't fire), so we should document that behavior too, and not sure if we should try to re schedule them, or leave that to the developer?

We should probably also provide some listener for the `ACTION_SCHEDULE_EXACT_ALARM_PERMISSION_STATE_CHANGED` broadcast, but it looks like only fires when changing from disabled to enabled according to the docs, so not sure if it would be very useful as I think it's more important the rejection of the exact than the approval.

theproducer:
>  so we should document that behavior too, and not sure if we should try to re schedule them, or leave that to the developer?

We could, but there could be scenarios where a developer wouldn't want notifications rescheduled if exact alarms werent available, so it could be argued either way.  I'd say for now we just document it.

local-notifications/README.md
```diff
@@ -20,7 +20,9 @@ Starting on Android 12, scheduled notifications won't be exact unless this permi
 <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
 ```
 
-Note that even if the permission is present, users can still disable exact notifications from the app settings.
+Note that even if the permission is present, users can still disable exact notifications from the app settings.  Use `checkExactNotificationSetting()` to check the the value of the setting.  If a user disables this setting, the app will restart and any notification scheduled with an exact alarm will be deleted.  If your application depends on exact alarms, be sure to check this setting in order provide fallbacks or alternative behavior.
```
jcesarmobile, line 23: the "be sure to check this setting" should mention something like in "app resume", or in `App.appStateChange` event, or on app launch since the app is being restarted if the setting changes.

local-notifications/src/definitions.ts
```diff
@@ -155,6 +155,27 @@ export interface LocalNotificationsPlugin {
    */
   requestPermissions(): Promise<PermissionStatus>;
 
+  /**
+   * Direct user to the application settings screen to configure exact alarms.
+   *
+   * In the event that a user changes the settings from granted to denied, the application
+   * will restart and any notification scheduled with an exact alarm will be deleted.
+   *
+   * Only available on Android >= 12.
```
jcesarmobile, line 164: The method is also available on Android < 12, this should say "Only available on Android", but then in the description mention that on Android < 12 it doesn't open the settings and just returns granted.

jcesarmobile (changes requested):
added some minor comments

jcesarmobile (approved):

PR #1842, opened 2023-10-16
chore(ci): Use macOS 13 and Xcode 15.0

jcesarmobile:


markemer (approved):

markemer:
It is a little concerning to me that these won't run. The ubuntu setup job fails, but that shouldn't break based on your changes. So maybe we found another bug? Also, we have a old version of cancel-workflow that needs to be updated, but that's a separate PR.

IT-MikeS (approved):

PR #1843, opened 2023-10-16
chore(ci): Use node 16 on changed-packages action

jcesarmobile:


markemer:
Try something like this instead - this isn't exactly right, but it's in the ballpark:
```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```

markemer:
Is it the undici version? Or is that a red herring?
`/home/runner/work/capacitor-plugins/capacitor-plugins/node_modules/undici/lib/fetch/util.js:77`

jcesarmobile:
latest update of `@actions/http-client` added undici dependency and it fails for some reason
I tried to bump the node version since undici requires node 14 and the script was using node 12, but bumping the node version doesn't fix the issue
If tomorrow is not fixed in `@actions/http-client` I'll pin previous version where undici was not installed.

PR #1849, opened 2023-10-20
fix(share): Address Android 14 changes

jcesarmobile:
This is the simpler solution I've found for addressing Android 14 changes.
I think `FLAG_ALLOW_UNSAFE_IMPLICIT_INTENT` is safe to use because we only use the pending intent for getting the package of the app the user selected to share, we null the value right before the share starts so if somebody managed to get our `BroadcastReceived` it will have no effect.
But if somebody thinks this is unsafe I can try some different approach that doesn't use `FLAG_ALLOW_UNSAFE_IMPLICIT_INTENT`.

### DIFF
--- a/share/android/src/main/java/com/capacitorjs/plugins/share/SharePlugin.java
+++ b/share/android/src/main/java/com/capacitorjs/plugins/share/SharePlugin.java
@@ -42,7 +42,12 @@ public class SharePlugin extends Plugin {
                     }
                 }
             };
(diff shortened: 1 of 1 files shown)


IT-MikeS (approved):

IT-MikeS (approved):

ItsChaceD (approved):
I don't think there is a  risk for using `FLAG_ALLOW_UNSAFE_IMPLICIT_INTENT`  for our use case, especially since it's null before the share starts. LGTM

bosh-code:
Sorry to comment on a closed PR, but can we please get this released in version 5.0.7?

jcesarmobile:
No, it’s a breaking change, targeting SDK 34 will only be supported in capacitor 6.

PR #1855, opened 2023-10-23
docs(google-maps/geolocation)!: Removing deprecated usage descriptions for iOS

theproducer:
Removing `NSLocationAlwaysUsageDescription` in exchange for `NSLocationAlwaysAndWhenInUseUsageDescription` as its now [deprecated](https://developer.apple.com/documentation/bundleresources/information_property_list/nslocationalwaysusagedescription).

### DIFF
--- a/geolocation/README.md
+++ b/geolocation/README.md
@@ -13,8 +13,10 @@ npx cap sync
 
 Apple requires privacy descriptions to be specified in `Info.plist` for location information:
 
-- `NSLocationAlwaysUsageDescription` (`Privacy - Location Always Usage Description`)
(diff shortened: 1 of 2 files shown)


markemer (approved):

jcesarmobile (changes requested):
`NSLocationAlwaysAndWhenInUseUsageDescription` is for when the app gets the location in the background and I don't think the plugin allows that, so I don't think we should add a replacement, just remove the `NSLocationAlwaysUsageDescription`.
No need to add a note about it being deprecated since it was also not needed.

I've marked it as breaking change so we document in the upgrade guide that users should remove the `NSLocationAlwaysUsageDescription` if they have it and don't use background geolocation (with another plugin)

jcesarmobile (approved):

PR #1862, opened 2023-10-30
feat(filesystem): add `removeAllListeners` method

robingenz:
This PR adds the missing method `removeAllListeners`.

### DIFF
--- a/filesystem/ios/Plugin/FilesystemPlugin.m
+++ b/filesystem/ios/Plugin/FilesystemPlugin.m
@@ -18,4 +18,5 @@ CAP_PLUGIN(FilesystemPlugin, "Filesystem",
            CAP_PLUGIN_METHOD(checkPermissions, CAPPluginReturnPromise);
            CAP_PLUGIN_METHOD(requestPermissions, CAPPluginReturnPromise);
            CAP_PLUGIN_METHOD(downloadFile, CAPPluginReturnPromise);
+           CAP_PLUGIN_METHOD(removeAllListeners, CAPPluginReturnNone);
 )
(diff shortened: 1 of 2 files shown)


jcesarmobile (changes requested):
Can you rebase or cherry-pick this commit into 5.x branch? 
In Capacitor 6 it won't be necessary to do this and we will remove those methods, but would be good to fix it for Capacitor 5 version.

Actually, it should be two separate PRs as the type changes are still needed for Capacitor 6, can you remove the .m changes in this PR and send a separate one to 5.x branch with both changes. Also the return type should be `CAPPluginReturnPromise`, not `CAPPluginReturnNone`

robingenz:
Done. I created a new PR: #1868

filesystem/ios/Plugin/FilesystemPlugin.m
```diff
@@ -18,4 +18,5 @@
            CAP_PLUGIN_METHOD(checkPermissions, CAPPluginReturnPromise);
            CAP_PLUGIN_METHOD(requestPermissions, CAPPluginReturnPromise);
            CAP_PLUGIN_METHOD(downloadFile, CAPPluginReturnPromise);
+           CAP_PLUGIN_METHOD(removeAllListeners, CAPPluginReturnNone);
```
jcesarmobile, line 21:
```suggestion
```
Remove this as we are removing them on main branch as it won't be needed for Capacitor 6
https://github.com/ionic-team/capacitor-plugins/pull/1867

jcesarmobile (approved):

PR #1873, opened 2023-11-01
feat(camera): Add size to returned object

### LINKED ISSUE #1872: Feature: Camera plugin should return file size
### Plugin
Camera

### Description
In order to validate selected files, it is common to check that the file is not too large. From camera we might not have much control over this, but from the photo gallery it's possible to select a large saved image.

### Platform(s)
iOS, Android, Web

### Preferred Solution
Add a 'size' field on the return object

### Alternatives
Could be appended the exif object, but given that is not very well defined to begin with, I think it's better to have a separate field.

kentora:
resolves #1872 

What version might this be released as? I marked since as 5.?.0 for now, but will change when correct version is known.


eric-horodyski:

More context:

PR #1878, opened 2023-11-02
Fixed iOS filesystem stat

oliveryasuna:
Was returning a `string` for `ctime` and `mtime`


jcesarmobile:
Thanks for the pull request, but I noticed that it was generating some warnings.

I've sent a separate PR that addresses the issue in a different way without generating warnings and also fixes the issue for readDir and not only for stat.
https://github.com/ionic-team/capacitor-plugins/pull/1947

PR #1886, opened 2023-11-07
feat(app,haptics,status-bar,keyboard): Supporting Swift Package Manager

theproducer:
Converting the App, Haptics, Status Bar and Keyboard plugins to support the Swift Package Manager.


keyboard/package.json
```diff
@@ -45,11 +45,11 @@
     "publish:cocoapod": "pod trunk push ./CapacitorKeyboard.podspec --allow-warnings"
   },
   "devDependencies": {
-    "@capacitor/android": "^5.0.0",
-    "@capacitor/cli": "^5.0.0",
-    "@capacitor/core": "^5.0.0",
+    "@capacitor/android": "file:../../capacitor/android",
```
IT-MikeS, line 48: probably don't want file refs

IT-MikeS (changes requested):
Does this change the plugins from being able to be used in a pods based config like current users projects?

theproducer:
> Does this change the plugins from being able to be used in a pods based config like current users projects?

No, we maintain support for Pods based projects.  The respective Podspecs just point to the new source directories.

IT-MikeS (approved):

markemer (approved):

PR #1907, opened 2023-11-17
docs(screen-reader): Use stateChange event in example code

dtarnawsky:
The example code snippet in docs for @capacitor/screen-reader does not compile. Use the `stateChange` event in example code rather than non-existent `screenReaderStateChange` event.


jcesarmobile (approved):
good catch!

theproducer (approved):

PR #1914, opened 2023-11-21
fix(screen-orientation): Add  OrientationLockType

dtarnawsky:
Typescript > v5.2 removes `OrientationLockType`. This adds the type back in.

See (https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1615).

Fixes WN-1457.


screen-orientation/src/definitions.ts
```diff
@@ -1,9 +1,22 @@
 import type { PluginListenerHandle } from '@capacitor/core';
 
 export interface OrientationLockOptions {
+  /**
+   * Note: OrientationLockType was removed from typescript >= 5.2.
```
jcesarmobile, line 5: The note is not very clear, I think it should mention that typescript 5.2 users might need to import the `OrientationLockType` from the plugin.

jcesarmobile (changes requested):
just a minor comment about the noe

I prefer this approach over https://github.com/ionic-team/capacitor-plugins/pull/1929 as that seems very complex and harder to maintain.

I tested this approach and didn't seem to generate conflicts in apps using typescript < 5.2.

I released a dev build of screen-orientation with dev version: 6.0.0-dev-1915-20231122T130858.0 with this change

screen-orientation/src/definitions.ts
dtarnawsky: I've rewritten this note

jcesarmobile (approved):

PR #1918, opened 2023-11-28
fix(google-maps): Emit onMyLocationClick on iOS

pcl:
Previously, the logic was emitting `onMyLocationButtonClick` events for both the button click and an actual position click. Looking at the adjacent comment and the Android code, it appears that this is a copy-paste error, and that `onMyLocationClick` should be emitted instead.


eric-horodyski:

More context:

Issue #1905, opened 2023-11-16
iOS 17: Keyboard.setStyle not updating keyboard style until the second time you open it


aeharding:
### Plugin(s)

Keyboard 5.0.6

### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 5.5.1
  @capacitor/core: 5.5.1
  @capacitor/android: 5.5.1
  @capacitor/ios: 5.5.1

Installed Dependencies:

  @capacitor/android: not installed
  @capacitor/cli: 5.5.1
  @capacitor/core: 5.5.1
  @capacitor/ios: 5.5.1

[success] iOS looking great! 👌
```

### Platform(s)

iOS (specifically, iOS 17)

### Current Behavior

- Call `Keyboard.setStyle` with new appearance
- Open keyboard, observe incorrect style
- Close and reopen keyboard. Observe correct style

### Expected Behavior

- Call `Keyboard.setStyle`
- Open keyboard, observe correct style

### Code Reproduction

- https://github.com/aeharding/cap-setstyle
- Clone, install, `ionic capacitor build ios`
- Follow current behavior instructions specified above

### Other Technical Details

- iOS 16.4 ✅ working [simulator]
- iOS 17.0.1 ❌ not working [simulator]
- iOS 17.1 ❌ not working [real device]

This appears to be a regression in iOS 17.

### Additional Context

Videos:

#### iOS 17.1 ❌ not working [real device]

https://github.com/ionic-team/capacitor-plugins/assets/2166114/18337683-332d-48ab-b73a-96e5b4e44c91

#### iOS 16.4 ✅ working [simulator]

https://github.com/ionic-team/capacitor-plugins/assets/2166114/dcb1f144-6895-4425-b9fe-4d5db7e499a6

PR #1920, opened 2023-11-28
fix(keyboard): Change keyboard style during setStyle

theproducer:
Fixes an issue on iOS 17 where setting keyboard style would not apply the new appearance.

closes: https://github.com/ionic-team/capacitor-plugins/issues/1905

### DIFF
--- a/keyboard/ios/Sources/KeyboardPlugin/Keyboard.m
+++ b/keyboard/ios/Sources/KeyboardPlugin/Keyboard.m
@@ -121,7 +121,6 @@ double stageManagerOffset;
 
 - (void)onKeyboardWillShow:(NSNotification *)notification
 {
-  [self changeKeyboardStyle:self.keyboardStyle];
   if (hideTimer != nil) {
     [hideTimer invalidate];
   }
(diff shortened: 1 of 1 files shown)


markemer (approved):

jcesarmobile (changes requested):
This doesn't fix the reported issue.
The reported issue is when the keyboard is hidden, if you change the style, next time the keyboard appears it still has the old style on iOS 17 devices, and on the second time it appears it has the correct style.
So calling `[self changeKeyboardStyle:self.keyboardStyle];` on `onKeyboardWillHide` will have no effect since the keyboard is hidden and `onKeyboardWillHide` will not be called.

What seems to work is to call `[self changeKeyboardStyle:self.keyboardStyle];` inside the `- (void)setStyle:(CAPPluginCall *)call`.

theproducer:
> ...when the keyboard is **hidden**...

Ah right, sorry, I was tunnel visioned in to testing toggling open keyboards back and forth.

jcesarmobile (changes requested):
This other call to the same method can be removed now
https://github.com/ionic-team/capacitor-plugins/blob/main/keyboard/ios/Sources/KeyboardPlugin/Keyboard.m#L124

Issue #1950, opened 2023-12-05
@capacitor/camera does not honor users selection order


montanarilo:
### Plugin(s)
@capacitor/camera

### Capacitor Version
```
Latest Dependencies:

  @capacitor/cli: 5.5.1
  @capacitor/core: 5.5.1
  @capacitor/android: 5.5.1
  @capacitor/ios: 5.5.1

Installed Dependencies:

  @capacitor/cli: 5.2.2
  @capacitor/core: 5.2.2
  @capacitor/android: 5.2.2
  @capacitor/ios: 5.2.2
```

### Platform(s)
iOS

### Current Behavior
Method `Camera.pickImages` exposed by the `@capacitor/camera` plugin does not return image details using the same order in which users selected those images. Furthermore, selecting images over and over again, may yield different results each time (result of the `pickImages` is not stable from a sorting point of view).

In other words, if I have images 1 through 50, `Camera.pickImages` may return images in any arbitrary order. So, rolling with the same example as before, if I select images 1 through 50, I may get details about them in orders like:

- `1,2,3,4,...49,50` (if I'm very lucky)
- `2,1,3,4,7,10,5,....50,49`
- or in any other permutation of those 50 numbers.

This is different from the behavior observed on Android where details about each image are returned using the same ordering in which user picked them. So, for example, in the first position of the array returned by the `pickImages` method, I'll find details about the firstly selected image. In the second position of that array, I'll find details of the second image selected by the user, and so on for each other image. 

Same thing goes for the underlying `<input type="file" multiple="multiple" />` we find in the web implementation of the `@capacitor/camera` plugin.

### Expected Behavior
Expected result is that the order of selection is preserved, even when using the iOS implementation. 
Additionally, selecting the same set of images multiple times, should yield to the same result.

### Code Reproduction
```typescript
const { photos } = await Camera.pickImages({
  quality: 100,
});

console.log(photos);
```

If the above snippet uses the web or Android implementations of the `pickImages` method, `photos` will be an array where each item is presented in the same order in which user selected images. If the above snippet uses the iOS implementation, `photos` may have different orders. This is especially true and evident when selecting many images (with 30 images you are pretty much certain to encounter this problem). 

### Additional Context
I did some debugging and it seems like the problem lies in the function `picker` defined in the `CameraPlugin` extension (file `CameraPlugin.swift`). 
[...]

PR #1951, opened 2023-12-05
fix(camera): now we honor users selection order

l0ll098-at-mmfg:
Fixed the issue described in #1950 that, due to a race condition, lead the underlying iOS implementation of the `Camera.pickImages` method to not honor the order in which users selected images. 

Fixes: #1950


eric-horodyski:

More context:

Issue #1050, opened 2022-06-19
@capacitor/camera pickImages cancellation is not catched in web


Newbie012:
### Capacitor Version

```
   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 3.6.0
  @capacitor/core: 3.6.0
  @capacitor/android: 3.6.0
  @capacitor/ios: 3.6.0

Installed Dependencies:

  @capacitor/cli: 3.5.0
  @capacitor/core: 3.5.0
  @capacitor/android: 3.5.0
  @capacitor/ios: 3.5.0
```

### Platform(s)

web

### Current Behavior

While triggering `Camera.pickImages` in the web platform and then canceling the operation, nothing is being thrown, and the function never resolves. This leads to a memory leak:

### Expected Behavior

While canceling the image picking, a rejection should be invoked for proper handling.

### Code Reproduction

1. Open https://stackblitz.com/edit/vitejs-vite-pveyhh?file=src/App.tsx
2. Notice the app starts with **Status: idle**
3. Click the pick image button
4. Text should be changed to **Status: pending**
5. Close file picking without any image selection.
6. Text should've been switched to `rejected`, but still in pending, and the promise is pending as well.

### Other Technical Details

While looking at the source code:
https://github.com/ionic-team/capacitor-plugins/blob/0fb107ac46bca30ba7c1a7c09a93c866e6c40f87/camera/src/web.ts#L47-L49

It appears that the web doesn't handle any rejects.

Newbie012:
After doing a bit of some digging, it turns out that apparently, there's no cancel event or any indication for that in some cases, which would probably explain why this is not implemented. Maybe we should add a timeout option for rejection?

LucEnden:
Any news on when this issue will be fixed? As of today I am still able to reproduce this error...

Code bellow displays the same behavior as described in the bug:
```typescript
const picture = await Camera.getPhoto({
    resultType: CameraResultType.Uri,
    source: CameraSource.Photos
}).then((result) => {
    console.debug('Picture upload result', result); // does not appear
    return result;
}).catch((error) => {
    console.error('Error while uploading a picture', error); // does not appear
    return undefined;
}).finally(() => {
    console.log('Done uploading') // does not appear
});
```

SvenBudak:
[...]

PR #1958, opened 2023-12-10
fix(camera): reject promise on web input cancel event

andrzejSulkowski:
This PR adds promise rejections for the pwa camera plugin upon cancelation of the input element.

Previously, in the web implementation of the PWA camera plugin, the promise would remain unresolved if the user cancelled the input, as it was solely reliant on the 'change' event. This behaviour differed from the iOS/Android implementation, where the promise is rejected under similar circumstances.

closes https://github.com/ionic-team/capacitor-plugins/issues/1050

### DIFF
--- a/camera/src/web.ts
+++ b/camera/src/web.ts
@@ -88,7 +88,11 @@ export class CameraWeb extends WebPlugin implements CameraPlugin {
     }
   }
 
-  private fileInputExperience(options: ImageOptions, resolve: any, reject: any) {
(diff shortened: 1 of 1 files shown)


jcesarmobile (approved):
Thanks for the PR, I've fixed some lint issues, next time you can run `npm run fmt` and it will fix the issues automatically.

cancel event is pretty new, so it will only fix the issue for new browsers, but that's better than nothig

PR #1966, opened 2023-12-18
feat(preferences): Support for iOS suites

micbis:
Added support for iOS suites - this is needed i.e. to share data with a notification service extension.


chnuessli:
Wow, this would be awesome. Thank you. 
Can you please approve this PR that we can test that?

JohanBeumer:
It's been a while since this PR was created and it is behind now. That's a shame, because this change is really wanted by at least me, but also others.

@micbis if you see the chance to update the branch with the base branch it would be greatly appreciated!

micbis:
@JohanBeumer Sure, we'll update the pull request in the next days.

PR #1972, opened 2023-12-20
feat(browser): implement close() method for android

ItsChaceD:
Closes #151 

This PR adds the `close()` method to the Android Browser Plugin.
We create an "intermediate" controller activity that launches the custom tabs, and when we want to close, we launch the same activity with `FLAG_ACTIVITY_CLEAR_TOP` which triggers `onNewIntent` where we can detect to destroy the activity & custom tabs, without sacrificing the existing activity stack.

### DIFF
--- a/browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserPlugin.java
+++ b/browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserPlugin.java
@@ -1,7 +1,6 @@
 package com.capacitorjs.plugins.browser;
 
 import android.content.ActivityNotFoundException;
-import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
 import com.getcapacitor.Logger;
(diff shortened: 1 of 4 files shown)


giralte-ionic (approved):

browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserPlugin.java
```diff
@@ -50,18 +62,37 @@ public void open(PluginCall call) {
 
         // open the browser and finish
         try {
-            implementation.open(url, toolbarColor);
+            //implementation.open(url, toolbarColor);
```
markemer, line 65: This just dead code, or do you need to comment it back in.

markemer:
This looks ok to me, but I'm going to wait on @theproducer's call on his group of real devices before I sign off.

browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserPlugin.java
```diff
@@ -1,6 +1,8 @@
 package com.capacitorjs.plugins.browser;
 
 import android.content.ActivityNotFoundException;
+import android.content.Context;
```
theproducer, line 4: Looks like you can remove this line if its not being used

theproducer (approved):

Issue #1314, opened 2022-12-03
Camera pickImages() doesn't work first time on iOS


Danchat5:
### Plugin(s)
"@capacitor/camera": "^4.1.1"

### Capacitor Version
4.0.1

### Platform(s)
iOS, works fine on Android devices.

### Current Behavior
When the pickImages() function runs, the option menu on iOS (that gives the options Gallery, Take Picture, or Files) does not appear the first time. Every consecutive time the function is called, it works as intended. 

```
  async pickImage () {
    const newPhotos = await Camera.pickImages({
      quality: 100,
      limit: 1,
      correctOrientation: true
    })
}
```

### Code Reproduction
I have not been able to replicate this error on a standalone app, which seems to imply a plugin issue.

### Additional Context
This appears to be a near duplicate of [this issue](https://github.com/ionic-team/capacitor-plugins/issues/702), except for the fact that Android was the offending platform. I ensured that the plugin that caused the issue on that problem was not installed in my application.

I will list the plugins I am using below:
```
"@angular/animations": "^14.2.2",
    "@angular/common": "^14.2.2",
    "@angular/compiler": "^14.2.2",
    "@angular/compiler-cli": "^14.2.2",
    "@angular/core": "^14.2.2",
    "@angular/forms": "^14.2.2",
    "@angular/platform-browser": "^14.2.2",
    "@angular/platform-browser-dynamic": "^14.2.2",
    "@angular/router": "^14.2.2",
    "@angular/service-worker": "^14.2.2",
    "@awesome-cordova-plugins/email-composer": "^5.45.0",
    "@awesome-cordova-plugins/globalization": "^5.45.0",
    "@capacitor/android": "^4.2.0",
    "@capacitor/app": "^4.0.1",
    "@capacitor/browser": "^4.0.1",
    "@capacitor/camera": "^4.1.1",
    "@capacitor/core": "^4.2.0",
    "@capacitor/dialog": "^4.0.1",
    "@capacitor/filesystem": "^4.1.1",
    "@capacitor/haptics": "^4.0.1",
    "@capacitor/ios": "^4.2.0",
    "@capacitor/keyboard": "^4.0.1",
    "@capacitor/preferences": "^4.0.1",
    "@capacitor/push-notifications": "^4.1.0",
    "@capacitor/splash-screen": "^4.0.1",
    "@capacitor/status-bar": "^4.0.1",
    "@fortawesome/angular-fontawesome": "^0.11.1",
[...]

PR #1978, opened 2023-12-30
bug(camera): pickImages() doesn't work first time on safari #1314

stephan-fischer:
Solution is, to early import the web package. Safari tracks time differences between click and code execution to open the input file picker.

### DIFF
--- a/camera/src/index.ts
+++ b/camera/src/index.ts
@@ -1,9 +1,11 @@
 import { registerPlugin } from '@capacitor/core';
 
 import type { CameraPlugin } from './definitions';
+// See https://github.com/ionic-team/capacitor-plugins/issues/1314
+import * as web from './web';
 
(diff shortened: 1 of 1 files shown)


jcesarmobile (changes requested):
Can you only import `CameraWeb` instead of `*` and remove the comment?

Take clipboard plugin as an example
https://github.com/ionic-team/capacitor-plugins/pull/342

stephan-fischer:
> Can you only import `CameraWeb` instead of `*` and remove the comment?
> 
> Take clipboard plugin as an example #342

Done 😊

jcesarmobile (approved):

luofan189:
Hi, thx a lot for fixing this bug on Safari. Quick question, will this merge back to 5.x release? currently, we are still using xcode 13 and ready to migrate to xcode 15 yet and capacitor 6 yet. thx a lot

PR #1980, opened 2024-01-03
feat!(keyboard): Improvements to keyboard show / hide events

theproducer:


markemer (approved):
I always love removing direct frame math.

jcesarmobile (changes requested):
This is removing the `resizeOnFullScreen` feature (https://capacitorjs.com/docs/apis/keyboard#configuration)
I don't think that's intended, the `resizeOnFullScreen` feature should continue to work.

IT-MikeS (approved):

tobeagram:
I've just upgraded to Capacitor v6 and noticed that the recent PR not only refactors but also changes how the keyboard height is calculated to include the navigation gesture bar. The keyboard height can vary depending on whether a user uses three-button navigation or swipe gesture navigation. This added value makes it impossible to predict the keyboard's height within the bounds of our Capacitor web view.

For our use case, we need to either roll back these changes or amend them to exclude the navigation bar height. Is anyone else encountering this issue?

seopei:
@tobeagram I have the same issue too

tobeagram:
For anyone wanting the original behaviour (@seopei) I've forked it [here](https://github.com/supernotes/capacitor-plugins/commit/f2a87c33159b74ba7a45cdf82d37a50a4505249e) and published it [here](https://www.npmjs.com/package/@supernotes/capacitor-keyboard).

chriswep:
This PR breaks keyboard listeners on Android  <= 10 devices when `android:windowSoftInputMode="adjustPan"` is set. It just doesn't call the listener callbacks when the keyboard is shown. Changing the input mode to `adjustResize` makes the listeners work again, however we used `adjustPan` since it allowed us to get the keyboardHeight without the webivew being resized. Looking at the diff it seems that the implmentation was basically replaced completely. Something about the new implmentation made it stop working for the `adjustPan` mode on older Android sdks.

What is working:
- Android 10 and Downgrading to v5 of capacitor-keyboard 
- `adjustPan` and Android 14 (haven't tested other SDKs)
- soft input mode `adjustResize`

xanather:
Also have this issue, downgrading for now.

I will create an issue to track this.

xanather:
See https://github.com/ionic-team/capacitor-plugins/issues/2168, might get more visibility there rather than commenting on this PR.

PR #2026, opened 2024-02-07
Update README.md

### LINKED ISSUE #2025: Add common issue to push-notifications/README.md about iOS simulator support
### Plugin

- push-notifications

### Description

Add a short section into the readme to let developers know that the iOS simulator does not support push notification registration.
I spent about 5 hours yesterday and today trying to figure out why the registration and registrationError listener events were not firing for me, and it turns out that for me, on Xcode 14.2 with an ios 16.1 simulator, the failure is totally silent and simply doesn't fire the event `didRegisterForRemoteNotificationsWithDeviceToken` or `didFailToRegisterForRemoteNotificationsWithError` in the appdelegate (confirmed with print statement debugging).

I confirmed that my project works fine on a physical device, so it is indeed only the iOS simulator which doesn't support registering a device token.

### Platform(s)

Docs only

### Preferred Solution

Adding the section into the readme.

### Alternatives

There might be a way to detect if the user is in a simulator and throw an error to the registrationError event.

### Additional Context

https://stackoverflow.com/questions/1080556/how-can-i-test-apple-push-notification-service-without-an-iphone/60085404#60085404

matthewbal:
Added a section to common issues, reminding users that on an iOS simulator the register event will fail silently.

Closes #2025

### DIFF
--- a/push-notifications/README.md
+++ b/push-notifications/README.md
@@ -130,6 +130,8 @@ On Android, there are various system and app states that can affect the delivery
 * If the device has entered [Doze](https://developer.android.com/training/monitoring-device-state/doze-standby) mode, your application may have restricted capabilities. To increase the chance of your notification being received, consider using [FCM high priority messages](https://firebase.google.com/docs/cloud-messaging/concept-options#setting-the-priority-of-a-message).
 * There are differences in behavior between development and production. Try testing your app outside of being launched by Android Studio. Read more [here](https://stackoverflow.com/a/50238790/1351469).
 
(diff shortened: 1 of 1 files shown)


eric-horodyski:
@matthewbal - Please update your PR if you would like to get it reviewed. We appreciate your contribution!

matthewbal:
@eric-horodyski Sorry this one slipped my mind - I can't figure out what I should update in this PR, could you please give me some guidance on that? I haven't done a lot of open source contributions before so I'm new to the process.

jcesarmobile:
Capacitor 8 requires Xcode 26, so I see no value in adding this note.

PR #2066, opened 2024-03-26
Fix java.lang.NullPointerException bug on Android

josharnow:
### Fixes "java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference"

Within `getInputStream()`, there is a conditional in which `.equals()` is invoked on `u.getScheme()`.

If `u.getScheme()` is `null` as can sometimes be the case, this will break functions such as `Filesystem.readFile()` by breaking the `getInputStream()` function.

This fix involves first checking that `u.getScheme() != null` within the conditional.


eric-horodyski:

More context:

Issue #2072, opened 2024-04-03
@capacitor/browser add BROWSER_CLOSED event


kurtiev:
### Plugin

@capacitor/browser

### Description

This is probably feature-request and bug issue as well :)
I faced with problem when browser window is minimized (not closed) and event **_browserFinished_** Is triggered.
I need to know when is was closed, not minimized. Take a look the attached screenshot, it has "minimize" control.

### Platform(s)

Android only.

### Preferred Solution

Add one more event [here](https://github.com/ionic-team/capacitor-plugins/blob/0ccb9c48af9740a24f1f084e9417d5d5e2201311/browser/android/src/main/java/com/capacitorjs/plugins/browser/Browser.java#L35) BROWSER_CLOSED or don't trigger BROWSER_FINISHED while inappbrowser is minimized.

### Additional Context

![IMG_2111](https://github.com/ionic-team/capacitor-plugins/assets/15800725/1be9b008-de93-4135-bf5d-b89951bab645)

cornflakes:
The PR: https://github.com/ionic-team/capacitor-plugins/pull/2125 would in turn fix this bug.

This bug is introduced in commit https://github.com/ionic-team/capacitor-plugins/commit/e2cdf3538b9741c5ec394abb14516ceaf5e12d43 by @ItsChaceD on January 22nd, 2024.

The issue arises because the URL is opened from an activity instead of the plugin itself. This causes the unbindService method (called from handleOnPause in the plugin) to be invoked before the open method (called from the activity callback). This results in group.enter() being called before group.reset(). Consequently, when group.leave() is triggered from the TAB_HIDDEN navigation event, the browserEventListener.onBrowserEvent(BROWSER_FINISHED) event fires incorrectly when the app goes into the background.

The pull request fixes the bug by changing the count threshold from count <= 0 to count < 0, preventing the unintended firing of the BROWSER_FINISHED event.

While the order of calls remains unchanged, modifying the count threshold achieves the same outcome of ensuring group.reset() happens before group.leave().

eric-horodyski:
This thread has been inactive for some time, and we’re cleaning up older items to keep the backlog focused and manageable.

Details on the effort:
[https://ionic.io/blog/keeping-capacitors-backlog-healthy](https://ionic.io/blog/keeping-capacitors-backlog-healthy)

If this is still relevant today, feel free to open a new issue with current information and we’ll review it.

cornflakes:
[...]

PR #2125, opened 2024-06-06
Proposed fix for #2072 - EventGroup count variable should allowed to be zero

cornflakes:
This would in turn fix:
https://github.com/ionic-team/capacitor-plugins/issues/2072

Due to the fact that TAB_HIDDEN is fired before TAB_SHOWN, the count variable would be zero and hence BROWSER_FINISHED would be fired prematurely.


aagavin:
Anyway we can get this merged?

eric-horodyski:
Thanks for opening this PR 🙏

Issue #749, opened 2022-01-07
How to remove or disable file picker in camera plugin


diadal:
how to prevent users from uploading files only allow the camera 

<img width="603" alt="Screen Shot 2022-01-07 at 2 54 53 PM" src="https://user-images.githubusercontent.com/19895516/148554098-448f4745-7f5f-4765-88ef-4528ae357cdd.png">

numerized:
I need that too thanks

eric-horodyski:
This thread has been inactive for some time, and we’re cleaning up older items to keep the backlog focused and manageable.

Details on the effort:
[https://ionic.io/blog/keeping-capacitors-backlog-healthy](https://ionic.io/blog/keeping-capacitors-backlog-healthy)

If this is still relevant today, feel free to open a new issue with current information and we’ll review it.

PR #2142, opened 2024-06-28
feat(camera): Add Camera option for disabling PWA elements picker

jcesarmobile:
I've made the default false to not break things, but I think for Capacitor 7 it should be made true to be consistent with the other platforms where there is no photo picker button on the Camera UI.

closes https://github.com/ionic-team/capacitor-plugins/issues/749


brune83:
Can you please release this PR? Thank you very much!

PR #2179, opened 2024-08-20
feat(statusbar)!: implement overlay and background color in iOS

andredestro:


status-bar/android/src/main/java/com/capacitorjs/plugins/statusbar/StatusBarPlugin.java
```diff
@@ -15,7 +16,35 @@ public class StatusBarPlugin extends Plugin {
...
     @Override
     public void load() {
-        implementation = new StatusBar(getActivity());
+        StatusBarConfig config = getStatusBarConfig();
+        implementation = new StatusBar(getActivity(), config);
+    }
+
+    private StatusBarConfig getStatusBarConfig() {
+        StatusBarConfig config = new StatusBarConfig();
+        String backgroundColor = getConfig().getString("StatusBarBackgroundColor");
```
theproducer, line 25: The name of the configuration string can probably just be `backgroundColor` (same with the others and on iOS).  It should be assumed that this is for the StatusBar plugin.

theproducer (changes requested):

status-bar/ios/Sources/StatusBarPlugin/StatusBarPlugin.swift
```diff
@@ -15,92 +15,93 @@ public class StatusBarPlugin: CAPPlugin, CAPBridgedPlugin {
...
+            config.style = style(fromString: configStyle)
         }
+        return config
     }
-
-    deinit {
-        if let observer = observer {
-            NotificationCenter.default.removeObserver(observer)
+    
+    private func style(fromString: String) -> UIStatusBarStyle {
```
markemer, line 40: Doesn't need to happen for this PR, but we could make this an `init()` on the style enum and have a computer variable that gives us a UIStatusBarStyle on the enum as well. Just putting that out into the ether for the future.

status-bar/ios/Sources/StatusBarPlugin/UIColor.swift
```diff
@@ -0,0 +1,40 @@
+import Capacitor
+
+public extension CapacitorExtensionTypeWrapper where T: UIColor {
```
markemer, line 3: Could this just be `public extension UIColor`?

status-bar/android/src/main/java/com/capacitorjs/plugins/statusbar/StatusBarConfig.java
```diff
@@ -0,0 +1,34 @@
+package com.capacitorjs.plugins.statusbar;
+
+import com.getcapacitor.util.WebColor;
+
+public class StatusBarConfig {
```
markemer, line 5: Fully up to you, but this file could have been Kotlin if you wanted. We should check, but I think fully new files can be Kotlin. Someone correct me if I'm wrong here.

status-bar/ios/Sources/StatusBarPlugin/UIColor.swift
andredestro: It could be.
But with just `public extension UIColor`, the method is called by `UIColor.hex(fromColor: backgroundColor)` only instead of `UIColor.capacitor.hex(fromColor: backgroundColor)` like other Capacitor extension methods.

status-bar/android/src/main/java/com/capacitorjs/plugins/statusbar/StatusBarConfig.java
andredestro: We need to add Kotlin support in the build.gradle first I think.

theproducer: No, Kotlin is already supported.  However, I don't think it makes sense to do it now.  We can make the Java -> Kotlin conversion its own PR.

status-bar/android/src/main/java/com/capacitorjs/plugins/statusbar/StatusBar.java
```diff
@@ -9,18 +11,26 @@
 import androidx.core.view.WindowCompat;
 import androidx.core.view.WindowInsetsCompat;
 import androidx.core.view.WindowInsetsControllerCompat;
+import androidx.localbroadcastmanager.content.LocalBroadcastManager;
```
theproducer, line 14: What is this `LocalBroadcastManager` used for?

theproducer (changes requested):

status-bar/ios/Sources/StatusBarPlugin/UIColor.swift
markemer: Do'h! Yep stick with this.

status-bar/ios/Sources/StatusBarPlugin/StatusBar.swift
```diff
@@ -0,0 +1,167 @@
+import Foundation
+import Capacitor
+
+@objc public class StatusBar: NSObject {
```
markemer, line 4: I don't think we need this `@objc`

markemer: I was thinking this could maybe be a UIView itself - but maybe that doesn't work tying to add it on top of the WebView.

theproducer, line 4: I agree with Mark, you don't really need `@objc` or `NSObject`.

Nor would you need `super.init()`.

theproducer (changes requested):

markemer (approved):

theproducer (approved):

PR #2192, opened 2024-09-09
feat(@capacitor/bowser) Add option to clear all website data while closing iOS browser window

GeroldSchierholz:
User Story:

After using the browser window for e.g. oauth login,  I would like to have the possibility to clear all website data to ensure that a new login is not using any cookie data and logging me in again.

Description:

Optional config to clear all website data during browser.close.

### DIFF
--- a/browser/README.md
+++ b/browser/README.md
@@ -34,7 +34,7 @@ const openCapacitorSite = async () => {
 <docgen-index>
 
 * [`open(...)`](#open)
-* [`close()`](#close)
+* [`close(...)`](#close)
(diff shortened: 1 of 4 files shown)


bentleyo:
@GeroldSchierholz nice! Out of curiosity, is it possible to add this as a method we can call programmatically whenever or does the browser have to be open? I was thinking it might be useful to clear the data without requiring the browser window to be open e.g. when hitting a sign-out button in an app.

If you wanted you could then programmatically call it on the `browserFinished` event to achieve the same functionality.

eric-horodyski:

More context:

PR #2222, opened 2024-10-24
feat(camera): only request permission to save to the gallery for Android <= 9

### LINKED ISSUE #2210: bug: camera opens after second attempt on Android with using Camera.getPhoto with saveToGallery set to true
On newer versions of android camera opens after 2 attempts. I call Camera.getPhoto with option saveToGallery set to true. The the system bottom sheet appears and I choose Take a photo, then the sheet disappears and nothing, I have to click the button again to call Camera.getPhoto and again choose Take a photo. But, if I set saveToGallery to false, everything works as expected. Also worth mentioning that after these manipulations camera opens from the first attempt until I restart the app.
I've tried disabling camera permissions and plugin doesn't ask any, I've tried manually request permissions but nothing pops up, I've tried manually enable permissions in settings and bug still presents. And one more thing, I've put console.log after await Camera.getPhoto and nothing appears in console on a first attempt when Camera didn't show off but on second my console.log has been fired.

I've tried to update these things as Android studio suggests despite Capacitor documentation says these versions should be set but it didn't help.
classpath 'com.google.gms:google-services:4.4.0'
classpath 'com.android.tools.build:gradle:8.2.1'
distributionUrl=https://services.gradle.org/distributions/gradle-8.2.1-all.zip

Capacitor dependencies
"@capacitor/android": "^6.1.2",
"@capacitor/app": "^6.0.1",
"@capacitor/camera": "^6.0.2",

alexgerardojacinto:
Fixes a bug on @capacitor/camera on Android, where when taking a photo with `saveToGallery = true`, the camera only opened on the second attempt because of the permission to save to the gallery not being given.

This only happened for Android >= 10, and for those versions, we don't even need to request for permission to save to the gallery. More info [here](https://developer.android.com/media/camera/camera-deprecated/photobasics#TaskPath).

Closes #2210 

References: https://outsystemsrd.atlassian.net/browse/RMET-3767

### DIFF
--- a/camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
+++ b/camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
@@ -204,7 +204,18 @@ public class CameraPlugin extends Plugin {
         boolean hasGalleryPerms = getPermissionState(SAVE_GALLERY) == PermissionState.GRANTED;
 
         // If we want to save to the gallery, we need two permissions
-        if (settings.isSaveToGallery() && !(hasCameraPerms && hasGalleryPerms) && isFirstRequest) {
+        // actually we only need permissions to save to gallery for Android <= 9 (API 28)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            // we might still need to request permission for the camera
+            if (!hasCameraPerms) {
+                requestPermissionForAlias(CAMERA, call, "cameraPermissionsCallback");
+                return false;
+            }
+            return true;
+        }
+
+        // we need to request permissions to save to gallery for Android <= 9
+        else if (settings.isSaveToGallery() && !(hasCameraPerms && hasGalleryPerms) && isFirstRequest) {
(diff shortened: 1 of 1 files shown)


camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java
```diff
@@ -204,7 +204,18 @@ private boolean checkCameraPermissions(PluginCall call) {
...
+            // we might still need to request permission for the camera
+            if (!hasCameraPerms) {
+                requestPermissionForAlias(CAMERA, call, "cameraPermissionsCallback");
+                return false;
+            }
+            return true;
+        }
+
+        // we need to request permissions to save to gallery for Android <= 9
+        else if (settings.isSaveToGallery() && !(hasCameraPerms && hasGalleryPerms) && isFirstRequest) {
```
OS-ricardomoreirasilva, line 218: Considering that the `if` statement above always returns, you don't need the `else` here.

IT-MikeS (approved):

PR #2240, opened 2024-11-08
Make Geolocation plugin work if Google Play Services are not enabled

pietervdvn:
Hi,

I'm currently in [the process of packaging](https://github.com/pietervdvn/MapComplete/issues/2112) [MapComplete](https://mapcomplete.org) into an Android app, with capacitor.

However, many of my users (including myselfs) are a bit paranoia and disable the Google Play Services or run custom ROMS without them. However, the geolocation plugin would refuse to work if those are not available.

As such, I have modified the geolocation-plugin to work with the google play services if they are available and to fallback to the stock Android API's if the Play Services are not available.

I've also moved some duplicated code into methods, in order to maintain some readability.


eric-horodyski:

More context:

PR #2244, opened 2024-11-13
docs(camera): Added Android 14 permission note

Jonas-Lieske:
Tested on multiple Android 14 devices and an Android 13 device. getPhoto() now requires the Camera permission to correctly prompt the user about the usage.

Example code:
```js
await Camera.requestPermissions();
const capturedPhoto = await Camera.getPhoto({
        resultType: CameraResultType.Uri,
        source: CameraSource.Camera,
        quality: 80,
});
```


jcesarmobile:
This is not true, Camera permission is not needed.

PR #2267, opened 2024-11-29
Added note on mobile-web-app-capable mete tag

jarrodchesney:
The text-zoom plugin would not work unless the following meta tag was added to index.html

<meta name="mobile-web-app-capable" content="yes" />


jcesarmobile:
Thanks for the PR, but I've removed that tag from my app and text-zoom plugin still works, I can't find any information about it being required for `webkitTextSizeAdjust` to work (what the plugin uses on iOS to set the zoom level)

Issue #2062, opened 2024-03-20
[Geolocation / Android] The GPS position is only updated every 5 seconds when using "watchPosition" function on Android


carelliguillaume:
### Plugin(s)
Geolocation

### Capacitor Version
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 5.7.3
  @capacitor/core: 5.7.3
  @capacitor/android: 5.7.3
  @capacitor/ios: 5.7.3

Installed Dependencies:

  @capacitor/cli: 5.5.1
  @capacitor/core: 5.5.1
  @capacitor/android: 5.5.1
  @capacitor/ios: 5.5.1

[success] iOS looking great! 👌
[success] Android looking great! 👌

### Platform(s)
Android

### Current Behavior
When you use "watchPosition" function of the Geolocation Plugin on Android, the GPS position is only updated every 5 seconds which is not sufficient for position tracking. The refresh on Android is jerky compared to iOS which is much smoother. Moreover, its hardcoded in the following class when calling method **setMinUpdateIntervalMillis** : 
https://github.com/ionic-team/capacitor-plugins/blob/main/geolocation/android/src/main/java/com/capacitorjs/plugins/geolocation/Geolocation.java

```
  LocationRequest locationRequest = new LocationRequest.Builder(10000)
      .setMaxUpdateDelayMillis(timeout)
      .setMinUpdateIntervalMillis(5000)
      .setPriority(priority)
      .build();
```

### Expected Behavior
We should be able to configure the **setMinUpdateIntervalMillis** to have more accurate position tracking on Android. We should have a parameter like the "timeout" parameter used in the **setMaxUpdateDelayMillis** call. At least we should be able to remove the **setMinUpdateIntervalMillis** call to let Android manage the refreshing automatically like iOS Geolocation Plugin.

### Code Reproduction
Just call the "watchPosition" of the Geolocation Plugin on Android

alexgerardojacinto:
@carelliguillaume we just released `@capacitor/geolocation@6.1.0` which adds a new `minimumUpdateInterval` parameter to `startWatch`, solving this issue.

PR #2272, opened 2024-12-05
feat(geolocation): add `minimumUpdateInterval` parameter for `startWatch`

alexgerardojacinto:
- Adds `minimumUpdateInterval` to `PositionOptions` so that it can be used when building the `LocationRequest` for Android on the `startWatch` feature.
- Context: We were hardcoding the minimum update interval for location updates to 5 seconds, which meant that if there was an update available in less than 5 seconds, that update would only be delivered after 5 seconds. Instead, this value should be set by the user/developer, so they can better control the frequency of location updates.

Solves: https://github.com/ionic-team/capacitor-plugins/issues/2062

References: https://outsystemsrd.atlassian.net/browse/RMET-3338


OS-martacarlos (approved):

OS-ricardomoreirasilva (approved):

geolocation/src/definitions.ts
```diff
@@ -180,6 +180,17 @@ export interface PositionOptions {
    * @since 1.0.0
    */
   maximumAge?: number;
+
+    /**
+   * The minumum update interval for location updates.
+   * 
+   * If location updates are available faster than this interval then an update 
+   * will only occur if the minimum update interval has expired since the last location update.
+   *
```
ItsChaceD, line 189: We should document that it is for Android only unless we are also adding this feature to iOS

ItsChaceD (changes requested):

geolocation/src/definitions.ts
alexgerardojacinto: Yeah good point. Already added the note as we agreed.

ItsChaceD (approved):

Issue #2224, opened 2024-10-28
Camera Prompt Landscape 


andrybany:
### Plugin(s)
Camera

### Capacitor Version

```
   Capacitor Doctor

Latest Dependencies:

  @capacitor/cli: 6.1.2
  @capacitor/core: 6.1.2
  @capacitor/android: 6.1.2
  @capacitor/ios: 6.1.2

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 5.7.8
  @capacitor/core: 6.1.2
  @capacitor/android: 6.1.2

[success] Android looking great! 👌
```

### Platform(s)

Android

### Current Behavior

As described by an other user in the https://github.com/ionic-team/capacitor-plugins/issues/1643, when device is in landscape mode, Camera.getPhoto() prompt/menu options is not correctly visualized. Only the "Photo" label is visible, the other two "From Photos" and "Take picture" labels are visible only after scrolled up the tab.

### Expected Behavior

The same behaviour of the portrait mode, with all the tab visible.

### Code Reproduction

https://github.com/andrybany/CapacitorCamera

### Other Technical Details

The app that I included in the above repo is very simple: it has only a button that call the getPhoto() method. 
To run the app on a pysical or emulated android device, I added a README.md file on the repo with the basic commands to do.

### Additional Context

JAVA_HOME jdk 20.0.2
node 16.16.0
npm 7.5.6
angular 12.0.4
capacitor 6

alexgerardojacinto:
@andrybany we have just released [@capacitor/camera@6.1.2](https://github.com/ionic-team/capacitor-plugins/releases/tag/%40capacitor%2Fcamera%406.1.2) which fixes the reported issue.

PR #2276, opened 2024-12-10
fix(camera): Android dialog not fully showing when in landscape

alexgerardojacinto:
- Fixes the `CameraSheetDialogFragment` that wasn't fully showing when orientation was landscape, by explicitly setting its state to `BottomSheetBehavior.STATE_EXPANDED`.
- Also removes unnecessary null check since `instanceof` will return `false` if `behavior` is null.

Closes: https://github.com/ionic-team/capacitor-plugins/issues/2224

### DIFF
--- a/camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraBottomSheetDialogFragment.java
+++ b/camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraBottomSheetDialogFragment.java
@@ -113,11 +113,12 @@ public class CameraBottomSheetDialogFragment extends BottomSheetDialogFragment {
         dialog.setContentView(parentLayout.getRootView());
 
         CoordinatorLayout.LayoutParams params = (CoordinatorLayout.LayoutParams) ((View) parentLayout.getParent()).getLayoutParams();
-        CoordinatorLayout.Behavior<?> behavior = params.getBehavior();
+        CoordinatorLayout.Behavior<View> behavior = params.getBehavior();
 
         if (behavior instanceof BottomSheetBehavior) {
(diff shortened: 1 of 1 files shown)


camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraBottomSheetDialogFragment.java
```diff
@@ -113,10 +113,11 @@ public void setupDialog(Dialog dialog, int style) {
...
 
         CoordinatorLayout.LayoutParams params = (CoordinatorLayout.LayoutParams) ((View) parentLayout.getParent()).getLayoutParams();
-        CoordinatorLayout.Behavior behavior = params.getBehavior();
+        CoordinatorLayout.Behavior<?> behavior = params.getBehavior();
 
-        if (behavior != null && behavior instanceof BottomSheetBehavior) {
-            ((BottomSheetBehavior) behavior).addBottomSheetCallback(mBottomSheetBehaviorCallback);
+        if (behavior instanceof BottomSheetBehavior) {
+            ((BottomSheetBehavior<?>) behavior).addBottomSheetCallback(mBottomSheetBehaviorCallback);
+            ((BottomSheetBehavior<?>) behavior).setState(BottomSheetBehavior.STATE_EXPANDED);
```
OS-ricardomoreirasilva, line 120: Couldn't we use `BottomSheetBehavior<View>` here? 
Also, wouldn't it be better to create a local variable and then use it for both calls? This way we could avoid having `((BottomSheetBehavior<?>) behavior)` duplicated.

OS-ricardomoreirasilva (approved):

IT-MikeS (approved):

Issue #2036, opened 2024-02-20
@capacitor/browser: browser.open Crashes Application 


emelampianakis:
### Plugin(s)

@capacitor/browser ^5.2.0

### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 5.7.0
  @capacitor/core: 5.7.0
  @capacitor/android: 5.7.0
  @capacitor/ios: 5.7.0

Installed Dependencies:

  @capacitor/cli: 5.7.0
  @capacitor/ios: 5.7.0
  @capacitor/core: 5.7.0
  @capacitor/android: 5.7.0

[success] iOS looking great! 👌
[success] Android looking great! 👌```

### Platform(s)
<!--
List the platforms that this bug affects.
-->
Android

### Current Behavior
<!--
Describe how the bug manifests. Be specific.
-->
This happens when the app is opened from the push listener 'notificationActionPerformed'
In the notification data: urlToOpen = 'www.google.com'
When using .open({url: urlToOpen}) app crashes & stays unresponsive until app data are cleared

If the urlToOpen =  'https://www.google.com/' --> works as expected.

### Expected Behavior
<!--
Describe what the behavior should be.
-->
Open the url in the in app browser

### Code Reproduction
<!--
To isolate the cause of the problem, we ask you to provide a minimal sample application that demonstrates the issue.
For full instructions, see: https://github.com/ionic-team/capacitor/blob/master/CONTRIBUTING.md#creating-a-code-reproduction
-->
```
 FirebaseMessaging.addListener('notificationActionPerformed', async (event) => {
        const urlToOpen = event.notification.data.urlToOpen;
        await Browser.open({
          url: urlToOpen,
        });
      });
```



### Other Technical Details
<!--
Please provide the following information with your request and any other relevant technical details (versions of IDEs, local environment info, plugin information or links, etc).
-->



### Additional Context
<!--
List any other information that is relevant to your issue. Stack traces, related issues, suggestions on how to fix, Stack Overflow links, forum links, etc.
-->
The crashlytics report:

```
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{com.example.myapp/com.capacitorjs.plugins.browser.BrowserControllerActivity}: android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat= pkg=com.android.chrome (has extras) }
[...]

PR #2282, opened 2024-12-15
Check if a browser supports Custom Tabs before trying to open an URL

DavidStrausz:
Adds a new method `areCustomTabsSupported` to the `Browser` class and uses the method in the `BrowserPlugin`class to check if any browser on the device can handle opening an URL via Custom Tabs. If no browser can handle it, fallback to opening the URL in the default browser.

I used the exact method suggested in the [official documentation](https://developer.chrome.com/docs/android/custom-tabs/howto-custom-tab-check#check_if_any_browser_on_the_device_supports_custom_tabs). The documentation also suggests to do this check to avoid `ActivityNotFoundException` exceptions:

> If no browser supporting Custom Tabs is available in the device and you launch a URL using customTabsIntent.launchUrl(context, url), the intent might fail, leading to an ActivityNotFoundException.

> Always perform a compatibility check to ensure a better user experience.

> You can fallback to a standard ACTION_VIEW Intent to open the URL in any available browser.

This will probably fix some of the exceptions mentioned in this issue: https://github.com/ionic-team/capacitor-plugins/issues/2036

### DIFF
--- a/browser/android/src/main/java/com/capacitorjs/plugins/browser/Browser.java
+++ b/browser/android/src/main/java/com/capacitorjs/plugins/browser/Browser.java
@@ -5,11 +5,17 @@ import static androidx.browser.customtabs.CustomTabsIntent.SHARE_STATE_ON;
 import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.content.pm.ResolveInfo;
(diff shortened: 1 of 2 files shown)


jetie000:
@DavidStrausz pls fix conflicts
@jcesarmobile @andredestro @alexgerardojacinto please look at this feature, I need this for business project

Issue #2241, opened 2024-11-11
PWA Camera cancel not working


co-dax:
### Plugin(s)
@capacitor/camera

### Capacitor Version
6.0

### Platform(s)
Web, PWA Anroid, PWA iOS

### Current Behavior
There is no way to cancel taking photos before chosing between the options "From Photos" and "Take Picture" when using @capacitor/camera plugin with PWA on both Android, iOS and Web and clicking on the backdrop does not close action sheet either. Below is an image to illustrate the issue:

![image](https://github.com/user-attachments/assets/3f2b3f56-7382-439d-88e7-e9687b17c440)

The only way to get out of this situation is to select either "From Photos" or "Take Picture" and then cancel.

jcesarmobile:
fixed in #2284

PR #2284, opened 2024-12-16
fix(camera): Allow cancelation of sheet in Web

OS-pedrogustavobilro:
The `actionSheet` prompt was previously not cancellable in Web, even though it is cancellable

- [on Android](https://github.com/ionic-team/capacitor-plugins/blob/main/camera/android/src/main/java/com/capacitorjs/plugins/camera/CameraPlugin.java#L172) -  `setCancelable(false)` not called, by default it is cancelable
- [and on iOS](https://github.com/ionic-team/capacitor-plugins/blob/main/camera/ios/Sources/CameraPlugin/CameraPlugin.swift#L417) - `UIAlertAction` with style `cancel` allows sheet to be canceled when clicking outside.

This PR fixes that.

Addresses https://github.com/ionic-team/capacitor-plugins/issues/2241 and https://github.com/ionic-team/pwa-elements/issues/134

### DIFF
--- a/camera/src/web.ts
+++ b/camera/src/web.ts
@@ -23,7 +23,7 @@ export class CameraWeb extends WebPlugin implements CameraPlugin {
           document.body.appendChild(actionSheet);
         }
         actionSheet.header = options.promptLabelHeader || 'Photo';
-        actionSheet.cancelable = false;
+        actionSheet.cancelable = true;
         actionSheet.options = [
           { title: options.promptLabelPhoto || 'From Photos' },
           { title: options.promptLabelPicture || 'Take Picture' },


alexgerardojacinto (approved):

OS-ricardomoreirasilva (approved):

jcesarmobile:
The reason why it's not cancelable on web is because pwa-elements action sheet doesn't provide any event to indicate the cancelation, so allowing to cancel the action sheet will make the camera plugin call to hang as it won't reject nor resolve.

So for making the prompt cancelable, pwa-elements action sheet should be updated to emit an event on the cancellation and in camera plugin listen for that event to reject the call with an error as iOS and Android do.

OS-pedrogustavobilro:
@jcesarmobile  made a PR for pwa-elements to add an event listener for when the action sheet is dismissed - https://github.com/ionic-team/pwa-elements/pull/138 

Then, in this PR, will add the equivalent cancelation that's already present on native Android and iOS.

```javascript
actionSheet.addEventListener('onCanceled', async () => {
    reject(new CapacitorException('User cancelled photos app'));
})
```

OS-ricardomoreirasilva (approved):

co-dax:
@jcesarmobile  When will this fix be released?

alexgerardojacinto (approved):

Issue #2154, opened 2024-07-17
Allow click-outside and back-button on ActionSheet on Android


funkyvisions:
### Plugin
action-sheet

### Description
Would be nice if tapping outside of the ActionSheet or using the hardware/software back button would close the ActionSheet.

### Platform(s)
Android

### Preferred Solution
Add the following to ActionSheetPlugin.java

Change
```
implementation.setCancelable(true);
```
to
```
           implementation.setCancelable(true);
           implementation.setOnCancelListener(() -> {
               implementation.dismiss();
               JSObject ret = new JSObject();
               ret.put("index", -1);
               call.resolve(ret);
           });
```

### Alternatives
None

### Additional Context
None

funkyvisions:
I also discovered on iOS for iPad that you need to do this inside of ActionSheetPlugin.swift
```
         var alertActions = [UIAlertAction]()
+        alertActions.append(UIAlertAction(title: "", style: .cancel, handler: {(_) -> Void in call.resolve(["index": -1])}))
```

eric-horodyski:
This thread has been inactive for some time, and we’re cleaning up older items to keep the backlog focused and manageable.

Details on the effort:
[https://ionic.io/blog/keeping-capacitors-backlog-healthy](https://ionic.io/blog/keeping-capacitors-backlog-healthy)

If this is still relevant today, feel free to open a new issue with current information and we’ll review it.

PR #2285, opened 2024-12-17
feat(action-sheet): Add cancelable parameter

OS-pedrogustavobilro:
Add parameter to allow action-sheet to be cancelable (clicking outside the sheet, and using the system back button on Android), and in case it is canceled, return it in the result.

Notes:

- There's some slight differences in behavior for iOS, because it supports the `ActionSheetButtonStyle` property that other platforms do not - explained in the doc updates that are also in this PR.
- For this to work with PWAs, it requires https://github.com/ionic-team/pwa-elements/pull/138

Fixes https://github.com/ionic-team/capacitor-plugins/issues/2154

### DIFF
--- a/action-sheet/ios/Sources/ActionSheetPlugin/ActionSheetPlugin.swift
+++ b/action-sheet/ios/Sources/ActionSheetPlugin/ActionSheetPlugin.swift
@@ -35,7 +35,8 @@ public class ActionSheetPlugin: CAPPlugin, CAPBridgedPlugin {
             }
             let action = UIAlertAction(title: title, style: buttonStyle, handler: { (_) -> Void in
                 call.resolve([
-                    "index": index
+                    "index": index,
+                    "canceled": false
                 ])
             })
             alertActions.append(action)
@@ -48,7 +49,10 @@ public class ActionSheetPlugin: CAPPlugin, CAPBridgedPlugin {
                     if (forceCancelableOnClickOutside) {
                         let gestureRecognizer = TapGestureRecognizerWithClosure {
                             alertController.dismiss(animated: true, completion: nil)
-                            call.reject("User canceled action sheet")
+                            call.resolve([
+                                "index": -1,
+                                "canceled": true
+                            ])
                         }
                         let backroundView = alertController.view.superview?.subviews[0]
                         backroundView?.addGestureRecognizer(gestureRecognizer)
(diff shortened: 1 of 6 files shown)


action-sheet/ios/Sources/ActionSheetPlugin/ActionSheetPlugin.swift
```diff
@@ -17,21 +17,26 @@ public class ActionSheetPlugin: CAPPlugin, CAPBridgedPlugin {
     @objc func showActions(_ call: CAPPluginCall) {
         let title = call.options["title"] as? String
         let message = call.options["message"] as? String
+        let cancelable = call.options["cancelable"] as? Bool ?? false
```
OS-ricardomoreirasilva, line 20: I believe there's a `getBool` with a default value within the `CAPPluginCall`, maybe we can use it.

alexgerardojacinto: Isn't there a `getString` as well?

alexgerardojacinto: Not super important, but I think if we change one we should change the others to keep the way we access `CAPPluginCall's` parameters consistent.

OS-pedrogustavobilro: Done in 99443aeefa17c7281542b810aac5837e512edb82

action-sheet/ios/Sources/ActionSheetPlugin/ActionSheetPlugin.swift
```diff
@@ -15,23 +15,28 @@ public class ActionSheetPlugin: CAPPlugin, CAPBridgedPlugin {
...
+        let title = call.getString("title")
+        let message = call.getString("message")
+        let cancelable = call.getBool("cancelable", false)
 
         let options = call.getArray("options", JSObject.self) ?? []
         var alertActions = [UIAlertAction]()
+        var forceCancelableOnClickOutside = cancelable
         for (index, option) in options.enumerated() {
             let style = option["style"] as? String ?? "DEFAULT"
             let title = option["title"] as? String ?? ""
```
OS-ricardomoreirasilva, line 27: If we're now touching the `call.options[...] ...` calls, shouldn't these be considered as well?

OS-pedrogustavobilro: Is there a method for `getString` for `JSObject`? It makes sense for there to be, but xcode is saying there is no such method

OS-ricardomoreirasilva: Nevermind, my mistake.

OS-ricardomoreirasilva (approved):

alexgerardojacinto (approved):

OS-pedrogustavobilro:
Hey @alexgerardojacinto @jcesarmobile I had to make some adjustments to the iOS implementation, due to changes in iOS 26 (and also fixing some stuff that was broken in the old implementation), also updated the PR description, feel free to review whenever you can 🙏

PR #2287, opened 2024-12-20
fix(android): run downloadFile asynchronously

ItsChaceD:
Closes #6861

This PR runs `downloadFile` in an async task so it does not block other plugin calls.

### DIFF
--- a/filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/Filesystem.java
+++ b/filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/Filesystem.java
@@ -3,6 +3,8 @@ package com.capacitorjs.plugins.filesystem;
 import android.content.Context;
 import android.net.Uri;
 import android.os.Environment;
+import android.os.Handler;
+import android.os.Looper;
 import android.util.Base64;
 import com.capacitorjs.plugins.filesystem.exceptions.CopyFailedException;
 import com.capacitorjs.plugins.filesystem.exceptions.DirectoryExistsException;
@@ -27,11 +29,9 @@ import java.nio.channels.FileChannel;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Locale;
-import org.json.JSONException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import android.os.Handler;
-import android.os.Looper;
+import org.json.JSONException;
 
 public class Filesystem {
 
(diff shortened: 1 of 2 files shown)


alexgerardojacinto (approved):

filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/Filesystem.java
```diff
@@ -303,81 +307,123 @@ public void copyRecursively(File src, File dst) throws IOException {
         }
     }
 
-    public JSObject downloadFile(PluginCall call, Bridge bridge, HttpRequestHandler.ProgressEmitter emitter)
-        throws IOException, URISyntaxException, JSONException {
+    public void downloadFile(PluginCall call, Bridge bridge, HttpRequestHandler.ProgressEmitter emitter,
+            FilesystemDownloadCallback callback) throws IOException, URISyntaxException, JSONException {
```
OS-pedrogustavobilro, line 311: I believe this method no longer throws the mentioned exceptions, but rather returns them in `FilesystemDownloadCallback`, therefore we should update the signature

```suggestion
    public void downloadFile(PluginCall call, Bridge bridge, HttpRequestHandler.ProgressEmitter emitter,
            FilesystemDownloadCallback callback) {
```

filesystem/android/src/main/java/com/capacitorjs/plugins/filesystem/Filesystem.java
```diff
@@ -303,81 +307,123 @@ public void copyRecursively(File src, File dst) throws IOException {
...
+
+        executor.execute(() -> {
+            AsyncTaskResult result = doDownloadInBackground(urlString, call, bridge, emitter);
+            handler.post(() -> {
+                if (result.error != null) {
+                    callback.onError(result.error);
+                } else {
+                    callback.onSuccess(result.result);
+                }
+            });
```
OS-pedrogustavobilro, line 324: It seems to me that it is unnecessary to have this `AsyncTaskResult`, since it either just wraps the result or error if it exists.

I feel like we could lift the try-catch out of `doDownloadInBackground`, and have it return the `JSObject` directly

```suggestion
            try {
                JSObject result = doDownloadInBackground(urlString, call, bridge, emitter);
                handler.post(() -> callback.onSuccess(result));
            } catch (Exception error) {
                handler.post(() -> callback.onError(error));
            }
```

Also, if you do this change, add the `throws IOException, URISyntaxException, JSONException` to the `doDownloadInBackground` signature.

What do you think?

OS-ricardomoreirasilva (approved):

OS-pedrogustavobilro (approved):

PR #2293, opened 2025-01-09
Add check for invalid use of multiple scheduler types

orjandh:
Added a check for invalid use of multiple scheduler types in the capacitor-local-notifications plugin. I think this is a good idea as the scheduler types are checked in an arbitrary order and it might not be clear from a developers side what would happen. I decided to handle the error in similar style as the rest of the function for the respective platform.

I also noticed the handling of the scheduled time being after current time check is handled with an exception on ios, but on android it is handled by exiting out of the function and logging an error, which seems inconsistent.


eric-horodyski:

More context:

PR #2305, opened 2025-01-30
Performance improvement for Camera plugin.

KirillBorunov:
Rotating the bitmap is expensive on some devices. Doing the resize before the rotation reduces notably the delay on these devices.


jcesarmobile:
camera plugin has been moved to https://github.com/ionic-team/capacitor-camera and it's being reworked

no more work will be done in this repository, if the PR is still relevant, please, reopen it on https://github.com/ionic-team/capacitor-camera repository

PR #2307, opened 2025-02-05
fix(filesystem): unexpected error when saving file lead to crashing the app

guitarooman14:
When you call the writefile plugin method from ionic with corrupted data or with not base64 data, it leads to a total crash of the app because the exception is not caught by the writefile method and it is only caught by the bridge.java that rethrow a runtimeException.

` catch (Exception ex) {
    Logger.error("Serious error executing plugin", ex);
    throw new RuntimeException(ex);
}`


OS-pedrogustavobilro:
Hey @guitarooman14, I don't think this issue happens in the latest version of the file plugin - 8.X

Therefore I'm closing this issue. If you think it happens - please open an issue in the new repository for filesystem - https://github.com/ionic-team/capacitor-filesystem

We do appreciate your contribution!

PR #2309, opened 2025-02-06
Fix Android Camera crash when opening photo gallery with limit

joewoodhouse:
Prevents a crash where if the user specifies a `limit` to `pickImages` that is above the system's maximum limit. Currently supplying a limit value of, say, 10,000 will result in crash with the following exception:

> Max items must be less or equals MediaStore.getPickImagesMaxLimit()

This change makes the check as suggested in the error.


eric-horodyski:

More context:

Issue #1835, opened 2023-10-09
@capacitor/filesystem downloadFile() fails to create missing directories under android


longinius:
## Bug Report

The [`downloadFile()`](https://capacitorjs.com/docs/apis/filesystem#downloadfile) method doesn't create missing directories under android.

### Plugin(s)
@capacitor/filesystem@5.1.4

### Capacitor Version
```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 5.4.2
  @capacitor/core: 5.4.2
  @capacitor/android: 5.4.2
  @capacitor/ios: 5.4.2

Installed Dependencies:

  @capacitor/core: 5.4.2
  @capacitor/android: 5.4.2
  @capacitor/cli: 5.4.2
  @capacitor/ios: 5.4.2

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)
Android

### Current Behavior
Download fails on android, when setting the option `recursive` to `true` and providing a `path` with subdirectories that doesn't exist.

```typescript
Filesystem.downloadFile({
  url: 'https://images.unsplash.com/photo-1510158105534-9b01f2462ce1',
  path: '/path/to/file/image.jpg',
  directory: Directory.Cache,
  recursive: true,
});
```

The above code will result in the following error.
```
ERROR Error: Uncaught (in promise): Error: Error downloading file: /data/user/0/io.ionic.starter/cache/path/to/file/image.jpg: open failed: ENOENT (No such file or directory)
Error: Error downloading file: /data/user/0/io.ionic.starter/cache/path/to/file/image.jpg: open failed: ENOENT (No such file or directory)
```

### Expected Behavior
The missing folders should be created and the file should be saved at the specified path. On iOS, the missing folders are created.

### Code Reproduction
[Repository with minimal reproduction](https://github.com/longinius/ionic-capacitor-filesystem)

### Other Technical Details
Tested on native android device with android version 13.

Changing or omitting the `directory` option does not change the behavior, only the error message becomes slightly different. If you create the folder beforehand with the `mkdir` function, the download also works.

### Additional Context
<details><summary>Error in Logcat</summary>
<pre>
Error downloading file: /data/user/0/io.ionic.starter/cache/path/to/file/image.jpg: open failed: ENOENT (No such file or directory)
[...]

PR #2324, opened 2025-03-04
fix(filesystem): respect downloadFile recursive option on Android and iOS

kbremner:
Resolution for https://github.com/ionic-team/capacitor-plugins/issues/1835

While on Android an error is thrown because missing directories are never created, the iOS implementation ignores the recursive option by always creating directories if they don't exist.

Both implementations updated to respect the recursive parameter.


OS-pedrogustavobilro:
Thank you for the PR. However we ended up fixing it differently, so we'll be closing this. We apologize for the inconvenience, but appreciate your contribution!

Context: https://github.com/ionic-team/capacitor-plugins/issues/1835#issuecomment-3062722644

PR #2325, opened 2025-03-10
fix: iOS push notifications alert --> banner

ThatOneCalculator:
The `alert` notification type has been deprecated since iOS 14, and should now be replaced by `banner`.


ThatOneCalculator:
Requesting review from @jcesarmobile

jcesarmobile:
thanks for the PR, but `.banner` is not a direct equivalent for `.alert`, it was split into `.banner` and `.list` for a more granular control, so the exact equivalent would be to request both.

I've reported this internally and we will probably deprecate/remove `alert` and add two new options, "banner" and "list", to mimic the native values.

PR #2340, opened 2025-04-07
fix(@capacitor/haptics): prevent thread overload by reusing CHHapticEngine

GorkaMinus:
### Fix: Prevent iOS app freeze by reusing `CHHapticEngine` instance

This PR addresses an issue where calling `Haptics.vibrate()` rapidly (e.g., from a fast-scrolling picker in my case) causes the app to hang or crash on iOS. The problem stems from creating a new `CHHapticEngine` instance on each call, which leads to excessive thread creation and eventual system overload.

#### Solution
- Reuses a single `CHHapticEngine` instance across multiple vibration calls.
- Ensures the engine is only created and started when necessary.
- Handles engine lifecycle and failure states gracefully.

This fix improves performance and stability, especially in high-frequency vibration scenarios.

### DIFF
--- a/haptics/ios/Sources/HapticsPlugin/Haptics.swift
+++ b/haptics/ios/Sources/HapticsPlugin/Haptics.swift
@@ -6,6 +6,46 @@ import CoreHaptics
 
     var selectionFeedbackGenerator: UISelectionFeedbackGenerator?
 
+    private var hapticEngine: CHHapticEngine?
+    private var idleTimer: Timer?
+    private let idleInterval: TimeInterval = 5.0
+
+    private func initializeEngine() {
+        do {
+            let engine = try CHHapticEngine()
+            engine.resetHandler = { [weak self] in
+                do {
+                    try self?.hapticEngine?.start()
+                } catch {
+                    self?.hapticEngine = nil
+                }
+            }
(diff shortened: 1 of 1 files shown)


OS-pedrogustavobilro:
Hey @GorkaMinus, sorry for the delay, I changed the branch to 7.x, as since 8.x, the capacitor haptics plugin now lives in its own repository - https://github.com/ionic-team/capacitor-haptics.

If you can, open an equivalent PR to that repository in order to be able to merge it into the latest version.

If you have any questions feel free to ask, thank you!

jcesarmobile (changes requested):

Ionitron:
It looks like this issue didn't get the information it needed, so I'll close it for now. If I made a mistake, sorry! I am just a bot.

Have a great day!
Ionitron 💙

GorkaMinus:
Opened the equivalent PR in https://github.com/ionic-team/capacitor-haptics/pull/24

jcesarmobile:
can you address the requested changes here too? we are still maintaining version 7.x, so both PRs should be the same

GorkaMinus:
> can you address the requested changes here too? we are still maintaining version 7.x, so both PRs should be the same

@jcesarmobile - done, both PRs are now updated with the requested changes

PR #2358, opened 2025-05-04
docs(README.md): bolden important info

rbalet:
Boldened that there is no need to add the Firebase SDK, as this doesn't make that much sense and you're not always reading the entire text before starting.


OS-pedrogustavobilro:
Thanks for the PR @rbalet!

PR #2381, opened 2025-06-26
chore(android): Publishing migration to Central Portal

OS-pedrogustavobilro:
We currently publish in Maven Central through OSSRH, however this needs to be changed due OSSRH being shutdown in after June 30th 2025.

Migration of `com.capacitor` namespace to Central Portal has already been done, this PR takes care of updating the publishing logic - which uses a new token that is now accessed in different github secrets.

Relevant References:

- https://central.sonatype.org/faq/what-is-different-between-central-portal-and-legacy-ossrh/
- https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-central

### DIFF
--- a/scripts/publish-android.sh
+++ b/scripts/publish-android.sh
@@ -42,7 +42,7 @@ publish_plugin () {
                 if grep --quiet "BUILD SUCCESSFUL" $LOG_OUTPUT; then
                     printf %"s\n\n" "Success: $PLUGIN_NAME published to MavenCentral."
                 else
-                    printf %"s\n\n" "Error publishing $PLUGIN_NAME, check $LOG_OUTPUT for more info! Manual publication review may be necessary at the Sonatype Repository Manager https://s01.oss.sonatype.org/"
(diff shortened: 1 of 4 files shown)


alexgerardojacinto (approved):

scripts/publish-android.sh
```diff
@@ -42,7 +42,7 @@ publish_plugin () {
                 if grep --quiet "BUILD SUCCESSFUL" $LOG_OUTPUT; then
                     printf %"s\n\n" "Success: $PLUGIN_NAME published to MavenCentral."
                 else
-                    printf %"s\n\n" "Error publishing $PLUGIN_NAME, check $LOG_OUTPUT for more info! Manual publication review may be necessary at the Sonatype Repository Manager https://s01.oss.sonatype.org/"
+                    printf %"s\n" "Error publishing, check $LOG_OUTPUT for more info! Manually review and release from the Central Portal may be necessary https://central.sonatype.com/publishing/deployments/"
```
eric-horodyski, line 45: Can we adjust the beginning of the print statement here to:

```
printf %"s\n\n" "Error publishing $PLUGIN_NAME, check $LOG_OUTPUT for more info!
```

Just so that the extra line break and name of the plugin are retained in the new error message.

PR #2382, opened 2025-07-03
fix(android): fixed issues on android 15+ white space above keyboard

OS-kepatotorica:
On android 15+ we need to take the gesture bar into account (as well as the three button nav)

Closes: https://github.com/kensodemann/issue-53807

Testing notes (screenshots on Jira ticket):
Pixel 9 (physical, gesture navigation) →  ✅ 
Pixel 9 (physical, 3 button navigation) → ✅   
Pixel 8 pro emulator →  ✅ 
Pixel 4 (android 14, emulator) → ✅   
Pixel 4 (android 13, emulator) →  ✅ 
Pixel 7a (android 15, emulator) →  ✅

### DIFF
--- a/.github/workflows/publish-android.yml
+++ b/.github/workflows/publish-android.yml
@@ -3,9 +3,9 @@ name: Publish Native Android Libraries
 on:
   workflow_call:
     secrets:
-      ANDROID_OSSRH_USERNAME:
+      ANDROID_CENTRAL_USERNAME:
         required: true
-      ANDROID_OSSRH_PASSWORD:
+      ANDROID_CENTRAL_PASSWORD:
         required: true
       ANDROID_SIGNING_KEY_ID:
         required: true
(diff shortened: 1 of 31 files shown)


benmccarty91 (approved):

jcesarmobile:
Doesn't look like the height is being calculated correctly yet.

If you add a fixed element, in example `<div style="background-color:blue; position: absolute; bottom: 0; height:20px">bottom</div>`, when the keyboard is present it shows only half of it.

Also if you change the `adjustMarginsForEdgeToEdge` to `none` (or just delete the config) the fixed element is not visible at all.
It's also not visible if the app's style has `<item name="android:windowOptOutEdgeToEdgeEnforcement" tools:ignore="NewApi">true</item>` (which OutSystems apps have)

OS-kepatotorica:
> Doesn't look like the height is being calculated correctly yet.
> 
> If you add a fixed element, in example `<div style="background-color:blue; position: absolute; bottom: 0; height:20px">bottom</div>`, when the keyboard is present it shows only half of it.
> 
> Also if you change the `adjustMarginsForEdgeToEdge` to `none` (or just delete the config) the fixed element is not visible at all. It's also not visible if the app's style has `<item name="android:windowOptOutEdgeToEdgeEnforcement" tools:ignore="NewApi">true</item>` (which OutSystems apps have)

I will look into this. I was able to confirm what you said about the blue bottom div being cut in half.

OS-kepatotorica:
> Doesn't look like the height is being calculated correctly yet.
> 
> If you add a fixed element, in example `<div style="background-color:blue; position: absolute; bottom: 0; height:20px">bottom</div>`, when the keyboard is present it shows only half of it.
> 
> Also if you change the `adjustMarginsForEdgeToEdge` to `none` (or just delete the config) the fixed element is not visible at all. It's also not visible if the app's style has `<item name="android:windowOptOutEdgeToEdgeEnforcement" tools:ignore="NewApi">true</item>` (which OutSystems apps have)

Hey Julio, I think this should be addressed now.

OS-kepatotorica:
> changing the public constructor is a breaking change, so we can't do this in a patch release

Saw this comment after merging, reverted and discussing next steps.

PR #2388, opened 2025-07-15
fix(android): fixed issues on android 15+ white space above keyboard

OS-kepatotorica:
On android 15+ we need to take the gesture bar into account (as well as the three button nav)

Closes: https://github.com/kensodemann/issue-53807

Testing notes (screenshots on Jira ticket):
Pixel 9 (physical, gesture navigation) → ✅
Pixel 9 (physical, 3 button navigation) → ✅
Pixel 8 pro emulator → ✅
Pixel 4 (android 14, emulator) → ✅
Pixel 4 (android 13, emulator) → ✅
Pixel 7a (android 15, emulator) → ✅

### DIFF
--- a/keyboard/android/src/main/java/com/capacitorjs/plugins/keyboard/Keyboard.java
+++ b/keyboard/android/src/main/java/com/capacitorjs/plugins/keyboard/Keyboard.java
@@ -2,7 +2,9 @@ package com.capacitorjs.plugins.keyboard;
 
 import android.content.Context;
 import android.graphics.Rect;
+import android.os.Build;
 import android.util.DisplayMetrics;
+import android.util.TypedValue;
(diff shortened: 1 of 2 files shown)


luisbytes:
@OS-kepatotorica, @jcesarmobile When is the stable plugin released on npmjs?

tafelnl:
The Capacitor docs says

> There is an Android bug that prevents the keyboard from resizing the WebView when the app is in full screen (i.e. if StatusBar plugin is used to overlay the status bar). This setting, if set to true, add a workaround that resizes the WebView even when the app is in full screen. Only available for Android

Could you explain to me what this bug is? And would it be a valid approach to try and git this fixed by the Android dev team? Or have you already tried that?

PR #2390, opened 2025-07-25
feat(app): Add support enabling predictive back gesture on Android

theproducer:
This PR adds a function to toggle on and off support for Android 15+ predictive back gesture, allowing developers to use that gesture functionality at the right time of their choosing in their application.

`togglePredictiveBack()` in essence temporarily disables the `OnBackPressedCallback` listener.

### DIFF
--- a/app/android/src/main/java/com/capacitorjs/plugins/app/AppPlugin.java
+++ b/app/android/src/main/java/com/capacitorjs/plugins/app/AppPlugin.java
@@ -25,6 +25,8 @@ public class AppPlugin extends Plugin {
     private static final String EVENT_RESUME = "resume";
     private boolean hasPausedEver = false;
 
+    private OnBackPressedCallback onbackPressedCallback;
+    
     public void load() {
         bridge
             .getApp()
@@ -59,7 +61,10 @@ public class AppPlugin extends Plugin {
                 }
             }
         };
+
         getActivity().getOnBackPressedDispatcher().addCallback(getActivity(), callback);
+
+        this.onbackPressedCallback = callback;
     }
 
     @PluginMethod
@@ -109,6 +114,19 @@ public class AppPlugin extends Plugin {
     @PluginMethod
     public void minimizeApp(PluginCall call) {
         getActivity().moveTaskToBack(true);
+        call.resolve();
+    }
+
+    @PluginMethod
+    public void togglePredictiveBack(PluginCall call) {
+        if (this.onbackPressedCallback == null) {
+            call.resolve();
+            return;
+        }
+
+        Boolean enabled = call.getBoolean("enabled");
+
+        this.onbackPressedCallback.setEnabled(!enabled);
         call.resolve();
     }
 
--- a/app/src/definitions.ts
+++ b/app/src/definitions.ts
@@ -126,6 +126,11 @@ export interface BackButtonListenerEvent {
 }
 
 export interface TogglePredictiveBackOptions {
+  /**
+   * Indicates whether to enable or disable predictive back gesture support.
+   *
+   * @since 8.0.0
+   */
   enabled: boolean;
 }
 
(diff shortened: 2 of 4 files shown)


app/android/src/main/java/com/capacitorjs/plugins/app/AppPlugin.java
```diff
@@ -112,6 +117,19 @@ public void minimizeApp(PluginCall call) {
         call.resolve();
     }
 
+    @PluginMethod
+    public void togglePredictiveBack(PluginCall call) {
+        if (this.onbackPressedCallback == null) {
+            call.resolve();
```
jcesarmobile, line 123: shouldn't this reject instead of resolve?

app/src/definitions.ts
```diff
@@ -171,6 +180,17 @@ export interface AppPlugin {
    */
   minimizeApp(): Promise<void>;
 
+  /**
+   * Toggles support for Android's predictive back gesture.
+   *
+   * Enabling this will disable this plugin's `backButton` listener.
+   *
+   * Only available for Android 15+.
```
jcesarmobile, line 188: isn't pre predictive back available from Android 13+?

jcesarmobile (changes requested):
Instead of adding the `this.onbackPressedCallback = callback;` new line, set it on the callback creation changing `OnBackPressedCallback callback = new OnBackPressedCallback(true)` to `this.onbackPressedCallback = new OnBackPressedCallback(true)` (will also need to replace other usages of `callback to `this.onbackPressedCallback`

While this is supposed to enable the predictive back on Android 15+, this disables the back button handling for all Android versions, so I think that should be documented as for the current comment, it appears to only have an effect on Android 15+, but it affects all Android versions.

I don't fully understand what the "predictive back" is supposed to do, I see no difference on enabling it or disabling it, other than the back button handling not being into effect, which leads to the app being minimized instead of navigating back.

jcesarmobile:
Also, I think this should be configurable from the capacitor config file to disable it from the beginning if desired, and the name should probably be about disabling back button handling as this is what the PR really does, which is required for predictive back to work, but you are disabling the back button handling in the end and people might want to use the app plugin just for the app information and not for the back button handling.

theproducer:
> Also, I think this should be configurable from the capacitor config file to disable it from the beginning if desired, and the name should probably be about disabling back button handling as this is what the PR really does, which is required for predictive back to work, but you are disabling the back button handling in the end and people might want to use the app plugin just for the app information and not for the back button handling.

I think you are right, especially the part about people may not wanting the back button handling from the plugin at all.  

I'll make those changes, and I'll keep the toggle as well (renaming it to be more generic) in case there are scenarios where devs want to temporarily enable / disable back button handling based on the state / routing of their app.

jcesarmobile (approved):

xEverth:
Hello Ionic Team!
Is this the place to ask for clarifications on the Hardware Back Button on an Ionic application?

The Ionic v8 documentation for Angular invites to use Platform.backButton.subscribe / subscribeWithPriority for the back button events on Android and that it's necessary to install this plugin @capacitor/app in order to have support for hardware back button in a Capacitor native app:
https://ionicframework.com/docs/developing/hardware-back-button

In my application however I find that those events are not fired, and I need to subscribe to backButton events using App.addListener('backButton'), which does not have the same priority-based handling.

If I set disableBackButtonHandler option introduced in this PR (and available since @capacitor/app@^7.1.0) to true in capacitor config, the aforementioned App listener does not fire anymore, but neither does the one in the Ionic documentation and instead the app is exited directly (even if I navigated to different pages in-app, using the Angular router).

Before I set up a sample app for reproduction, may I ask you what is the expected behavior? Should the two APIs (Ionic's and Capacitor/App) be able to used independently or does the latter disable the former?

Thank you

Issue #2337, opened 2025-03-31
Capacitor status-bar v7 breaks app resizing with iOS split screen


aeharding:
### Plugin(s)

@capacitor/status-bar

### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 7.1.0
  @capacitor/core: 7.1.0
  @capacitor/android: 7.1.0
  @capacitor/ios: 7.1.0

Installed Dependencies:

  @capacitor/android: not installed
  @capacitor/cli: 7.1.0
  @capacitor/core: 7.1.0
  @capacitor/ios: 7.1.0

[success] iOS looking great! 👌
```

### Platform(s)

- iOS

### Current Behavior

When using iOS split screen, the app doesn't resize properly.

To reproduce:

- Create a new capacitor project
- Add the status bar plugin
- Open split screen view, see doesn't resize properly

This used to work in capacitor v6, so this is a regression.

### Expected Behavior

Should resize properly, like capacitor v6.

### Code Reproduction

https://github.com/aeharding/resize-bug-spm

### Other Technical Details

- https://github.com/aeharding/voyager/issues/1932
- https://github.com/aeharding/voyager/issues/1882

### Additional Context

https://github.com/user-attachments/assets/d84bcd70-2076-4e5b-bcc7-3706d0657f7a

lubomirfiala:
I'm having same issue with ipadOS stage manager

masonicboom:
https://github.com/ionic-team/capacitor-plugins/pull/2384 fixes this for me

OS-pedrogustavobilro:
Hey everyone, this should be fixed in version 7.0.3 of `capacitor/status-bar`. Will be closing this issue. Thank you!

PR #2394, opened 2025-08-07
fix(statusbar): using window size instead of screen size for resize bounds

theproducer:
This PR fixes an issue on iPad 26 that prevents the WebView / Status Bar from properly resizing when using the new windowing mode.


alexgerardojacinto (approved):

latata:
It seems to fix https://github.com/ionic-team/capacitor-plugins/issues/2337 along with some related issues. When do you plan to release it?

OS-pedrogustavobilro:
@latata this fix has been released in version 7.0.3 of the plugin.

PR #2395, opened 2025-08-08
fix(docs): update Capacitor plugin links in README for accuracy

riderx:


OS-pedrogustavobilro:
Thanks for the PR @riderx!

PR #2403, opened 2025-08-30
fix(ios): prevent black QuickType bar when using Magic Keyboard on iPad

Daniele-rolli:
## Context

When using an iPad with a hardware keyboard (e.g. Magic Keyboard), iOS still fires `UIKeyboardWillShow / UIKeyboardDidShow` even though no software keyboard is rendered. Instead, only the QuickType suggestion bar is displayed at the bottom of the screen.

Currently, Capacitor interprets this as a "real keyboard" event and resizes the `WKWebView`, this results in a black bar or visual glitch below the webview.

| Before    | After     | 
| ------- | ------------ |
| IMG_0059 | IMG_0062|

## Solution

Add a height threshold guard inside onKeyboardWillShow and onKeyboardDidShow in the iOS KeyboardPlugin:
If `UIDevice.userInterfaceIdiom == .pad` and `keyboardHeight < 100px`, treat it as a QuickType event:

- Do **not resize** the webview.
- Fire JS events with `keyboardHeight = 0` to indicate "no real keyboard."

Otherwise, continue with normal resizing behavior.

**This ensures:**
- iPad + Magic Keyboard: no resize, no black bar.
- iPad + software keyboard: resize works normally.
- iPhone: unchanged.

### DIFF
--- a/keyboard/ios/Sources/KeyboardPlugin/Keyboard.m
+++ b/keyboard/ios/Sources/KeyboardPlugin/Keyboard.m
@@ -48,6 +48,20 @@ typedef enum : NSUInteger {
 // protocol conformance for this class is implemented by a macro and clang isn't detecting that
 @implementation KeyboardPlugin
 
+/// Heights below this on iPad are treated as QuickType bar, not a real keyboard
+static const CGFloat QUICKTYPE_IGNORE_THRESHOLD = 100.0;
+
+- (BOOL)isIPad {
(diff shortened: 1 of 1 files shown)


eric-horodyski:
Thanks for opening this PR 🙏

Details: https://ionic.io/blog/keeping-capacitors-backlog-healthy

This PR will need to be applied to the dedicated Keyboard plugin repository: https://github.com/ionic-team/capacitor-keyboard/

PR #2434, opened 2025-11-07
chore(ios): update spm from 7.0.0 to 8.0.0-beta

OS-ruimoreiramendes:
Updated `capacitor-swift-pm.git` reference in all plugin `Package.swift` files to `8.0.0-beta`.


markemer:
I've got some ideas on automating this, hopefully we can start doing that with the GA of Capacitor 8.

PR #2435, opened 2025-11-07
chore(android): update Gradle dependencies to latest versions

alexgerardojacinto:
- Updates Gradle dependencies of all plugins to latest
- Tested in `capacitor-testapp` repository.


alexgerardojacinto:
> The new non common default values (material, firebase, etc.) also need to be updated on the README of the plugins that use them.

@jcesarmobile thanks for the heads up! I updated the README files accordingly, can you check?

PR #2440, opened 2025-11-11
chore: upgrade npm dependencies for all plugins

ItsChaceD:
* Updated NPM dependencies across all plugins and root package, including major version upgrade to typescript (`5.9.3`). Prettier updates will be in a separate PR
* Verified all plugins build successfully and have no peer dependency conflicts


OS-ruimoreiramendes (approved):

.github/actions/changed-packages/action.yml
```diff
@@ -8,5 +8,5 @@ outputs:
   paths:
     description: An array of relative paths to packages that were changed
 runs:
-  using: node12
+  using: node20
```
OS-pedrogustavobilro: Minor: Since we're using Node 22 elsewhere, should we use it here as well-'

ItsChaceD:
> `@ionic/prettier-config`, `prettier` and `prettier-plugin-java` should be updated. The `npx @capacitor/plugin-migration-v6-to-v7@latest` command should be able to assist you with the required changes. Or you can check [ionic-team/create-capacitor-plugin@2fa0317](https://github.com/ionic-team/create-capacitor-plugin/commit/2fa031725e72c73dc40786219ca43a60a1589d03) for manual changes and probably this too [ionic-team/create-capacitor-plugin#123 (files)](https://github.com/ionic-team/create-capacitor-plugin/pull/123/files)
> 
> Since you updated the actions to not use `esm` you can probably remove the dependency

If we upgrade those to their latest versions then it results in 100+ code changes when linting and I wasn't sure if we wanted to do that days before a beta release. Yes they should all just be style / cosmetic changes, but didn't know if we wanted to take that risk. I also wanted to discuss and potentially add it to a separate PR so that it doesn't add too much noise to this PR review.  

Unless you are only talking about doing minor upgrades and adding the `prettier-plugin-java` command to the fmt script?

Thoughts?

jcesarmobile:
I mean major updates, not minor, if you do a minor then the `prettier-plugin-java` command on the npm script is not needed as it's needed on version 2 of `prettier-plugin-java`.

I'm ok on doing it in a separate PR, but the changes on the root package.json are also unrelated to the original task, so would be good to also move them to a separate PR.

ItsChaceD:
> I mean major updates, not minor, if you do a minor then the `prettier-plugin-java` command on the npm script is not needed as it's needed on version 2 of `prettier-plugin-java`.
> 
> I'm ok on doing it in a separate PR, but the changes on the root package.json are also unrelated to the original task, so would be good to also move them to a separate PR.

Opened #2443 (root npm updates) and #2444 (prettier updates)

OS-pedrogustavobilro (approved):

PR #2442, opened 2025-11-12
fix(status-bar)!: remove notification related extensions

jcesarmobile:
Marking it as breaking in case other plugins were listening for those notifications, but the notifications are now provided by`@capacitor/ios`


jcesarmobile:
There won't be double notification, these extensions on status-bar plugin were preventing the equivalent `CAPBridgeViewController` methods from being called, so only one fires, but not the correct one.

PR #2443, opened 2025-11-12
chore: update root folder npm dependencies

ItsChaceD:


package.json
```diff
@@ -22,19 +22,18 @@
...
+    "@actions/http-client": "^2.2.3",
     "@ionic/prettier-config": "^1.0.1",
-    "@types/node": "~18.11.19",
-    "@types/prompts": "^2.0.8",
-    "cross-spawn": "^7.0.3",
-    "esm": "^3.2.25",
-    "glob": "^10.3.3",
-    "lerna": "^7.1.3",
-    "nx": "^16.3.1",
+    "@types/node": "^22.19.0",
```
OS-pedrogustavobilro: Notice you put `^24.10.1` in https://github.com/ionic-team/capacitor-plugins/pull/2440 but `^22.19.0` here. Which one should it be?

PR #2444, opened 2025-11-12
chore: update prettier across all plugins

ItsChaceD:
* Updates `prettier` to `^3.6.2` in all plugins
* Updates `@ionic/prettier-config` to `^4.0.0` in all plugins
* Updates `prettier-plugin-java` to `^2.7.7` in all plugins
* Updates `prettier` command syntax to use `prettier-plugin-java`
* Removes `.prettierignore` from all plugins since it now respects `.gitignore` and is not necessary


markemer:
Do we already do this in cap? Or does that need doing also?

ItsChaceD:
> Do we already do this in cap? Or does that need doing also?

Looks like it was already done in Capacitor awhile ago (https://github.com/ionic-team/capacitor/pull/7710)

markemer (approved):

PR #2441, opened 2025-11-12
docs(statusbar): Android 16+ behavior change for overlaysWebView and backgroundColor

andredestro:

PR #2447, opened 2025-11-13
feat(status-bar)!: update foreground color of status bar on `setBackgroundColor` in Android

OS-ruimoreiramendes:
This PR updates the StatusBar Android implementation to correctly handle status bar behavior across different Android versions and target SDK, especially for **Android 15** and **Android 16**.

- Update `setBackgroundColor` to make sure that the foreground color of the status bar is set to either Light or Dark depending on the background color.
- Only set background and foreground color of status bar on Android versions where they can actually be set - below Android 15 or Android 15 with edge to edge opt out.

Manually tested on Android with different Android versions and `targetSdkVersion` to ensure correct behavior:
- **Android 12**: StatusBar background and foreground colors work as expected.
- **Android 15**: depends on `windowOptOutEdgeToEdgeEnforcement`, sets background/foreground when opt-out is enabled.
- **Android 16**: No effect.

### DIFF
--- a/status-bar/README.md
+++ b/status-bar/README.md
@@ -159,6 +159,7 @@ setBackgroundColor(options: BackgroundColorOptions) => Promise<void>
 ```
 
 Set the background color of the status bar.
+Calling this function updates the foreground color of the status bar if the style is set to default, except on iOS versions lower than 17.
 Not available on Android 15+.
 
 | Param         | Type                                                                      |
--- a/status-bar/src/definitions.ts
+++ b/status-bar/src/definitions.ts
@@ -170,6 +170,7 @@ export interface StatusBarPlugin {
 
   /**
    * Set the background color of the status bar.
+   * Calling this function updates the foreground color of the status bar if the style is set to default, except on iOS versions lower than 17.
    * Not available on Android 15+.
    *
    * @since 1.0.0


jcesarmobile:
Is this going to be backported to version 7 of the plugin?
If not, this scenario is not needed
> Android 16 (targetSdk 35): depends on windowOptOutEdgeToEdgeEnforcement, sets background/foreground when opt-out is enabled.

Version 8 of the plugin should only be used with targetSDK 36.

BTW, there is another PR for updating the documentation already
https://github.com/ionic-team/capacitor-plugins/pull/2441

alexgerardojacinto:
> Is this going to be backported to version 7 of the plugin? If not, this scenario is not needed
> 
> > Android 16 (targetSdk 35): depends on windowOptOutEdgeToEdgeEnforcement, sets background/foreground when opt-out is enabled.
> 
> Version 8 of the plugin should only be used with targetSDK 36.
> 
> BTW, there is another PR for updating the documentation already #2441

That's a good point, maybe we can remove it.

alexgerardojacinto:
We removed the logic for Android targets below 36.

alexgerardojacinto:
@OS-pedrogustavobilro re-requested your review since I updated the code in https://github.com/ionic-team/capacitor-plugins/pull/2447/commits/fad9bc63f7c7f08ea371ba1b301418ccf50c45f8 to only update foreground color if style is `DEFAULT`, matching iOS.

ItsChaceD (approved):

andredestro (approved):

PR #2468, opened 2026-01-20
fix: AGP 9.0 no longer supporting `proguard-android.txt`

terencehonles:
This change replaces the default proguard file `proguard-android.txt` with `proguard-android-optimize.txt` which allows proguard optimizations.

Alternatively, AGP 9 compatibility can be achieved by opting out of the change with `android.r8.globalOptionsInConsumerRules.disallowed=false`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,6 +111,8 @@ jobs:
         plugin: ${{ fromJson(needs.setup.outputs.plugins) }}
     steps:
       - run: sudo xcode-select --switch ${{ matrix.xcode }}
+      - run: xcrun simctl list > /dev/null
+      - run: xcodebuild -downloadPlatform iOS
       - uses: actions/setup-node@v6
         with:
           node-version: 22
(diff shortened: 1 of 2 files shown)


terencehonles:
Since the core package PR has been accepted and merged I've opened PRs (linked above) for each of the split out repositories to make the same changes. These will all need to be released to unblock AGP 9.

OS-pedrogustavobilro (approved):

jordanorc:
Hi @OS-pedrogustavobilro! Thanks for merging this fix. I’m trying to plan my Android build and I was wondering if there is any expected release version or timeline where this change will be published. Could you please share if there’s a scheduled release including this fix? Thanks!

OS-pedrogustavobilro:
Hey @jordanorc I don't have an exact date to share, I reckon some time this week this fix should be released.

However, [as mentioned in the docs](https://capacitorjs.com/docs/updating/8-0#upgrade-android-studio) (updated after this PR was opened), you don't need to update to AGP 9; Capacitor 8 is built with AGP 8.13, we recommend updating to that specific version that matches Capacitor 8 (unless you specifically need something from AGP 9), otherwise, you could run into other build issues (e.g. with community plugins). And if you use 8.13, this bug/fix shouldn't impact you.

jordanorc:
Hi @OS-pedrogustavobilro, thanks for getting back to me! I ended up missing the note in the documentation about keeping the AGP at version 8.13.0 — my mistake 😅
I’ll test it by keeping the version pinned to 8.13.0. Thanks!

PR #2472, opened 2026-01-29
chore(ci): download iOS simulators

jcesarmobile:
github removed the simulators for Xcode 26.0 too, so we need to revert the PR that removed the workaround

Reverts ionic-team/capacitor-plugins#2466


OS-pedrogustavobilro:
@jcesarmobile the verify-ios didn't run in CI, I because it didn't detect changes in the plugins. Should we manually do some changes just to test and then revert back before merging?

jcesarmobile:
I think there is no need, this is literally a revert of the last commit we merged two weeks ago because "it wasn't needed anymore", but sadly didn't last long

PR #2491, opened 2026-03-11
fix(action-sheet): correct some cancelable misbehaviors

jcesarmobile:
Android was not returning `canceled: false` when not canceled.
Web was not setting a default false value, passing undefined to PWA elements, which caused issues on the first presentation.
iOS was not behaving as documented when there was a Cancel button and it can support the `cancelable` option on iOS 18 and older if there is no Cancel button.

Also, added a few changes in the definition docs.


OS-pedrogustavobilro (approved):
Tested with the dev build on the [testapp PR](https://github.com/ionic-team/capacitor-testapp/pull/776), all's working well on Android iOS and Web.

Thank you for looking into this and making the fixes!

I left a comment, that maybe is more subjective, but I'll approve the PR nonetheless.

PR #2513, opened 2026-04-14
chore: do not publish camera to CocoaPods

OS-pedrogustavobilro:
## Description

Remove the cocoa pods publish command for Camera Plugin.

Because CocoaPods Publish workflow is triggered per plugin directory, removing this command means that if someone tries to push Camera to Trunk from this Repo (needs to explicitly list camera in plugin list input to workflow), the GitHub action will fail.

After this one is approved / merged I will also open a PR for 7.x, for filesystem and geolocation.

## Change Type
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation
- [x] Other

## Rationale / Problems Fixed

Capacitor camera is now housed in https://github.com/ionic-team/capacitor-camera, and `main` / `8.x` in this repo is meant to serve only for historical purposes.

To avoid GitHub Workflow failures and unforeseen issues, we should not try to push camera plugin 8.x to Trunk from this repo.

## Notes / Comments

In terms of avoiding GitHub Workflow failures in Android, refer to https://github.com/ionic-team/capacitor-plugins/pull/2512

### DIFF
--- a/camera/package.json
+++ b/camera/package.json
@@ -43,8 +43,7 @@
     "build": "npm run clean && npm run docgen && tsc && rollup -c rollup.config.mjs",
     "clean": "rimraf ./dist",
     "watch": "tsc --watch",
-    "prepublishOnly": "npm run build",
-    "publish:cocoapod": "echo 'This repository does not publish the camera plugin. Refer to https://github.com/ionic-team/capacitor-camera'"
+    "prepublishOnly": "npm run build"
   },
   "devDependencies": {
     "@capacitor/android": "^8.0.0",
--- a/camera/package.json
+++ b/camera/package.json
@@ -44,7 +44,7 @@
     "clean": "rimraf ./dist",
     "watch": "tsc --watch",
     "prepublishOnly": "npm run build",
-    "publish:cocoapod": "pod trunk push ./CapacitorCamera.podspec --allow-warnings"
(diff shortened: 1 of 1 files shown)


jcesarmobile (commented):
I don't think this is needed since camera was removed from the workspace in the package.json, so since the ci publish runs `lerna run publish:cocoapod`, it won't be run for camera plugin.

If you want to prevent a manual publish, maybe you can remove the whole script rather than changing it to an echo

OS-pedrogustavobilro:
> I don't think this is needed since camera was removed from the workspace in the package.json, so since the ci publish runs lerna run publish:cocoapod, it won't be run for camera plugin.

Don't think that's the case, looking at https://github.com/ionic-team/capacitor-plugins/blob/main/.github/workflows/publish-ios.yml, it does `npm run publish:cocoapod` from each plugin directory (where the list of plugins is an input to the workflow). The root package.json has the lerna command as you state, but it doesn't actually get triggered in the workflow

> If you want to prevent a manual publish, maybe you can remove the whole script rather than changing it to an echo

Yeah, that works too, my idea was to pass silently instead of failing, but I suppose it's better to be an explicit failure, I'll make that change

jcesarmobile (approved):

PR #2523, opened 2026-05-08
chore: update dependencies for Capacitor 9

jcesarmobile:
## Change Type
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Platforms Affected
- [ ] Android
- [ ] iOS
- [ ] Web

### DIFF
--- a/camera/package.json
+++ b/camera/package.json
@@ -46,10 +46,10 @@
     "prepublishOnly": "npm run build"
   },
   "devDependencies": {
-    "@capacitor/android": "^8.0.0",
-    "@capacitor/core": "^8.0.0",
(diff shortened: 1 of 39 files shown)


camera/package.json
OS-pedrogustavobilro, line 1: We can update the Camera here, but it's kind of a no-op, because the Camera Plugin now lives in a separate repo. Last year we removed the `camera` folder when going to the next major. Perhaps we'll want to do that soon as well (in another PR that is).

OS-pedrogustavobilro (approved):

PR #2524, opened 2026-05-08
feat(splash-screen): implement launchFadeOutDuration support on iOS

OS-ruimoreiramendes:
## Description

Implemented `launchFadeOutDuration` support on iOS for cross-platform consistency with Android.

## Change Type
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

`launchFadeOutDuration` was not being read from config on iOS, and the `isLaunchSplash` flag was missing, causing `hide()` to always use the JS call's `fadeOutDuration` instead of the configured value. This PR implements full `launchFadeOutDuration` support on iOS for cross-platform consistency with Android.

Relates to: [RMET-5084](https://outsystemsrd.atlassian.net/browse/RMET-5084)

## Platforms Affected
- [ ] Android
- [x] iOS
- [ ] Web

[RMET-5084]: https://outsystemsrd.atlassian.net/browse/RMET-5084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


splash-screen/ios/Sources/SplashScreenPlugin/SplashScreen.swift
```diff
@@ -32,7 +35,9 @@ import Capacitor
     }
 
     public func hide(settings: SplashScreenSettings) {
-        hideSplash(fadeOutDuration: settings.fadeOutDuration, isLaunchSplash: false)
+        let fadeOutDuration = self.isLaunchSplash ? config.launchFadeOutDuration : settings.fadeOutDuration
```
OS-pedrogustavobilro, line 38: Can you explain the rationale behind this logic? Didn't seem like the equivalent existed on Android but maybe I didn't see right.

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreenConfig.java
OS-pedrogustavobilro, line 1: With which app did you test btw, capacitor-testapp?

splash-screen/ios/Sources/SplashScreenPlugin/SplashScreen.swift
OS-pedrogustavobilro, line 1: In my eyes it would make more sense to separate the iOS implementation in a different PR, particularly for release not generation. But I suppose it's not that big of a deal.

splash-screen/android/src/main/java/com/capacitorjs/plugins/splashscreen/SplashScreenConfig.java
OS-ruimoreiramendes: yup, with capacitor-testapp, by adding `StatusBar.setStyle({ style: Style.Light });` right after `SplashScreen.hide();` in the App.tsx file

splash-screen/ios/Sources/SplashScreenPlugin/SplashScreen.swift
OS-ruimoreiramendes: On Android (using the Android 12 API), `launchFadeOutDuration` is applied inside `OnExitAnimationListener`, which is a system callback that runs before `hide()` is ever called from JS, so the system already knows it's closing the launch splash and applies the right duration at that point.
On iOS there's no equivalent system hook. The only way to dismiss the splash (including the launch one) is through `hideSplash()`, which is triggered by `hide()`. The problem is that `hide()` receives a `SplashScreenSettings` object coming from the js call, and that object has no context about whether it's hiding the launch splash or a programmatically shown one. Without the `isLaunchSplash` flag, `hide()` would always use `settings.fadeOutDuration` from the js parameters (default 200ms), ignoring the `launchFadeOutDuration` set in `capacitor.config.ts`.
So the flag is needed to let `hide()` decide:

"am I closing the launch splash, so use `config.launchFadeOutDuration`" 
vs 
"am I closing a programmatic splash, so use `settings.fadeOutDuration` from the js call".

splash-screen/ios/Sources/SplashScreenPlugin/SplashScreen.swift
OS-ruimoreiramendes: ok, I can split in two different prs

OS-ruimoreiramendes: I've removed the Android changes from this PR and created a separate one for the fix https://github.com/ionic-team/capacitor-plugins/pull/2526

splash-screen/ios/Sources/SplashScreenPlugin/SplashScreen.swift
OS-pedrogustavobilro: Ok makes sense, thanks for the detailed explanation!

OS-pedrogustavobilro (approved):
Tested with `capacitor-testapp`, changing values of `launchFadeOutDuration` in capacitor config, and also adding `fadeOutDuration` to `show` or `hide`. All seems to be working on different iOS versions (tested 15, 18 and 26)

PR #2529, opened 2026-05-14
feat(push-notifications): add banner and list presentation options for iOS

OS-ruimoreiramendes:
## Description

Added `.banner` and `.list` as new `presentationOptions` values for iOS, replacing the deprecated `.alert` presentation option. For backwards compatibility, `.alert` is kept but now maps to both `.banner` and `.list` on iOS. 
On Android, `.banner` and `.list` trigger the same behavior as `.alert`.

Reference: https://outsystemsrd.atlassian.net/browse/RMET-5102

## Change Type
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

`.alert` in `UNNotificationPresentationOptions` is deprecated since iOS 14 and replaced by the more granular `.banner` and `.list`. 
This PR adds support for the new values while keeping `.alert` working for backwards compatibility.

## Tests or Reproductions

Tested on a physical iOS device using the Xcode Push Notifications Console with the following `presentationOptions` combinations:

- `["banner", "list", "sound"]`: banner appears + sound plays + notification in notification center
- `["banner", "sound"]`: banner appears + sound plays
- `["list", "sound"]`: no banner + sound plays + notification in notification center
- `["alert", "sound"]`: backwards compatibility behaves like `["banner", "list", "sound"]`

## Platforms Affected
- [x] Android
- [x] iOS
- [ ] Web

### DIFF
--- a/push-notifications/src/definitions.ts
+++ b/push-notifications/src/definitions.ts
@@ -15,8 +15,8 @@ declare module '@capacitor/cli' {
        *   - `badge`: badge count on the app icon is updated (default value)
        *   - `sound`: the device will ring/vibrate when the push notification is received
        *   - `alert`: **Deprecated on iOS.** Use `banner` and `list` instead. On Android, this value is still used to display the notification.
-       *   - `banner`: the push notification is displayed as a banner
-       *   - `list`: the push notification is displayed in the notification center
+       *   - `banner`: the push notification is displayed as a banner. On Android, defaults to the same behavior as `alert`.
+       *   - `list`: the push notification is displayed in the notification center. On Android, defaults to the same behavior as `alert`.
        *
        * An empty array can be provided if none of the options are desired.
        *
--- a/push-notifications/src/definitions.ts
+++ b/push-notifications/src/definitions.ts
@@ -2,7 +2,7 @@
 
 import type { PermissionState, PluginListenerHandle } from '@capacitor/core';
 
-export type PresentationOption = 'badge' | 'sound' | 'alert';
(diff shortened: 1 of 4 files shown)


push-notifications/src/definitions.ts
```diff
@@ -14,14 +14,16 @@ declare module '@capacitor/cli' {
        * This is an array of strings you can combine. Possible values in the array are:
        *   - `badge`: badge count on the app icon is updated (default value)
        *   - `sound`: the device will ring/vibrate when the push notification is received
-       *   - `alert`: the push notification is displayed in a native dialog
+       *   - `alert`: **Deprecated on iOS.** Use `banner` and `list` instead. On Android, this value is still used to display the notification.
+       *   - `banner`: the push notification is displayed as a banner
+       *   - `list`: the push notification is displayed in the notification center
```
OS-pedrogustavobilro: Worth stating that banner and list default to alert on Android?

OS-pedrogustavobilro (commented):
One minor comment

push-notifications/src/definitions.ts
OS-ruimoreiramendes: Yeah, makes sense, clarifying that!

OS-ruimoreiramendes: I've updated the doc for banner and list to clarify that on Android they default to the same behavior as alert

OS-pedrogustavobilro (approved):

PR #2542, opened 2026-06-08
feat(app): adding edge gesture handlers

theproducer:
## Description
Adds a new `edgeGesture` event that adds support for system edge-swipe gestures on both iOS and Android.

## Change Type
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Platforms Affected
- [x] Android
- [x] iOS
- [ ] Web


app/ios/Sources/AppPlugin/AppPlugin.swift
```diff
@@ -125,4 +134,104 @@ public class AppPlugin: CAPPlugin, CAPBridgedPlugin {
...
+        }
+    }
+
+    @objc func handleEdgePan(_ recognizer: UIScreenEdgePanGestureRecognizer) {
+        guard let view = bridge?.webView else { return }
+
+        let translation = recognizer.translation(in: view)
+        let touch = recognizer.location(in: view)
+        let viewWidth = view.bounds.width
+        let viewHeight = view.bounds.height
```
OS-ruimoreiramendes: `viewHeight` is declared but never used, can be removed.

app/android/src/main/java/com/capacitorjs/plugins/app/AppPlugin.java
```diff
@@ -21,33 +30,40 @@
...
-            });
-        bridge
-            .getApp()
-            .setAppRestoredListener((result) -> {
-                Logger.debug(getLogTag(), "Firing restored result");
-                notifyListeners(EVENT_RESTORED_RESULT, result.getWrappedResult(), true);
-            });
-        this.onBackPressedCallback = new OnBackPressedCallback(!disableBackButtonHandler) {
+        this.backButtonHandlerEnabled = !getConfig().getBoolean("disableBackButtonHandler", false);
+        this.edgeGestureHandlerEnabled = getConfig().getBoolean("enableEdgeGestureHandler", false);
```
OS-ruimoreiramendes: I think there might be an issue here worth investigating.
On Android < 14, if the developer opts into `enableEdgeGestureHandler`, this assignment makes `edgeGestureHandlerEnabled = true`. `applyBackButtonHandlerState()` then disables the backButton handler (via `backButtonHandlerEnabled && !edgeGestureHandlerEnabled`), but `setupBackGestureHandlers()` no-ops because `OnBackAnimationCallback` requires API 34+.

Net effect for users on Android < 14,  backButton stops dispatching.

Suggest gating the flag by SDK:
```
boolean canUseEdgeGesture = Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE;
this.edgeGestureHandlerEnabled = canUseEdgeGesture && getConfig().getBoolean("enableEdgeGestureHandler", false);
```
Same fix needed in `toggleEdgeGestureHandler` .
Could you double-check this scenario? Might be that I'm missing something.

app/android/src/main/java/com/capacitorjs/plugins/app/AppPlugin.java
```diff
@@ -116,26 +138,51 @@ public void minimizeApp(PluginCall call) {
...
+        } else {
+            teardownBackGestureHandlers();
+        }
+
+        call.resolve();
+    }
+
+    private void applyBackButtonHandlerState() {
+        if (this.onBackPressedCallback != null) {
+            this.onBackPressedCallback.setEnabled(backButtonHandlerEnabled && !edgeGestureHandlerEnabled);
```
OS-ruimoreiramendes, line 186: Tied to the earlier comment on line 53: on Android < 14, if the developer opts into `enableEdgeGestureHandler`, this line evaluates `setEnabled(true && !true) = setEnabled(false)` and disables the backButton handler. But `setupBackGestureHandlers()` no-ops on those versions because `OnBackAnimationCallback` requires API 34+. backButton stops dispatching.
Silent regression with no error. Gating the flag at assignment (see line 53) would prevent this.

OS-ruimoreiramendes (commented):
Besides the inline comments I left, I tested the changes locally on a test app and everything seems to be working correctly. Nice work overall.

app/android/src/main/java/com/capacitorjs/plugins/app/AppPlugin.java
theproducer: I can confirm this as well in testing on Android 14.  Good catch!

OS-ruimoreiramendes (approved):

PR #2550, opened 2026-06-18
chore: bump version to 9.0.0-alpha.0

jcesarmobile:


OS-pedrogustavobilro (approved):
Not necessarily to be done in this PR, but I think for the prereleases to work from `next` branch, I think we need to update [lerna.json](https://github.com/ionic-team/capacitor-plugins/blob/main/lerna.json#L8) which currently has `"allowBranch": "main"`

PR #2445, opened 2025-11-13
fix(android): notification Bundles with non-string values

frederikbosch:
The method `getDeliveredNotifications()` throws because `getString(key)` is called on `notification.extras` but the `extras` [`Bundle`](https://developer.android.com/reference/android/os/Bundle) object contains more than just strings. Booleans cannot be casted to string, see the exception stack trace below.

```log
Key android.progressIndeterminate expected String but value was a java.lang.Boolean.  The default value <null> was returned. 
Attempt to cast generated internal exception:
    java.lang.ClassCastException: java.lang.Boolean cannot be cast to java.lang.String
        at android.os.BaseBundle.getString(BaseBundle.java:1448)
        at com.capacitorjs.plugins.pushnotifications.PushNotificationsPlugin.getDeliveredNotifications(PushNotificationsPlugin.java:147)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.getcapacitor.PluginHandle.invoke(PluginHandle.java:138)
        at com.getcapacitor.Bridge.lambda$callPluginMethod$0(Bridge.java:837)
        at com.getcapacitor.Bridge.$r8$lambda$FNYM7cvgeBk0k8YXQH7M96Mrf-c(Unknown Source:0)
        at com.getcapacitor.Bridge$$ExternalSyntheticLambda0.run(D8$$SyntheticClass:0)
        at android.os.Handler.handleCallback(Handler.java:995)
        at android.os.Handler.dispatchMessage(Handler.java:103)
        at android.os.Looper.loopOnce(Looper.java:273)
        at android.os.Looper.loop(Looper.java:363)
        at android.os.HandlerThread.run(HandlerThread.java:85)
```

PR #2552, opened 2026-06-18
refactor(android): suppress Bundle.get(String) deprecation warning

OS-ruimoreiramendes:
## Description

Replace the deprecated `Bundle.get(String)` calls in `push-notifications` and `local-notifications` with a small helper that uses typed Bundle getters and `BundleCompat.getParcelable` / `getSerializable` as type-safe fallbacks.

Ref: https://outsystemsrd.atlassian.net/browse/RMET-5239

## Change Type
- [ ] Fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

`BaseBundle.get(String)` has been deprecated since Android API 33. It was reintroduced in `push-notifications` and `local-notifications` by #2445  to allow non-`String` values in notification bundles, but Android does not provide a single type-safe replacement for "iterate over keys of unknown type", the recommended migration is to use type-specific accessors.

**Tooltip from Android Studio**
"This method was deprecated in API level 33. Use the type-safe specific APIs depending on the type of the item to be retrieved, eg. getString(String)."

## Tests or Reproductions

Manually validated on a physical Android using `capacitor-testapp`.
Across all keys (extras such as `android.title`, `android.text`, `android.progress`, etc), the final JSON produced by the new helper matched the JSON produced by the deprecated `bundle.get(key)` path.

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web


local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationsPlugin.java
```diff
@@ -287,6 +292,42 @@ public static void fireReceived(JSObject notification) {
         }
     }
 
+    private static Object extractBundleValue(Bundle bundle, String key) {
```
OS-pedrogustavobilro: Are we certain this covers all types that can come in `Bundle`. I think you covered the ones in `BaseBundle`, but in theory there could be others in `bundle` e.g. [array of ints](https://developer.android.com/reference/android/os/Bundle#getIntegerArrayList(java.lang.String)).

For local notifications `extra` is type `any` in Typescript, for push notifications it comes from firebase, and we cannot control either of those.

It feels like for this kind of situations (payloads in bundle that we cannot control), the deprecated `BaseBundle#get` is what actually does the trick.

Or if not that, we need to explicitly try all methods in both `Bundle` and `BaseBundle`?

Maybe I'm thinking of really low-likelihood edge case scenarios, and if so, this is fine.

Let me know your thoughts!

OS-ruimoreiramendes: > Are we certain this covers all types that can come in Bundle. I think you covered the ones in BaseBundle, but in theory there could be others in bundle e.g. array of ints.

Arrays are handled implicitly by the Serializable fallback at the bottom of the helper. All Java arrays (`int[]`, `boolean[]`, `Parcelable[]`, etc.) and `ArrayList<*>` implement Serializable, so `BundleCompat.getSerializable(bundle, key, Serializable.class)` catches them.

> For local notifications extra is type any in Typescript, for push notifications it comes from firebase, and we cannot control either of those.

the helper makes no assumption about the payload shape, it tries typed getters in order of likelihood and falls through to Parcelable/Serializable for anything else.

> It feels like for this kind of situations (payloads in bundle that we cannot control), the deprecated BaseBundle#get is what actually does the trick.

That's exactly what `BundleCompat.getParcelable/getSerializable` is, the end result is functionally the same as `bundle.get(key)` for any Parcelable/Serializable value.

> Or if not that, we need to explicitly try all methods in both Bundle and BaseBundle?

We could go further and emit proper `JSONArrays` via `getIntArray`/`getLongArray`/`getBooleanArray`/etc. instead of relying on the `Serializable.toString()` fallback for primitive arrays.

> Maybe I'm thinking of really low-likelihood edge case scenarios, and if so, this is fine.

I tested both plugins on a real device, comparing the new helper's output side-by-side with the deprecated `bundle.get(key)` for every key and they returned the same values.
The only divergence I found initially was on Parcelable values (e.g. `android.appInfo` > `ApplicationInfo`), the helper returned null while the deprecated path produced the stringified object via `Object.toString`. I addressed that by adding the `BundleCompat.getParcelable/getSerializable` fallbacks, with those in place, both paths now return matching values across all keys I observed.

Let me know whether you think the current approach is sufficient, or if you'd prefer me to add the explicit array handling on top.

OS-pedrogustavobilro: Alrighty, makes sense yeah, these do seem to cover enough scenarios, and if from your tests it covers int arrays and stuff like that, we should be good!

OS-pedrogustavobilro (approved):

local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationsPlugin.java
OS-ruimoreiramendes: Yeah, I made sure to test both implementations side by side, with both empty and custom data payloads (strings, numbers/booleans sent as strings, URLs, emoji + escaped quotes, plus the nested `gcm.n.analytics_data` Bundle). The only divergence I caught was the appInfo Parcelable I mentioned earlier, which is what led me to add the BundleCompat fallbacks.

jcesarmobile (commented):
why is the extractBundleValue method static? I don't think it needs to be.

OS-ruimoreiramendes:
> why is the extractBundleValue method static? I don't think it needs to be.

yeah, no reason for it to be static. I likely added it during an earlier iteration and forgot to clean up

jcesarmobile:
I've been testing and it generates a ton of `java.lang.ClassCastException` warnings (one per key and per type check), while they are not crashes, I think they can scare/confuse users.

Maybe we should keep get despite it's deprecated and just silence it? I doubt google will remove it since there is no real alternative.

OS-ruimoreiramendes:
> I've been testing and it generates a ton of `java.lang.ClassCastException` warnings (one per key and per type check), while they are not crashes, I think they can scare/confuse users.
> 
> Maybe we should keep get despite it's deprecated and just silence it? I doubt google will remove it since there is no real alternative.

@jcesarmobile Hmm, I missed the warnings because my logcat filter was silencing everything outside my custom tag.
I'll simplify to a single bundle.get(key) and suppress the warning

jcesarmobile (approved):

PR #2555, opened 2026-06-24
feat(network): expose connection condition flags

birkskyum:
## Description

Adds optional `constrained` and `expensive` fields to `Network.getStatus()` and `networkStatusChange`.

This keeps `connectionType` unchanged and exposes connection condition signals separately:

- iOS uses `NWPath.isConstrained` and `NWPath.isExpensive`
- Android uses metered network capabilities, Data Saver restrictions, and Android 15 bandwidth-constrained capabilities
- Web uses `navigator.connection.saveData` for `constrained`

Offline status reports both fields as `false`.

## Change Type

- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

Some apps need to adjust network behavior based on more than connection type. For example, a connection can still be Wi-Fi or cellular while also being metered, expensive, or under a platform data-saving mode.

This adds platform-backed condition flags without changing the existing `connectionType` API.

## Tests or Reproductions

Tested with:

- `CAP_PLUGIN_PUBLISH=true CAPACITOR_VERSION=8.4.1 npm run verify`
- `npm run lint`
- `xcodebuild test -scheme CapacitorNetwork -destination 'platform=iOS Simulator,name=iPhone 17,OS=26.5'`
- `xcodebuild build -scheme CapacitorNetwork -destination 'id=00008150-001155E01191401C'`

Also verified the iOS target builds for a physical iPhone 17 Pro on iOS 26.5.1.

## Screenshots / Media

N/A

## Platforms Affected

- [x] Android
- [x] iOS
- [x] Web

## Notes / Comments

No related issue that I could find.

### DIFF
--- a/network/README.md
+++ b/network/README.md
@@ -99,6 +99,8 @@ Represents the state and type of the network connection.
 | -------------------- | --------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------- | ----- |
 | **`connected`**      | <code>boolean</code>                                      | Whether there is an active connection or not.                                                                                 | 1.0.0 |
 | **`connectionType`** | <code><a href="#connectiontype">ConnectionType</a></code> | The type of network connection currently in use. If there is no active network connection, `connectionType` will be `'none'`. | 1.0.0 |
+| **`constrained`**    | <code>boolean</code>                                      | Whether the active connection is constrained by platform data-saving or bandwidth-reduction signals.                          | 8.1.0 |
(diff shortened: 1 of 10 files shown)


ingalls:
@jcesarmobile Would you be able to review these changes or point us in the direction of someone who could? This feature would be incredibly useful for our app which tracks first responders engaged in wildland fire response when on a satellite connection vs LTE.

PR #2556, opened 2026-06-25
chore: Align plugin baselines with core

jcesarmobile:
update android:
- gradle wrapper to 9.5.1
- gradle plugin to 9.2.1
- compileSdk to 37
- minSdkVersion to 26

remove targetSdkVersion

Update iOS:
- s.ios.deployment_target to 16
- platforms: .iOS to 16


markemer (approved):
Now that we're doing this, we should probably start setting up the next branch to publish 9 alphas

local-notifications/android/build.gradle
theproducer, line 1: @jcesarmobile I think this plugin is missing the class path version upgrades?

theproducer (changes requested):

local-notifications/android/build.gradle
jcesarmobile: good catch

theproducer (approved):

PR #2558, opened 2026-06-29
docs(motion): add orientation listener example to README

### LINKED ISSUE #2539: Orientation document of capacitor/motion is wrong.
### Capacitor Version

```
PASTE OUTPUT HERE
```

### Current Behavior

it describes only the accel event listener and not the orientation.

### Expected Behavior

should describe both

OS-ruimoreiramendes:
README

## Description

Add an 'orientation' listener example to the Motion plugin README alongside the existing 'accel' one.

Closes #2539.

## Change Type
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [x] Documentation

## Notes / Comments

 Plan to backport to 7.x.

### DIFF
--- a/motion/README.md
+++ b/motion/README.md
@@ -23,10 +23,9 @@ import { Motion } from '@capacitor/motion';
 let accelHandler: PluginListenerHandle;
 let orientationHandler: PluginListenerHandle;
 
-myButton.addEventListener('click', async () => {
+myAccelerationButton.addEventListener('click', async () => {
   try {
     await DeviceMotionEvent.requestPermission();
-    await DeviceOrientationEvent.requestPermission();
   } catch (e) {
     // Handle error
     return;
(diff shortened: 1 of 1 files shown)


markemer (approved):

jcesarmobile (commented):
can this be separate examples? 
users might want to listen for the events separately, should look more to what we do in the example app
https://github.com/ionic-team/capacitor-testapp/blob/main/src/pages/Motion.tsx

OS-ruimoreiramendes:
> can this be separate examples? users might want to listen for the events separately, should look more to what we do in the example app https://github.com/ionic-team/capacitor-testapp/blob/main/src/pages/Motion.tsx

@jcesarmobile  I've split the doc into two separate examples. Take another look when you can.

jcesarmobile:
The text says "most browsers require permission check", but that's not true, the permissions are iOS only, the current examples result in the code not working on Android/Web, it should check if the requestPermission method is available before trying to call it, check MDN examples.


BTW, Capacitor 7 is out of support, so there is no need to cherry-pick anymore.
But there are unreleased changes on 7.x branch, so if you do, we can do a final release and release those changes too.

OS-ruimoreiramendes:
> BTW, Capacitor 7 is out of support, so there is no need to cherry-pick anymore.
> But there are unreleased changes on 7.x branch, so if you do, we can do a final release and release those changes too.

@jcesarmobile Updated the intro to make the permission check iOS-only, and guarded both `requestPermission()` calls with `typeof ... === 'function'` following the MDN pattern. Also added the `!== 'granted'` check so we don't try to subscribe when the user denies.

Regarding the 7.x backport, I'll go ahead and cherry-pick this once merged so we can bundle it with the other pending 7.x changes for the final release.

motion/README.md
```diff
@@ -11,23 +11,28 @@ npx cap sync
 
 ## Permissions
 
-This plugin is currently implemented using Web APIs. Most browsers require
-permission before using this API. To request permission, prompt the user for
-permission on any user-initiated action (such as a button click):
+This plugin is currently implemented using Web APIs. On iOS, permission must
+be requested before accessing device motion or orientation events. Other
+platforms do not require the check. To request permission, prompt the user
```
jcesarmobile: It's iOS devices, including web platform when testing on iOS devices, not just iOS platform, so "Other platforms" is not correct.

OS-ruimoreiramendes: Take another look please

jcesarmobile (approved):

PR #2560, opened 2026-07-03
chore: add dist to root gitignore

OS-pedrogustavobilro:
Plugins already have `dist` inside their own .gitignore, this is to prevent `dist` files appearing untracked after switching to a branch that doesn't have a certain plugin. A current example:

1. Have the camera plugin in `main` branch with build `dist` folder.
2. Switch to `next` branch where camera was removed
3. Without this PR, you'll find `camera` folder in untracked files, because of the `dist` folder that wasn't in source control therefore it was never technically removed.

With `dist` in root .gitignore we won't accidentally commit those files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
+dist
 .DS_Store
 package-lock.json
 Podfile.lock


jcesarmobile (approved):
`npm run build` in the root no longer builds camera plugin, you have to explicitly build the plugin from the folder, which nobody should be doing anymore since it was moved to another repository, but anyway, we can merge this

OS-pedrogustavobilro:
> npm run build in the root no longer builds camera plugin, you have to explicitly build the plugin from the folder, which nobody should be doing anymore since it was moved to another repository, but anyway, we can merge this

The reason why this happens for me then is because I have built it before that was disabled. Which may happen again in the future if we start moving other plugins out of this repository.

PR #2561, opened 2026-07-03
fix(browser): only fire browserFinished when the Custom Tab actually terminates

### LINKED ISSUE #2554: [Bug]: `browserFinished` on Android triggers when browser is minimized
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 8.4.0
  @capacitor/core: 8.4.0
  @capacitor/android: 8.4.0
  @capacitor/ios: 8.4.0

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/core: 8.4.0
  @capacitor/android: 8.4.0
  @capacitor/cli: 8.4.0

### Other API Details

```Shell

```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

For repro I used `npm init @capacitor/app@latest` and then replaced index.html body with

```html
<button>Open browser</button>
<pre></pre>
```

and `capacitor-welcome.js` with

```js
import { Browser } from '@capacitor/browser';

Browser.addListener('browserFinished', async () => {
  document.querySelector('pre').innerText += 'browserFinished\n';
});

document.querySelector('button').addEventListener('click', () => {
  Browser.open({
    url: 'https://example.com',
  });
});
```

Steps:

1. Click "open browser"
2. Minimize the browser
3. `browserFinished` has fired
4. It won't fire again when maximizing/minimizing or closing the browser

https://github.com/user-attachments/assets/1d781e02-7387-45f3-8b36-568a0d7b8dab

### Expected Behavior

I expect `browserFinished` to tell me when the browser was _closed_

I open Auth0 in a browser (because of https://github.com/ionic-team/capacitor/issues/7367). But I can't reliably know when the browser is done to remove the "loading" state (spinner) in my app.

### Project Reproduction

https://github.com/Prinzhorn/capacitor-android-browser-close-event

### Additional Information

_No response_

OS-ruimoreiramendes:
## Description

Refactor the Android browser plugin to use `ActivityResultLauncher` for detecting when the Custom Tab has actually terminated, instead of inferring closure from lifecycle events via the previous `EventGroup` counter.
The `EventGroup` helper is removed. Browser now exposes `setCustomTabLauncher` and `BrowserControllerActivity` registers a launcher via `registerForActivityResult`, the `browserFinished` event is only fired when the launched Custom Tab activity ends.

Fixes #2554.

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

When the user minimizes the Chrome Custom Tab to Picture-in-Picture (or backgrounds the app), the previous heuristic counted TAB_HIDDEN + unbindService and inferred the tab had closed firing `browserFinished` even though the tab was still open. After that, subsequent minimize/maximize/close events wouldn't fire the event again.

`ActivityResultLauncher` provides a single, deterministic callback only when the Custom Tab activity
truly finishes, so `browserFinished` now reflects the actual closed state.

## Tests or Reproductions

Reproduced with the repo:
https://github.com/Prinzhorn/capacitor-android-browser-close-event

Manually tested on a physical Android device:
  - Open browser → close via back gesture / X button → browserFinished fires once
  - Open browser → minimize to PiP → maximize → close → browserFinished fires once, only on close
  - Open browser → send app to background → return to app → close browser → browserFinished fires once,
  only on close

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web


browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserControllerActivity.java
```diff
@@ -1,19 +1,28 @@
...
+import androidx.activity.result.ActivityResultLauncher;
+import androidx.activity.result.contract.ActivityResultContracts;
 import androidx.annotation.Nullable;
 
-public class BrowserControllerActivity extends Activity {
+public class BrowserControllerActivity extends ComponentActivity {
 
-    private boolean isCustomTabsOpen = false;
+    private ActivityResultLauncher<Intent> customTabLauncher;
+    private Browser implementation;
```
OS-pedrogustavobilro, line 14: Perhaps we should set `implementation` in `onDestroy` just to clear the reference?

browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserControllerActivity.java
```diff
@@ -28,25 +37,15 @@ protected void onNewIntent(Intent intent) {
         }
     }
```
OS-pedrogustavobilro, line 38: Unsure if we should add the `notifyBrowserFinished` here as well.

I think beforehand the intent wouldn't fire an event back to the plugin in this situation so maybe not relevant?

OS-pedrogustavobilro (commented):
A couple comments, nothing major. Haven't tested yet but overall looks good, noice!

browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserControllerActivity.java
OS-ruimoreiramendes: Good point. Clearing it makes sure we always get a fresh reference on the next open.

browser/android/src/main/java/com/capacitorjs/plugins/browser/BrowserControllerActivity.java
OS-ruimoreiramendes: No need, the event already fires here via the launcher callback.
When `finish()` runs, it also closes the Custom Tab, and that triggers the launcher's result callback, which calls `notifyBrowserFinished()`. 
Tested it, and adding it here on top just causes a duplicate.

Also confirmed on `main` that the event already fired in this case before the PR, so behavior is preserved.

browser/android/src/main/java/com/capacitorjs/plugins/browser/Browser.java
OS-pedrogustavobilro, line 1: I am getting an issue here on Android 12 and 13 (didn't test below 12 but might happen there too). Doesn't seem to happen on 14 or higher.

If I minimize the custom tab, then close it, I don't get the event that it closed, only when backgrounding the app and re-opening it. Can you test and check on your end, and see if the code maybe needs to be ammended?

Attaching video of what I mean (Nokia with Android 12; got it as well on some Samsungs and a Pixel): 

https://github.com/user-attachments/assets/574e0169-cc93-49b1-b830-5bf29a5c81de

Thanks!

OS-ruimoreiramendes: Thanks for flagging! I'll take a look and try to reproduce on my end.

OS-pedrogustavobilro (approved):
Tested on devices with different Android version, looking good. After this is merged we should mention in the original issue that this fix could only be done for Android 14 and higher - below Android 13 similar behavior as before. Sound good?

OS-ruimoreiramendes:
> Tested on devices with different Android version, looking good. After this is merged we should mention in the original issue that this fix could only be done for Android 14 and higher - below Android 13 similar behavior as before. Sound good?

yup, sounds good

PR #2564, opened 2026-07-14
feat(network): Add internet reachability support

sc-nick:
## Description
Extends `@capacitor/network` with actual internet reachability information while preserving the existing connection detection behavior and API structure.

This plugin now reports whether the internet is actually reachable, allowing applications to distinguish between being connected to a network and having a real internet access.

### Changes
- Adds `internetReachable` to indicate whether internet access is available
- Adds `state` with the following values:
  - `offline`: no network connection
  - `online`: connected with internet access
  - `limited`: connected to a network without internet access
- Keeps the existing `connected` and `connectionType` fields unchanged
- Keeps the existing `getStatus()` method and `networkStatusChange` event

### Platform implementation
- **iOS:** keeps the existing `Reachability` connection monitoring and adds HTTP probs to verify internet access
- **Android:** uses `NET_CAPABILITY_VALIDATED` to determine whether the active network has internet access

### Example payload
`{
      "connected": true,
      "connectionType": "wifi",
      "internetReachable": false,
      "state": "limited" }`

## Change Type
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed
The `@capacitor/network` plugin currently doesn't report whether internet access is actually available. As a result application using the plugin may incorrectly assume that online operations can succeed whenever a network connection is available.

## Platforms Affected
- [x] Android
- [x] iOS
- [ ] Web

### DIFF
--- a/network/android/src/main/java/com/capacitorjs/plugins/network/Network.java
+++ b/network/android/src/main/java/com/capacitorjs/plugins/network/Network.java
@@ -77,9 +77,10 @@ public class Network {
             android.net.Network activeNetwork = this.connectivityManager.getActiveNetwork();
             NetworkCapabilities capabilities = this.connectivityManager.getNetworkCapabilities(this.connectivityManager.getActiveNetwork());
             if (activeNetwork != null && capabilities != null) {
-                networkStatus.connected =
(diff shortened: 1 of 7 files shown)


sc-nick:
@alexgerardojacinto

Issue #1384, opened 2023-01-22
LocalNotification set autoCancel false is not working


ysneo:
### Plugin(s)

 "@capacitor/local-notifications": "^4.1.4",

### Capacitor Version

```
v4
```

### Current Behavior

LocalNotification set autoCancel false is not working
      autoCancel: false,

### Expected Behavior

**autoCancel set to false**

### Code Reproduction

  autoCancel: false,  this property is not woring; when click notification, still disappear

### Other Technical Details

no

### Additional Context

no

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

Please see the Contributing Guide for [how to create a Code Reproduction](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
Ionitron 💙

Ionitron:
It looks like this issue didn't get the information it needed, so I'll close it for now. If I made a mistake, sorry! I am just a bot.

Have a great day!
Ionitron 💙

PR #2568, opened 2026-07-20
fix(local-notifications): honor autoCancel when a notification is tapped

begezavr2:
Closes https://github.com/ionic-team/capacitor-local-notifications/issues/10

## What

On Android, `handleNotificationActionPerformed` dismissed the tapped notification unconditionally via `dismissVisibleNotification(notificationId)`, which made `autoCancel: false` a no-op on the tap path — the notification disappeared from the shade on every tap regardless of the flag (the builder's `setAutoCancel(false)` never got a chance to matter, since the plugin issues its own explicit `NotificationManager.cancel()`).

## How

Read `autoCancel` from the notification payload the tap intent already carries (`NOTIFICATION_OBJ_INTENT_KEY`) and only call `dismissVisibleNotification` when it is `true`. The default stays `true` (also when the payload is absent or unparseable), so behavior is unchanged for every caller that didn't explicitly set `autoCancel: false`.

## Testing

Manually on a physical device (Android 14, Capacitor 8.4.2):

- `schedule({ notifications: [{ id, title, body, autoCancel: false }] })` → tap → app opens, notification stays in the shade.
- Same without `autoCancel` and with `autoCancel: true` → tap → notification is dismissed (unchanged behavior).
- Re-scheduling with the same `id` still replaces the delivered notification in place.

## Context

Fixes the behavior reported in ionic-team/capacitor-plugins#1384 (closed by the bot for lack of reproduction, locked; the code path is unchanged since then).

### DIFF
--- a/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationManager.java
+++ b/local-notifications/android/src/main/java/com/capacitorjs/plugins/localnotifications/LocalNotificationManager.java
@@ -83,8 +83,6 @@ public class LocalNotificationManager {
         }
         String menuAction = data.getStringExtra(LocalNotificationManager.ACTION_INTENT_KEY);
 
-        dismissVisibleNotification(notificationId);
-
         dataJson.put("actionId", menuAction);
         JSONObject request = null;
         try {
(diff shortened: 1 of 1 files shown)


OS-pedrogustavobilro:
Hey @begezavr2, thanks for the PR!

Some time after this PR was opened, the local-notifications Capacitor Plugin has since been moved to its own GitHub repository https://github.com/ionic-team/capacitor-local-notifications. Therefore we are no longer accepting PRs in this repo for that plugin, and will be closing PR.

We ask you to:

1. Check if the issue you are trying to fix has been addressed in the latest plugin version. If so, https://github.com/ionic-team/capacitor-local-notifications/issues/10 can be closed.
2. If it hasn't been fixed, can you open a similar PR to that repository instead? Keep in mind the new repository uses Kotlin for Android code, so you'll need to do some tweaks; unfortunately won't be a straight copy-paste from this PR.

We appreciate your contribution, and apologize for any inconvenience caused.

PR #2572, opened 2026-07-28
fix(push-notifications): show foreground notification image on Android

### LINKED ISSUE #1830: Issue with images and icons not being displayed in push messages received while in the foreground on Android devices.
### Plugin(s)
capacitor/push-notification

### Capacitor Version
    "@capacitor-community/media": "^5.0.0",
    "@capacitor/android": "5.0.3",
    "@capacitor/app": "5.0.2",
    "@capacitor/app-launcher": "^5.0.2",
    "@capacitor/clipboard": "^5.0.2",
    "@capacitor/core": "^5.0.4",
    "@capacitor/device": "^5.0.6",
    "@capacitor/filesystem": "^5.0.2",
    "@capacitor/haptics": "5.0.2",
    "@capacitor/ios": "^5.0.3",
    "@capacitor/keyboard": "^5.0.6",
    "@capacitor/local-notifications": "^5.0.6",
    "@capacitor/network": "^5.0.2",
    **"@capacitor/push-notifications": "^5.0.6",**
    "@capacitor/splash-screen": "^5.0.2",
    "@capacitor/status-bar": "5.0.2",

### Platform(s)
android

### Current Behavior
Push messages received in the foreground on Android devices do not display images and icons. However, push messages received in the background on Android devices display images and icons correctly. Please provide a solution to this issue.

### Expected Behavior
Icons and images should also be properly displayed in push notifications received while in the foreground.

Swiezu:
## Description

When a notification message carrying an image arrives while the app is in the foreground on Android, the plugin now downloads the image and displays it as a BigPictureStyle (also set as the large icon), matching how the same push renders when it arrives in the background. The download runs off the main thread with a 5s timeout; on failure or timeout the notification is still posted without the image.

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

Fixes #1830. On Android, foreground notifications are built and posted by the plugin itself via `CommonNotificationBuilder.createNotificationInfo(...)` + `notificationManager.notify(...)`. That path never reads `notification.getImageUrl()`, so images and icons that show on background notifications are missing when the same push arrives in the foreground. Background notifications work only because FCM's own service downloads the image before posting.

This change replicates that download for the foreground path, reusing the same internal FCM helper (`com.google.firebase.messaging.ImageDownload`) the code already depends on (it already uses `CommonNotificationBuilder` / `NotificationParams`), so no new dependency is introduced.

## Tests or Reproductions

To reproduce the bug: send an FCM notification containing an `image` to a device with the app in the foreground — the image is missing, while the same push shows it when the app is backgrounded.

Running in a production Capacitor 8 app: with this change a foregrounded push with an image now renders the picture identically to the background case. Verified the download runs off the main thread and falls back to a text-only notification on failure/timeout.

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web

## Notes / Comments

Foreground notifications are posted by the plugin, background ones by FCM — this aligns the two so they look the same. iOS is unaffected (rich media there is handled by the notification service extension).


Swiezu:
👀

OS-ruimoreiramendes:
Hi @Swiezu, thanks for the fix! Tested this on a physical device using capacitor-testapp:

- Before the patch: FCM push with image renders correctly in background, but the image is missing when the app is in foreground.
- With the patch applied: image now renders identically in both states.
- Also tested with an invalid image URL, the notification still posts (without the image), so the fallback path works as expected.

LGTM 👍

OS-ruimoreiramendes (approved):

PR #2597, opened 2026-09-16
docs(screen-orientation): fix iOS orientation snippet for UIScene

aihdene-genetec:
The `supportedInterfaceOrientationsFor` override documented in the iOS section reads the root view controller from `self.window`:

```swift
return UIInterfaceOrientationMask(rawValue: (self.window!.rootViewController as! CAPBridgeViewController).supportedInterfaceOrientations.rawValue)
```

Capacitor 8.5 [adopted the UIScene lifecycle](https://capacitorjs.com/docs/next/updating/8-5), where the window is created and owned by `SceneDelegate`:

```swift
window = UIWindow(windowScene: windowScene)
window?.rootViewController = CAPBridgeViewController()
```

`AppDelegate.window` is therefore always `nil`, so the force unwrap traps the moment UIKit queries the orientation mask — which happens during `window?.makeKeyAndVisible()`, before the app ever renders. Following the docs as written on 8.5+ gives a crash on every launch:

```
Exception Type:  EXC_BREAKPOINT (SIGTRAP)

0  libswiftCore.dylib  _assertionFailure(_:_:file:line:flags:)
2  App.debug.dylib     AppDelegate.application(_:supportedInterfaceOrientationsFor:) (AppDelegate.swift:51)
3  UIKitCore           _UISafelyGetOrientationMask
4  UIKitCore           -[UIApplication _supportedInterfaceOrientationsForWindow:]
...
13 UIKitCore           -[UIWindow _mainQueue_makeKeyAndVisible]
14 App.debug.dylib     SceneDelegate.scene(_:willConnectTo:options:) (SceneDelegate.swift:20)
```

## Changes

- Read the root view controller from the `window` passed into the method. UIKit supplies this under both lifecycles, so the snippet works before and after 8.5 without branching on version.
- Use a conditional cast instead of `as!`, returning `.all` when the root view controller isn't a `CAPBridgeViewController`. `.all` imposes no app-level restriction, so the view controller and `Info.plist` still govern, matching the previous effective behavior when the bridge controller is present.
- Add a note with a before/after diff, since apps that already copied the old snippet need to change it to launch on 8.5+.

Docs-only; no plugin source is touched. The edit is outside the `<docgen-*>` markers, so `docgen` won't overwrite it, and markdown isn't part of the package's prettier glob (`**/*.{css,html,ts,js,java}`).

## Testing

Verified on a Capacitor 8.5.2 app (Xcode 27.0, iPad A16 simulator, iOS 27.0): the documented snippet crashes on launch as above, and the updated snippet launches and stays running, with orientation locking still applied to presented view controllers.


screen-orientation/README.md
```diff
@@ -16,10 +16,24 @@ For also lock presented View Controllers, this code can be added to the app's `A
...
 func application(_ application: UIApplication, supportedInterfaceOrientationsFor window: UIWindow?) -> UIInterfaceOrientationMask {
-  return UIInterfaceOrientationMask(rawValue: (self.window!.rootViewController as! CAPBridgeViewController).supportedInterfaceOrientations.rawValue)
+  guard let bridgeViewController = window?.rootViewController as? CAPBridgeViewController else {
+    return .all
+  }
+  return bridgeViewController.supportedInterfaceOrientations
 }
 ```
 
+> [!IMPORTANT]
```
OS-ruimoreiramendes:
```suggestion
> **Warning**
>
```

Suggesting **Warning** for consistency with the pattern used in other plugin READMEs

OS-ruimoreiramendes (commented):
Hi @aihdene-genetec, tested it and reproduced the crash on Capacitor 8.5.2 with the current snippet, confirmed the fix launches fine on both lifecycles. Left a small callout style suggestion

aihdene-genetec:
@OS-ruimoreiramendes merged your suggestion.

